PR #5, opened 2020-03-16
Fix jwt_required attributes

Chris-Boe:
jwt_required should not have ()


htharker42 (commented):
thanks for catching, strangely enough, it wasn't picked up by my linter

PR #9, opened 2020-03-27
Put properties

DaveTrost:
Code changes per sprint ticket [Add PUT method feature to /properties resource](https://github.com/codeforpdx/dwellingly-app/issues/105)

Status:
- [x] Modify database tables in data.db to include a column for "archived"
- [x] Add boolean Archived to property model
- [x] PUT method finds property by id
- [x] If the property exists the method sets Archived value = True (if Archived is false) or False (if Archived was true). **However** - if the body of the PUT request includes an archived status, that status will be used, bypassing the toggle logic

PR #10, opened 2020-03-28
Add test users

Chris-Boe:
You can login with the following users (username/pw):
* user1/1234
* user2/1234
* user3/1234


Chris-Boe:
We'll wait to merge this PR until PR #9 is in

PR #15, opened 2020-04-03
Find by role

DaveTrost:
Addresses the items listed in [Issue #120](https://github.com/codeforpdx/dwellingly-app/issues/120)

- [x ] build a new User Model method `find_by_role` which returns all users matching a role
- [x] The end point will be a POST which requires a string of "Userrole"
- [x] JWT claim and admin role are required
- [x] Endpoint is /users/role


htharker42 (dismissed):

htharker42 (approved):
Looks good

PR #23, opened 2020-04-16
Email

htharker42:
Added email endpoint
difficult to test right now as gmail blocks it at the server 
it does hit the server correctly, however, and I verified the Message object created by the request
updated readme

### DIFF
--- a/resources/email.py
+++ b/resources/email.py
@@ -29,7 +29,10 @@ class Email(Resource):
         message = Message(data.title, sender="noreply@codeforpdx.org", body=data.body )
         
         user = UserModel.find_by_id(data.userid)
-        message.recipients = [user.email] 
+        if user.email:
+            message.recipients = [user.email] 
+        else:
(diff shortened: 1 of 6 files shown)


resources/email.py
```diff
@@ -0,0 +1,44 @@
...
+
+class Email(Resource):
+    parser = reqparse.RequestParser()
+    parser.add_argument('userid')
+    parser.add_argument('title') 
+    parser.add_argument('body') 
+
+    @jwt_required
+    def post(self):
+        check if is_admin exist if not discontinue function
```
Chris-Boe: I think this should be a comment like in user.py line 32? Getting a build error here

Edit: when I comment it out it builds

Chris-Boe (changes requested):

resources/email.py
htharker42: Got it! Thanks

Chris-Boe (approved):

PR #24, opened 2020-04-16
Fix test user emails

Chris-Boe:
We had .com instead of .org for the users


Chris-Boe:
Just going to merge this since it's a Readme update

PR #25, opened 2020-04-16
Role optional for new user

DaveTrost:
New users will receive a role of "pending" until an admin approves their access level


Chris-Boe (approved):
👍

PR #27, opened 2020-04-16
List properties

zfregin:
Review changes to model and resources required for /properties GET endpoint according to data required for properties list page (page 14 of mockup)


db.py
```diff
@@ -1,2 +1,19 @@
 from flask_sqlalchemy import SQLAlchemy
+# from models.user import UserModel
+# from models.property import PropertyModel
+# from models.revoked_tokens import RevokedTokensModel
 db = SQLAlchemy()
+
+# def seedData():
```
Chris-Boe, line 7: Should we just delete all these commented lines? Or replace these with the ones added in app.py?

models/property.py
```diff
@@ -9,15 +10,21 @@ class PropertyModel(db.Model):
     city = db.Column(db.String(50))
     state = db.Column(db.String(50))
     zipcode = db.Column(db.String(20))
+    propertyManager = db.Column(db.Integer(), db.ForeignKey('users.id'))
+    tenants = db.Column(db.Integer())
+    dateAdded = db.Column(db.String(50))
```
Chris-Boe, line 15: We may want to consider using the DateTime data type for this dateAdded column

Chris-Boe (approved):

Chris-Boe:
Looks great, just had the two questions!
Created an issue for this so it's separately tracked from the List Properties Page issue

db.py
htharker42: yeah, db.py should just be the two lines of code

htharker42 (approved):
Looks good!

PR #31, opened 2020-05-01
Updated readme setup instructions

HeyImMatt:
- Added wheel install command
- Added step to follow on first runthrough when the virtual env will already be active.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,27 +11,37 @@ NOTE: Database is SQLite3 via SQLAlchemy
 ### To Start Server
 
 1. Clone the repo (`git clone https://github.com/codeforpdx/dwellinglybackend.git`)
+
 2. Install Python ( https://realpython.com/installing-python/ )
-3. Install virtualenv onto system (if not already installed)
+
+3. Install Python Wheels
+
+    - `pip install wheel`
+
+4. Install virtualenv onto system (if not already installed)
 
     - Mac + Linux: `python3 -m venv env`
     - Windows: `python -m venv env`
 
-4. Install dependencies while in main directory of dwellinglybackend forked copy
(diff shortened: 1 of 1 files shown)


HeyImMatt:
Added the couple steps that worked for me. Pretty new to environment setup so feel free to suggest edits if I've stated anything incorrectly!

README.md
```diff
@@ -11,27 +11,37 @@ NOTE: Database is SQLite3 via SQLAlchemy
 ### To Start Server
 
 1. Clone the repo (`git clone https://github.com/codeforpdx/dwellinglybackend.git`)
+
 2. Install Python ( https://realpython.com/installing-python/ )
-3. Install virtualenv onto system (if not already installed)
+
+3. Install Python Wheels
+
+    - `pip install wheel`
```
Chris-Boe, line 19: What is `wheel` used for? I don't think we've needed this before

wittejm:
@Chris-Boe Related question about the existing project setup: why does it rely on both `virtualenv` and `pipenv`? I got the flask server running with only pipenv.

KentShikama:
@HeyImMatt Could you please see if the setup instructions at https://github.com/codeforpdx/dwellinglybackend/pull/32 works?

Chris-Boe:
> @Chris-Boe Related question about the existing project setup: why does it rely on both `virtualenv` and `pipenv`? I got the flask server running with only pipenv.

Not sure! I'm not too familiar with the setup of Python projects so I'd have to ask @htharker42

Chris-Boe:
Hey Matt, if you would still like to contribute to the Readme, please add either comments of what you'd like to add to #38 ! Thanks

PR #33, opened 2020-05-04
Use a decorator to restrict admin required routes - reduce code duplication

DaveTrost:
Most routes in Dwellingly require admin privileges to access them. The code to pull in the JWT and check for the admin role is redundant and exists across many routes. 

This PR introduces a custom decorator that can be used as follows:
```  
    @admin_required
    def post(self):
        # process request
```
 
Old code (for reference):
```
    @jwt_required
    def post(self):
        #check if is_admin exist if not discontinue function
        claims = get_jwt_claims() 
        
        if not claims['is_admin']:
            return {'message': "Admin Access Required"}, 401

        # process request
```

### DIFF
--- a/app.py
+++ b/app.py
@@ -2,6 +2,7 @@ from flask import Flask
 from flask_restful import Api
 from flask_jwt_extended import JWTManager, jwt_refresh_token_required, create_access_token, get_jwt_identity
 from flask_cors import CORS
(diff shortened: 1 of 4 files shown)


Chris-Boe (approved):
Nice! Looks great

PR #36, opened 2020-05-07
Add pending user to backend seed

KentShikama:


Chris-Boe (changes requested):
I like your Readme changes but it looks like @chrisandrewca got to the pending users portion first 😄 If you want to take that commit out we can merge this as a Readme update PR!

PR #38, opened 2020-05-16
Update readme

mgrienauer:
I updated the README to include a section for installation for plebeian Windows users like myself. I spent quite a few hours troubleshooting this and I think it was a pathing issue. Also, I haven't used git / github in projects with more than like 3 contributors, so if there's any etiquette I'm missing please let me know. Thanks, and excited to get to work on more issues!


README.md
```diff
@@ -1,106 +1,134 @@
...
+
+   - Run: `python seed_db.py`
+     - To re-seed the database from scratch, delete data.db before running the script
+   - Look for the file data.db to be created in the root directory
+
+5. Make sure to install flask with `pip install flask`
+6. Run `pipenv install` to auto install remaining dependencies
+7. Start the server using the flask environment (required every time the project is re-opened):
+   - Run: `pipenv run flask run`
+   - Run + restart the server on changes: `pipenv run flask run --reload`
```
KentShikama: Steps 3 ~ 7 are duplicating the non-Windows instructions. Perhaps just a reference to the above would be enough?

mgrienauer: Yea, I can change it to reflect that no problem

README.md
```diff
@@ -1,106 +1,125 @@
...
 
 Queries can be made with the Postman Collection link ( https://www.getpostman.com/collections/a86a292798c7895425e2 )
 
+### Note For Windows Users
+
+Python does not come by default for Windows users. Sometimes the PATH variable in Windows will point to the wrong path and you will have trouble running the `python` and `pipenv` commands. If you don't have Python installed then follow these steps. If the steps listed above did not work for you, try the following.
+
+1. Clone the repo (`git clone https://github.com/codeforpdx/dwellinglybackend.git`)
+2. Hit the Windows key on your keyboard and type `Microsoft Store`. Click on Microsoft store and search for Python. Download Python from the official Microsoft store. When you are installing **make sure you tick the Add to PATH checkbox [ ]** If you don't do this, it could result in you being unable to run the python or pip / pipenv commands.
+3. Once Python is installed, run `pip3 install --user pipenv` If this command doesn't work, you can also try running `python -m pip` followed by running `python -m pip install --user pipenv`
```
KentShikama: Missing a period after the first sentence. `python -m pip` just prints out the usage instructions; `python -m pip install --user pipenv` should be enough. This step also applies for non-Windows users if for some reason their pip binary is a different version from what is expected.

mgrienauer: Ok, thanks for the review. Updated again

htharker42 (dismissed):

dbridenbeck:
Below are the changes I'd recommend based on my experience getting the backend running. Perhaps having a "Troubleshooting" section would be better than having a bullet for troubleshooting in each step. I'm open to either option!

**At the bottom of “4. Seed the database” step, add the following:**
Troubleshooting Tip: If you get errors for missing modules you will need to manually install them one at a time by running:
```
pip install PACKAGE-NAME
```
Then re-run python seed_db.py, and continue installing any missing plugins until you stop getting errors about missing modules. When installing packages, their names should follow the exact syntax from the Pipfile in the root of the project (e.g. - Flask-JWT-Extended).
	
**At the bottom of “5. Start the server using… “ add:**
Troubleshooting Tip: If you get the error "pipenv: command not found", uninstall pipenv and re-install it.

README.md
```diff
@@ -1,106 +1,125 @@
...
-    - Look for the file data.db to be created in the root directory
 
-5. Start the server using the flask environment (required every time the project is re-opened):
-    - Run: `pipenv run flask run`
-    - Run + restart the server on changes:  `pipenv run flask run --reload`
+   - Run: `python seed_db.py`
+     - To re-seed the database from scratch, delete data.db before running the script
+   - Look for the file data.db to be created in the root directory
+
+5. Run `pipenv install`
```
KentShikama, line 23: I think `pipenv install` should come before `pipenv run python seed_db.py`. It may not matter now but if in the future the `seed_db.py` script requires any dependencies, then it would become required.

Chris-Boe: _yes_ `pipenv install` does need to come before seeding the db as we just saw in a meeting 😄

Chris-Boe (changes requested):

Chris-Boe (approved):

PR #40, opened 2020-06-03
Tickets

htharker42:
Added:
Notes Model
Tickets Model
Tickets Resource 
Seed for Database 

Be sure to reseed DB 
New endspoints: tickets/<id>, /tickets


resources/tickets.py
```diff
@@ -1,32 +1,87 @@
...
         if ticket:
             ticket.delete_from_db()
 
         return{'Message': 'Ticket Removed from Database'}
 
+    @admin_required
+    def put(self, id):
+        data = Ticket.parser.parse_args()
+        ticket = TicketModel.find_by_id(id)
+
```
DaveTrost, line 40: Ideally, the ticket's existence should be verified before continuing to modify and save the updates. This can provide better error messages from the server.

DaveTrost (changes requested):
`@admin_required` decorators for the routes in tickets.py probably need to be changed to `@jwt_required` because property manager users will need to have access to view and modify the tickets. 

One more comment in the code should be resolved (regarding the PUT route)

DaveTrost (dismissed):
all good

models/notes.py
```diff
@@ -0,0 +1,57 @@
+from db import db
+from datetime import datetime
+from models.user import UserModel
+
+class NotesModel(db.Model):
+    __tablenname__ = "Notes"
```
Chris-Boe: Looks like a typo in "tablename" here

resources/tickets.py
```diff
@@ -1,32 +1,110 @@
...
+                try:
+                    note.save_to_db()
+                except:
+                    return {'Message': 'An Error Has Occured'}, 500
+
+            if updated:
+                timestamp = dateTime.strftime("%d-%b-%Y (%H:%M)")
+                print(ticket.updated)
+                ticket.updated = timestamp
+                print(ticket.updated)
```
Chris-Boe: Do we want both these print statements here? or were they just for debugging?

Chris-Boe (changes requested):
Just the typo and a question!

models/notes.py
htharker42: fixed it -- strange that it would work with the typo. Don't know what to make of that.

resources/tickets.py
htharker42: Those were just for debugging -timestamp update - are removed now.

PR #42, opened 2020-06-06
Fix foreign key violations in seed_db.py script for Postgres

KentShikama:
We need to commit the session so that the property models can used the updated DB (with users). Ditto for the tenant models. Note this only seems to apply to postgres.


DaveTrost (approved):
Looks good. 
Thank you Kent for continuing to pave the way for deployment!!

PR #49, opened 2020-07-02
Add to README on setuptool version related error

sokampdx:
A lot of good work here - also addressed some other issues we were having with the server


sokampdx:
Sorry, forgot to push the commit before creating the pr. The commit is there now.

htharker42 (approved):

PR #50, opened 2020-07-02
Update tickets

DaveTrost:
The frontend tickets dashboard module needs information about age of tickets relative to the most recent status change. Since the datetime info coming from the backend is formatted, it is difficult to calculate deltas on the frontend. Please update the /tickets/ endpoint to expose the following data points for each ticket:

Date of ticket creation
Date the ticket changed status
Time since last change in status

To Do:

- [x] modify the "updated" property in the ticket model. It is a date string that updates when the ticket's status changes
- [x] create a "minsPastUpdate" property in the ticket model. It is an integer that represents the number of minutes that have passed since the status was last updated

### DIFF
--- a/README.md
+++ b/README.md
@@ -223,20 +223,38 @@ Authorization Bearer < JWT access token >
 
 ```
 
+
+#### ENDPOINT: TICKETS
+
+| method | route              | action (all actions require user to be logged in)   |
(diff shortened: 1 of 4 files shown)


htharker42 (approved):
tested in Postman,
functions correctly 
Good work Dave!

PR #53, opened 2020-07-13
Test emergency contacts

spensbot:
I did some refactoring of the pytest fixtures to make them more modular. I believe this will make future unit tests easier to write and understand. To make this work I needed to refactor the User tests to match. All changes are subject to veto:

- Removed the `seeded_database` fixture, and replaced it with `test_database`
-- I don't think these tests should rely on the seeded database. People should be able to modify how the seeded DB is populated as needed during development without breaking tests.
- Created the `auth_headers` fixture to provide tests with authorization headers for all roles (`admin`, `property-manager`, and `pending`)
- Created `is_valid(response, status_code)` function that checks for valid response headers and status code
- Removed fixtures with duplicate functionality
- Added User tests for various roles (`admin`, `property-manager`, and `pending`)
- Deleted the data.db file and add it to .gitignore (this stupid file prevented me from pulling repo updates any time I ran the server)

I realize that @dcellucci is also working on tests. I will be happy to fix any merge commits caused by the refactor when he submits.

Let me know if you have any questions/concerns


htharker42 (approved):
Tests run correctly!
Good job!

PR #54, opened 2020-07-14
Travis ci

zfregin:
Pytest and TravisCI integration


htharker42 (changes requested):
It looks like we should rebase this branch with development as there have been some significant changes.

PR #55, opened 2020-07-23
Include properties for property manager role

sokampdx:


htharker42 (approved):
Tested in Postman
returns as requested
Good work!

PR #56, opened 2020-07-23
Property multiple tenants

dbridenbeck:
## What issue is this solving?
Closing Backend Issue [#236](https://github.com/codeforpdx/dwellingly-app/issues/236)

## Any helpful knowledge/context for the reviewer?
**Is a reseeding of the database necessary?** Nope!
**Any new dependencies to install?** Nope!
**Any special requirements to test?:** Yes! See below:
- Make sure that all tenants API routes work as expected
- Make sure that all property API routes work as expected
- Property Name should now exist on TenantModel
- An array of Tenants should now exist on PropertyModel

## Other comments:
I've opened issue [#254](https://github.com/codeforpdx/dwellingly-app/issues/254) to add two new attributes to the TenantModel: "Unit", and an "Added On" date.

## Please make sure you've attempted to meet the following coding standards
 **Code builds successfully** - CHECK!
**Code has been tested and does not produce errors** - I did manual testing here. Let me know if you want me to write tests for this.
**Code is readable and formatted** - Check! :)
**There isn't any unnecessary commented-out code** - nope!


htharker42 (dismissed):
Looks good, However, when you refactored tenants it changed the structure of the model. This does cause the seed_db.py to throw the following error (TypeError: __init__() got an unexpected keyword argument 'tenants') - if you could please update the seed file to account for the change in the model. Push when complete.

sokampdx:
@htharker42  @dbridenbeck There is times when we want the property info without the detail list of tenants (e.g /userrole for property-manager, which only list property). Therefore, should we keep the original property and tenant model but using an array of tenant id for tenants? 

I feel the json method in property model file should returns only the info from that table. When we need additional information via resources, we can add other methods to query those information and append to the json. Would that be a good way of minimizing query that we don't need? A way to separating responsibility.

Lastly, we should rebase this on development because https://github.com/codeforpdx/dwellinglybackend/pull/55/files has merged.

dbridenbeck:
Just pushed changes to fix seedData.py and to make PropertyModel.json() return an array of tenantID's.

It seems like we might want to update the Property resource's GET method to return an array of actual tenants instead of ID's, per @sokampdx's suggestion. Let me know if this makes sense, and I can make this change!

models/tenant.py
```diff
@@ -36,8 +34,6 @@ def json(self):
             'phone': self.phone,
             'propertyID': self.propertyID,
             'propertyName': self.property.name if self.property else None,
```
Chris-Boe, line 36: Do we want to get rid of this `propertyName` too?

Chris-Boe (dismissed):
I see the tenantId's coming through the network call, I can open a PR to update the Properties page to references the new variable name

Chris-Boe:
I'm not sure if we need the actual `tenant` objects coming back through the GET call but if you see somewhere that does need that info, let's chat about it!

models/tenant.py
dbridenbeck: Page 17 of the Figma docs suggests we should keep this on the Tenant model, but I could be thinking about this wrong.

I believe the idea of accomplishing page 17 would be to GET Users/?id and the returned data would have a tenants property that lists all of the data needed in the `Tenants` section of page 17, including propertyName.

Let me know if you see a better way of accomplishing this, and I'd be happy to re-work this as needed!

resources/user.py
```diff
@@ -48,9 +48,10 @@ def get(self, user_id):
         user_info = user.json()
 
         if user.role == 'property-manager':
-            user_info['properties'], tenants_ids = zip(*((p.json(), p.tenants) for p in PropertyModel.find_by_manager(user_id) if p))
-            tenants_list = [TenantModel.find_by_id(t) for t in set(tenants_ids)]
-            user_info['tenants'] = [t.json() for t in tenants_list if t]
+            user_info['properties'], tenants_ids = zip(*((p.json(), p.tenantIDs) for p in PropertyModel.find_by_manager(user_id) if p))
```
dbridenbeck: I'm getting the error: `AttributeError: 'PropertyModel' object has no attribute 'tenantIDs'` when trying to do this. Which is confusing, because `tenantIDs` definitely exists on the PropertyModel. @htharker42 @Chris-Boe do you have any suggestions or thoughts on what I might be doing wrong?

**EDIT:** Figured this out, please disregard. Sorry for the noise!

dbridenbeck: Figured this out, please disregard. Sorry for the noise!

Chris-Boe (approved):
Nice work! Thanks for fixing up those tests and everything

PR #74, opened 2020-08-20
Request access

zfregin:
Issues #213 and #145

Is a re-seeding of the database necessary? YES!


app.py
```diff
@@ -81,7 +82,7 @@ def check_for_admins():
     )
     assert (os.path.isfile('./data.db')), errorMsg
     try:
-        admins = UserModel.find_by_role('admin')
+        admins = UserModel.find_by_role(4)
```
NickSchimek: Using integers increases the context a developer needs to have to work on the project. Can we keep the strings or use an Enum?

NLI-ZachFregin: You're right, I should have imported the enum from the user model. Do you think we should store the value in the database as an enum type instead of int? I think we last decided a simple int would suffice that gets evaluated against the enum.

zfregin: You're right, I should have imported the enum from the user model. Do you think we should store the value in the database as an enum type instead of int? I think we last decided a simple int would suffice that gets evaluated against the enum.

NickSchimek: That is a good question, a question I would answer differently depending on the situation. I have no strong opinions one way or another. In this case I would probably prefer to use an enum in the database, this would have the added benefit of having a default role (Pending) added automatically by the db to all newly created users. It turns out we are currently not using the defined UserRole enum anywhere. So if you want to change it then now would be a good time.

Chris-Boe: I do agree we should be using the UserRole enum as much as possible

Chris-Boe (changes requested):
Tested again with the front end and everything works great! Let's resolve the conflicts and try to use that UserRole enum and this'll be good to go

resources/user.py
```diff
@@ -74,8 +77,20 @@ def patch(self,user_id):
         if not user:
             return {"Message": "Unable to find user."}, 400
         
+        # user.role = RoleEnum(data['role'])
```
Chris-Boe: Is this commented out code needed later?

Chris-Boe (changes requested):
Tested again with the front end changes and things look pretty good, just got some unnecessary commented out code

resources/user.py
zfregin: Thanks, somehow missed it during cleanup

htharker42 (approved):

Chris-Boe (approved):

Chris-Boe:
Nice work!

PR #84, opened 2020-09-10
Readmeupdate09092020 deux

htharker42:
Update Readme with database info 

- [ X] Code builds successfully
- [X ] Code has been tested and does not produce errors
- [X ] Code is readable and formatted
- [X ] There isn't any unnecessary commented-out code

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ NOTE: Database is SQLite3 via SQLAlchemy
 4. Run `pipenv install -d`
    - If you get the error `ImportError: cannot import name 'Feature' from 'setuptools'`, your setuptools version might be at 46 or later. You may be able to get it to work using version 45 (e.g. `pip3 install setuptools==45`) 
 5. Seed the database
-   - Run: `pipenv run manage.py create`
+   - Run: `pipenv run python manage.py create`
    - To re-seed the database from scratch, delete data.db before running the script
    - Look for the file data.db to be created in the root directory
    - If you get the error `ImportError: No module named flask` or similar, you may need to run `pipenv shell` to launch virtual environment.
@@ -71,9 +71,9 @@ If you are still having issues or if your command prompt is throwing an error th
 
 
 ### Database commands. 
-- Create and Seed the Database: `pipenv run manage.py create`
-- Seed the Database: `pipenv run manage.py recreate`
-- Delete the Database: `pipenv run mange.py drop`
+- Create and Seed the Database: `pipenv run python manage.py create`
+- Delete, create, Seed the Database: `pipenv run python manage.py recreate`
+- Delete the Database: `pipenv run python mange.py drop`
 
 
 ### Contributing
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -66,6 +69,13 @@ If you are still having issues or if your command prompt is throwing an error th
 6. Run: `pipenv run flask run`
 7. Login using one of the accounts below and you should be good to go!
 
+
+### Database commands. 
+- Create and Seed the Database: `pipenv run manage.py create`
+- Seed the Database: `pipenv run manage.py recreate`
```
NickSchimek: Does recreate seed the database?

NickSchimek:
Why is the user test file being deleted? 🤔

README.md
htharker42: I believe so. Let me double check

README.md
```diff
@@ -18,13 +18,16 @@ NOTE: Database is SQLite3 via SQLAlchemy
...
+
 1. Clone the repo (`git clone https://github.com/codeforpdx/dwellinglybackend.git`)
 2. Install Python ( https://realpython.com/installing-python/ )
 3. Install [pipenv](https://pipenv.pypa.io/en/latest/#install-pipenv-today)
 4. Run `pipenv install -d`
-   - If you get the error `ImportError: cannot import name 'Feature' from 'setuptools'`, your setuptools version might be at 46 or later. You may be able to get it to work using version 45 (e.g. `pip3 install setuptools==45`)
+   - If you get the error `ImportError: cannot import name 'Feature' from 'setuptools'`, your setuptools version might be at 46 or later. You may be able to get it to work using version 45 (e.g. `pip3 install setuptools==45`) 
 5. Seed the database
-   - Run: `pipenv run python seed_db.py`
+   - Run: `pipenv run manage.py create`
```
NickSchimek: this should be `pipenv run python manage.py create`
```
5. Create and Seed the database
        Run: `pipenv run python manage.py create`
```

README.md
NickSchimek: I've checked on this. `recreate` will drop, create, and seed db.
All manage.py commands need to be prefixed with `pipenv run python`

tests/integration/test_users.py
```diff
@@ -1,151 +0,0 @@
-from models.user import UserModel
-from conftest import is_valid, log
-from freezegun import freeze_time
-from models.user import RoleEnum
-
-def test_user_auth(client, test_database, admin_user):
```
NickSchimek: I'm not sure why this file is being removed.  if it is not needed perhaps we can do this in a separate PR.

NickSchimek (changes requested):
@htharker42 the commands as written do not work on my machine. Please check comments for what I think they should be. 
Also, I would prefer not to remove test files in a documentation PR. If these tests must be removed can we create a separate issue/PR for it so that we can document why the removal is needed?

README.md
htharker42: you are correct...

NickSchimek (approved):
Looks good 👍

PR #85, opened 2020-09-11
Add addedOn and unit to tentant model

cphpitts:
https://github.com/codeforpdx/dwellingly-app/issues/254
Add 'addedOn' and 'unit' properties to the Tenant model

Is a re-seeding of the database necessary? Yes

### DIFF
--- a/models/tenant.py
+++ b/models/tenant.py
@@ -2,6 +2,7 @@ from sqlalchemy.orm import relationship
 from db import db
 from models.user import UserModel
 from models.base_model import BaseModel
+from datetime import datetime
 
 
 class TenantModel(BaseModel):
@@ -14,12 +15,12 @@ class TenantModel(BaseModel):
     propertyID = db.Column(db.Integer, db.ForeignKey('properties.id'))
     # leaseID = db.Column(db.Integer, db.ForeignKey('lease.id'))
     addedOn = db.Column(db.Date())
-    unit = db.Column(db.String(20))
+    unitNum = db.Column(db.String(20))
 
     # relationships
     staff = relationship('UserModel', secondary='staff_tenant_links')
 
-    def __init__(self, firstName, lastName, phone, propertyID, staffIDs, unit, addedOn):
+    def __init__(self, firstName, lastName, phone, propertyID, staffIDs, unitNum):
         self.firstName = firstName
         self.lastName = lastName
         self.phone = phone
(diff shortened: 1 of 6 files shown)


resources/tenants.py
```diff
@@ -21,6 +22,8 @@ class Tenants(Resource):
     parser.add_argument('phone',type=str,required=True,help="This field cannot be blank.")
     parser.add_argument('propertyID',required=False,help="This field can be provided at a later time.")
     parser.add_argument('staffIDs',action='append',required=False,help="This field can be provided at a later time.")
+    parser.add_argument('unit',action='append',required=False,help="This field can be provided at a later time.")
+    parser.add_argument('addedOn',type=datetime,required=True,help="This field cannot be blank.")
```
NickSchimek: It's my understanding that the addedOn column would be the date the object is created? Is this date coming from the frontend or can we create it in the backend when we save to the db?

NickSchimek (commented):
Thanks for the PR @cphpitts - I have one concern regarding the required addOn parameter. Please see the below comment. Otherwise everything looks good. There seems to be some failing tests after these changes, I didn't take a look at them, however, they will need to be fixed before we can merge. Let me know if you have any questions.

resources/tenants.py
NickSchimek: It looks like addedOn parameter is not even being used in the endpoint. This line can simply be removed as it is not a required argument.

NickSchimek (changes requested):

models/tenant.py
```diff
@@ -36,7 +40,9 @@ def json(self):
             'phone': self.phone,
             'propertyID': self.propertyID,
             'propertyName': self.property.name if self.property else None,
-            'staff': [user.json() for user in self.staff] if self.staff else None
+            'staff': [user.json() for user in self.staff] if self.staff else None,
+            'unit': self.unit,
+            'addedOn': self.addedOn
```
NickSchimek: This is causing the majority of the failures. A date object is not serializable. It needs to be converted to a string.
`self.addedOn.strftime("%m/%d/%Y, %H:%M:%S")`

cphpitts: This is resolved

cphpitts:
@NickSchimek I believe I have resolved the issues. Let me know if there's anymore issues.

models/tenant.py
NickSchimek: The strftime method was placed in the initializer. It should be here in the json method. We want to store the date in the db as a date object, and when we send to front end it should be the string representation.

After making these changes and running tests locally you should then see 2 failing tests. Those also will need to be resolved. We can pair on it if you like.

cphpitts:
All errors have been resolved.

NickSchimek (dismissed):
Thanks @cphpitts . Looks good!

NickSchimek (approved):

PR #86, opened 2020-09-12
Issue143: Added API to get users by role/name and to get tickets by tenantID

jeremydavidson:
https://github.com/codeforpdx/dwellingly-app/issues/143

Added API to get users by role and name for searching JOIN staff in Tenants view. Added API to get tickets by tenantID.

### DIFF
--- a/models/tickets.py
+++ b/models/tickets.py
@@ -73,3 +73,7 @@ class TicketModel(BaseModel):
             'notes': message_notes
         }
         # notes.json() for note in self.notes.all()]
+    
+    @classmethod
+    def find_by_tenantID(cls, tenantID):
+        return cls.query.filter_by(tenant=tenantID).all()
--- a/resources/tickets.py
+++ b/resources/tickets.py
@@ -89,9 +89,9 @@ class Tickets(Resource):
     def get(self):
         data = Tickets.parser.parse_args()
         if data["tenant"]:
-            return {'Tickets': [ticket.json() for ticket in TicketModel.find_by_tenantID(data["tenant"])]}
+            return {'tickets': [ticket.json() for ticket in TicketModel.find_by_tenantID(data["tenant"])]}
         else:
-            return {'Tickets': [ticket.json() for ticket in TicketModel.query.all()]}
+            return {'tickets': [ticket.json() for ticket in TicketModel.query.all()]}
 
     @jwt_required
     def post(self):
--- a/resources/tickets.py
+++ b/resources/tickets.py
@@ -87,7 +87,11 @@ class Tickets(Resource):
 
     @jwt_required
     def get(self):
-        return {'tickets': [ticket.json() for ticket in TicketModel.query.all()]}
+        data = Tickets.parser.parse_args()
+        if data["tenant"]:
+            return {'Tickets': [ticket.json() for ticket in TicketModel.find_by_tenantID(data["tenant"])]}
+        else:
+            return {'Tickets': [ticket.json() for ticket in TicketModel.query.all()]}
 
     @jwt_required
     def post(self):
--- a/resources/user.py
+++ b/resources/user.py
@@ -154,11 +154,15 @@ class UserLogin(Resource):
 class UsersRole(Resource):
     parser = reqparse.RequestParser()
     parser.add_argument('userrole',type=int,required=True,help="This field cannot be blank.")
+    parser.add_argument('name',type=str,required=False)
 
     @admin_required
     def post(self):
         data = UsersRole.parser.parse_args()
-        users = UserModel.find_by_role(RoleEnum(data['userrole']))
+        if data["name"]:
+            users = UserModel.find_by_role_and_name(RoleEnum(data['userrole']), data['name'])
+        else:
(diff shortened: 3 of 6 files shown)


resources/user.py
```diff
@@ -154,11 +154,15 @@ def post(self):
...
+    parser.add_argument('name',type=str,required=False)
 
     @admin_required
     def post(self):
         data = UsersRole.parser.parse_args()
-        users = UserModel.find_by_role(RoleEnum(data['userrole']))
+        if data["name"]:
+            users = UserModel.find_by_role_and_name(RoleEnum(data['userrole']), data['name'])
+        else:
+            users = UserModel.find_by_role(RoleEnum(data['userrole']))
```
NickSchimek, line 165: I'm curious if the else branch of this will ever evaluate. Depending on how the front end uses/implements this endpoint it might never evaluate. @Chris-Boe what do you think?

resources/tickets.py
```diff
@@ -87,7 +87,11 @@ class Tickets(Resource):
 
     @jwt_required
     def get(self):
-        return {'tickets': [ticket.json() for ticket in TicketModel.query.all()]}
+        data = Tickets.parser.parse_args()
+        if data["tenant"]:
+            return {'Tickets': [ticket.json() for ticket in TicketModel.find_by_tenantID(data["tenant"])]}
+        else:
+            return {'Tickets': [ticket.json() for ticket in TicketModel.query.all()]}
```
NickSchimek: `{'Tickets':` <- These capital `T`'s are causing a failure. I'm not opposed to using a capitalized version, however I think we have a convention of using lowercase everywhere.

resources/tickets.py
```diff
@@ -87,7 +87,11 @@ class Tickets(Resource):
 
     @jwt_required
     def get(self):
-        return {'tickets': [ticket.json() for ticket in TicketModel.query.all()]}
+        data = Tickets.parser.parse_args()
+        if data["tenant"]:
+            return {'Tickets': [ticket.json() for ticket in TicketModel.find_by_tenantID(data["tenant"])]}
```
NickSchimek: Can we use the `find_by_id` method here?

models/tickets.py
```diff
@@ -73,3 +73,7 @@ def json(self):
             'notes': message_notes
         }
         # notes.json() for note in self.notes.all()]
+    
+    @classmethod
+    def find_by_tenantID(cls, tenantID):
```
NickSchimek, line 78: I believe this duplicates existing behavior. Could we remove this and use the inherited `find_by_id` method instead?
https://github.com/codeforpdx/dwellinglybackend/blob/development/models/base_model.py#L8

NickSchimek (changes requested):
Thanks for the PR @jeremydavidson ! Overall the PR looks good. There are some comments that need to be addressed. There is one failing test that needs to be addressed too. Speaking of tests, it would be a good idea to add a couple to ensure this new behavior is doing what we expect it. Although I don't think adding tests is being enforced.

resources/user.py
jeremydavidson: The else-branch was pre-existing, the UI calls it for a couple different userrole arguments, so I didn't want to risk breaking that. The if-branch is for a new feature in the Tenants view where you can search users by name in addition to role.

jeremydavidson: This reminds me to ask: would it be good to refactor this to be a GET with query string arguments instead of a POST? Conventionally, POST is for creating a restful resource, while GET is for retrieving.

resources/tickets.py
jeremydavidson: Thanks for catching this, it was a typo, sorry I missed the failing test.

resources/tickets.py
jeremydavidson: This is for a new feature where we display Tickets on the Tenant view, so we want to retrieve Tickets by Tenant-ID, not by Ticket-ID.

resources/user.py
Chris-Boe: @NickSchimek  Yeah we use this on the dashboard, add-tenant, property managers, and JOIN staff pages to get users by role. I was surprised to see it as a POST when I started adding a call to it, so if you wanted to switch it to a GET that'd be fine with me!

NickSchimek: Yes, I was suprised too. a GET refactor would be welcomed. 

My concern is that if this is used as a form, and if the user does not fill in the name and it is sent as an empty string then in python an empty string is truthy. Hence the if conditional will always evaluate to true. - However if the frontend sends an empty parameter as None then there wont be an issue.

resources/tickets.py
NickSchimek: ah I see. Thanks for clarifying!

NickSchimek (approved):
This looks good. I still have one concern, but we can always come back to it later if it actually causes issues. If @Chris-Boe is OK with it then I'm OK with it.

Chris-Boe (approved):
We can create an issue to refactor that POST to a GET later

resources/user.py
htharker42: This should be a GET request.

PR #87, opened 2020-09-12
Create reset password feature

NickSchimek:
Resolves #codeforpdx/dwellingly-app/issues/277 and Resolves #codeforpdx/dwellingly-app/issues/289

Created two new endpoints. One endpoint fires off an email to the user. The 2nd endpoint verifies the link the user uses in the email. The email sends a short lived (10 minute) jwt containing the user_id to the address on file to verify a user's identity.

Fixed email setup/configs and created tests for the email class.

Other stuff I did are all in the commits, might be best to review commit by commit as I tried to keep each change in the commit small, and wrote an explanation in the commit message when I felt compelled.

Note to self: one json result returns 'msg' when all the others return 'message'. Will probably have to change all of them to 'msg' to match the libraries implementation. It would be good to match the format across the app. We can probably customize the library to return what we want.

Here are some images:

Screen Shot 2020-09-12 at 10 11 06 PM

Screen Shot 2020-09-12 at 10 31 34 PM

Screen Shot 2020-09-12 at 10 13 06 PM

### DIFF
--- a/resources/email.py
+++ b/resources/email.py
@@ -6,17 +6,17 @@ from models.user import UserModel
 
 class Email(Resource):
     parser = reqparse.RequestParser()
-    parser.add_argument('userid', required=True)
-    parser.add_argument('title', required=True)
+    parser.add_argument('user_id', required=True)
+    parser.add_argument('subject', required=True)
     parser.add_argument('body', required=True)
 
     @admin_required
     def post(self):
         data = Email.parser.parse_args()
 
-        message = Message(data.title, sender="noreply@codeforpdx.org", body=data.body )
+        message = Message(data.subject, sender="noreply@codeforpdx.org", body=data.body )
 
-        user = UserModel.find_by_id(data.userid)
+        user = UserModel.find_by_id(data.user_id)
         message.recipients = [user.email]
 
         current_app.mail.send(message)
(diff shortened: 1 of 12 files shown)


resources/email.py
```diff
@@ -1,8 +1,7 @@
-from flask import Flask
+from flask import Flask, current_app
```
htharker42: I was looking for a way to do this without importing app. gtk

resources/email.py
```diff
@@ -17,8 +17,6 @@ def post(self):
         if not data.userid or not data.title or not data.body:
             return {'Message': 'Bad Request'}, 400
 
-        #TODO add feature to send to multiple emails
```
htharker42, line 21: We don't need to worry about batching emails yet, at least not in this version. I was thinking we would setup an email notification system at some point. The idea being that that the backend would periodically audit the existing tickets and dispatch emails based on how long they hadn't been reviewed, changes, other flags. In this use case, we would need to be able to send a batch of messages.

htharker42 (commented):
this corrects a non issue as we have changed the user model

htharker42 (approved):
Testing passes, good work building a complex system. Thank you Nick!

PR #88, opened 2020-09-12
Fix instructions in README for seeding db

cprice404:
The instructions for seeding the db on macosx seemed out of date;
they suggested running `seed_db.py`, which doesn't exist.  When
the flask app fails due to missing db, it instead suggests running
`manage.py create`.  That command worked in my environment and allowed
me to get the flask app up and running without an invalid database error,
so this commit updates the README to recommend that command instead.


NickSchimek (approved):
Thank you!

PR #96, opened 2020-09-15
Add get users api endpoint

jls-github:
https://github.com/codeforpdx/dwellingly-app/issues/296

1. Added in "DEV" as user role 5.
2. Built authorization for dev role. 
3. Data on all users is now accessible by users with this dev role. 

We may want to build the dev authorization somewhat differently. There are two possible issues. 

The first issue is that devs do not have admin access by default. We could add in a check to the @admin_required decorator that checks to see if the user is a dev if not an admin to resolve this issue.

Second, the dev authorization does not have an "is_dev" flag like it has an "is_admin" flag. This may be more an issue of style than anything else. The consequence that we have to run an additional operation and verify the user is a dev by checking the user's email against the database role. 

Yes, there is now a dev user.
No
No

### DIFF
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "8434cfe67ec6f56f900273a4f39faf9ba73795b3dda91caf0ceb6c6ae8d3658d"
+            "sha256": "62135f3ff6e582b1537641d5ae5ef7a3043537ee62661f3c8f32e3b907a1f88c"
         },
         "pipfile-spec": 6,
         "requires": {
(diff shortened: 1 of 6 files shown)


NickSchimek:
Hi @jls-github Thank you for the PR and all the work you put into this. I read the issue that this is for and I do not agree that we should be creating this endpoint for the sole purpose of a developer to use. I have lots of strong opinions about this.

I do not think this endpoint is necessary for the sole purpose of seeing how the database changes, because there are many tools available to developers that allow them to see db changes. One such tool would be https://sqlitebrowser.org/ for sql-lite database and pgadmin for postgres databases. @htharker42 and @Chris-Boe any thoughts on this?

NickSchimek:
There is also a flask shell that allows us to query the database. Here is one way to do that. 
```
➜  dwellinglybackend git:(development) pipenv run flask shell

>>> from models.user import UserModel
>>> UserModel.query.all() # <- You can see this command returns a list of 9 users
[<UserModel 1>, <UserModel 2>, <UserModel 3>, <UserModel 4>, <UserModel 5>, <UserModel 6>, <UserModel 7>, <UserModel 8>, <UserModel 9>]
>>> u = UserModel.query.all()
>>> u[0]
<UserModel 1>
>>> u[0].json()
{'id': 1, 'firstName': 'user1', 'lastName': 'tester', 'email': 'user1@dwellingly.org', 'phone': '555-555-5555', 'role': 4, 'created': '2020-09-18 00:28:50 ', 'archived': False, 'lastActive': '2020-09-18 00:28:50 '}
```

Thank you for this PR. I wish I had seen the issue earlier as I would've raised concerns earlier preventing unnecessary work. After confirming with @htharker42 we will close this PR and the related issue as will not implement. An update to the readme will need to be done to prevent anymore future confusions. @htharker42 had said that this was actually implemented at the beginning to test a few things, but was eventually removed. Thank you and I am looking forward to your future contributions.

PR #97, opened 2020-09-16
Add new token to user self patch

jls-github:
https://github.com/codeforpdx/dwellingly-app/issues/288

This solution only returns a jwt token if the user patches his own data. 

No
No
No

### DIFF
--- a/resources/user.py
+++ b/resources/user.py
@@ -80,10 +80,12 @@ class User(Resource):
 
         if not user:
             return {"Message": "Unable to find user."}, 400
+
+        print(get_jwt_claims())
       
-        if user_id != get_jwt_identity() and user.role != RoleEnum.ADMIN:
+        if user_id != get_jwt_identity() and not get_jwt_claims()['is_admin']:
             return {"Message": "You cannot change another user's information unless you are an admin"}, 403
-            
+
         if data['role']:
           user.role = RoleEnum(data['role'])
         if (data['firstName'] != None):
--- a/resources/user.py
+++ b/resources/user.py
@@ -67,8 +67,6 @@ class User(Resource):
 
         user = UserModel.find_by_id(user_id)
 
-        if user_id != get_jwt_identity() and user.role != RoleEnum.ADMIN:
-            return {"Message": "You cannot change another user's information unless you are an admin"}, 403
 
         parser = reqparse.RequestParser()
         parser.add_argument('role', type=int, required=False, help="This field is not required.")
@@ -83,6 +81,9 @@ class User(Resource):
         if not user:
             return {"Message": "Unable to find user."}, 400
       
+        if user_id != get_jwt_identity() and user.role != RoleEnum.ADMIN:
+            return {"Message": "You cannot change another user's information unless you are an admin"}, 403
+            
         if data['role']:
           user.role = RoleEnum(data['role'])
         if (data['firstName'] != None):
--- a/resources/user.py
+++ b/resources/user.py
@@ -61,7 +61,7 @@ class User(Resource):
 
         return user_info, 200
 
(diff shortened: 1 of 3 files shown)


resources/user.py
```diff
@@ -77,6 +81,12 @@ def patch(self,user_id):
         if not user:
             return {"Message": "Unable to find user."}, 400
       
+        if user_id != get_jwt_identity() and not get_jwt_claims()['is_admin']:
```
NickSchimek, line 84: I would like to see a couple tests added to verify this logic and prevent regression.
I currently see only one test that checks if an admin can update other users here: https://github.com/codeforpdx/dwellinglybackend/blob/development/tests/integration/test_users.py#L128

Can we add two more tests. One to check that a non-admin cannot change another users info. And another to check that a user can update their info and that they do receive an updated jwt upon a successful completion?

NickSchimek (changes requested):
Hi @jls-github  - This looks great! I believe this completely fulfills the requirements stated in the issue. Thank you! Since the behavior of this endpoint has changed, my only suggestion would be to add a couple tests. Ideally endpoints would have very little logic, but when an endpoint like this is overloaded to handle multiple responsibilities it can get complex quite fast. Hence, some added tests would be great. Perhaps in the future we will refactor this to be two endpoints.

resources/user.py
jls-github: @NickSchimek Sounds great. I'll write a couple of tests for the validations.

jls-github:
I added three tests.

1. Validates that a non-admin cannot edit another user's data.
2. Validates that a non-admin can edit their own data.
3. Validates that a non-admin cannot edit their role.

This was my first time writing tests like this, so if there are any errors or if it could be styled better, please let me know. :)

NickSchimek (approved):
Thanks! This looks great. In my opinion, all the tests in this file can be styled better.  let's plan to do it together.

jls-github:
> Thanks! This looks great. In my opinion, all the tests in this file can be styled better. let's plan to do it together.

I'd be happy to do help. I've had an interviewer ask me about writing tests and haven't had a good answer, so that will be helpful.

jls-github:
After thinking a bit more about this, wouldn't it be better to just create another API route instead of have this all in one endpoint? This could be confusing for people working on the front end, and the code seems all over the place with validations. @NickSchimek Do you have any thoughts?

NickSchimek:
@jls-github hmmm that is a good question. We could add another endpoint, but then that wouldn't be very RESTful. And we could end up with a lot of duplicated endpoint just for the admins to use. However, with the current implementation we could also end up with a lot of endpoints having similar authorization logic. 
Just as we wouldn't want authentication logic within an endpoint we also shouldn't have authorization logic. The user should be authenticated and authorized before the endpoint is called; otherwise we should return the correct response. We could probably do this ourselves or see if there is any popular libraries available to use that handle authorization. I plan to audit all the endpoints and create the necessary issues to get them cleaned up, but I would like to ensure we have tests for everything first.

NickSchimek:
@jls-github The role update should not be in this endpoint as its presence here is not RESTful. Regarding that line of code then Yes, we should create a new endpoint for that.

Or maybe it does belong here. I've been staring at this code for too long. We can discuss when we meet.

NickSchimek:
Ideally a new resource would be created for updating user roles. 
The endpoint would be something like: `/api/users/<:id>/roles/<:id>` - Then we can add the admin decorator to this.

And for this endpoint we would remove the ability to update the roles attribute and add authorization so that only the admin or the user in question can access the endpoint.

The way the code is now is OK. Feel free to propose a change, however it will need to be coordinated, and approved with the frontend.

PR #100, opened 2020-09-18
Use python 3.8 to run tests

NickSchimek:
Removed python 3.6, 3.7, and nightly python build.

Running tests under other versions of python is not necessary for our
application, and doing so only increases CI time and could trigger
unnecessary CI failures.

This change reduced the CI time from `Total time 7 min 9 sec` down to `1 min 28 sec`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,17 +1,7 @@
 language: python
-
 python:
-  - "3.6"
-  - "3.7"
   - "3.8"
-  - "nightly"
-
-matrix:
-  allow_failures:
-    - python: "nightly"
-
 install:
   - pip install pipenv --upgrade-strategy=only-if-needed
   - pipenv install --dev
-
 script: pytest


.travis.yml
```diff
@@ -1,17 +1,7 @@
...
 python:
-  - "3.6"
-  - "3.7"
   - "3.8"
-  - "nightly"
-
-matrix:
-  allow_failures:
-    - python: "nightly"
-
```
zfregin, line 12: What about leaving the nightly test? This just tests against future releases. If it fails that is allowed, but if it is passing then we know there is no concern about updating to a major release

NickSchimek: I don't think anyone is going to be watching if the nightly test passes or fails. And python in general is pretty conservative with new updates. I believe they like to stay backwards compatible as much as possible. Leaving this in increases the amount of time one has to wait for the CI to run on their PR. And if the nightly test does start failing it could be out of our control to fix. For example: The failure could be caused due to a library we are using. I would suggest leaving it out.

zfregin (approved):

PR #102, opened 2020-09-18
Replace unmaintained flask-mail dep

### LINKED ISSUE #351: Flask-mail is unmaintained for last 6 years
We may want to move off of the Flask-mail extension soon. I went to look through their source code and realized it is no longer maintained. It seems there are some forked versions of it out there, along with completely different libraries that could be used. We should find one that is well maintained, easy to use, and preferably popular.

NickSchimek:

Fixes https://github.com/codeforpdx/dwellingly-app/issues/351


NickSchimek:
I think I jumped too soon. Lets do some more research for a better candidate.

PR #103, opened 2020-09-19
Fix/Update app configuration

### LINKED ISSUE #252: BACKEND: running pytest coverage test crash currently running backend server
This is happening at latest development branch (f455a67). 
Follow README to setup the dwellinglybackend environment.

Start the backend server or reload the backend server:
`pipenv run flask run --reload`

Run pytest coverage:
`pipenv run pytest --cov .`

The following things happend:
-  data.db file would updated 
-  the backend server would crash with error similar to this
`...
  File "/Users/kso/.local/share/virtualenvs/dwellinglybackend-n8ZLeKs6/lib/python3.8/site-packages/sqlalchemy/engine/default.py", line 588, in do_execute
    cursor.execute(statement, parameters)
sqlalchemy.exc.OperationalError: (sqlite3.OperationalError) no such table: revoked_tokens
[SQL: SELECT revoked_tokens.id AS revoked_tokens_id, revoked_tokens.jti AS revoked_tokens_jti 
FROM revoked_tokens 
WHERE revoked_tokens.jti = ?
 LIMIT ? OFFSET ?]
[parameters: ('ee0dee3b-db48-417e-8e5a-de186f770b6c', 1, 0)]
(Background on this error at: http://sqlalche.me/e/e3q8)
`

The behavior we should get is
- data.db should remove from git repo and clear and re-seed when test is done.
- the server should not crash after running test (this might be fix by fixing the first item)

### LINKED ISSUE #307: Refactor and secure app configs
## Description

App configurations in the app.py file should be moved to a separate file, env vars should be secured, and multiple config environments should be setup.

The RecordSponge has a good example of this and this resource also looks solid https://hackersandslackers.com/configure-flask-applications/

NickSchimek:
## Description

Created three environments for the app to run in. Development,
Production, and Testing. Before this refactoring the app always ran in a
production environment which is OK for a while, but can cause issues
when developing. For instance: The DB would be wiped out everytime the
test suite was ran which could cause dev frustration and the need to
recreate the DB a lot. Furthermore, libraries may contain environment
specific configurations depending on which environment the app is
running in, and it is best to take advantage of that.

Removed the method and associated test that checked to see if the DB was
seeded. This was helpful when there was a single environment for the app
to run in. However, now I don't see it being necessary as the dev
environment can be seeded once and will not be wiped anymore when tests
are ran. Plus removing non-app specific code improves the code base. The
friendly dev message it provided is nice, however it would get buried in
the stack trace. I think it would be more helpful to let the SQLalchemy
db exception propagate to the top which makes it pretty clear that the db
is the issue.

Finally, secured all env vars. There are now two env files to store
environment variables in for developing. The first is a public file that
can be used to store, well, public env vars `.flaskenv`. And the second
is a private file that is used to store private (secret) env vars.

Currently there are only two env vars needed for this app to run. The
first is flask specific and is stored in `.flaskenv` which tells flask
what environment to use when developing. For example: when using any
Flask commands such as `flask run`. The second is the secret_key which
is stored in the private `.env` file.

Updated the README to include instructions for copying the
`.env.example` file.

**Update**
Removed PROPAGATE_EXCEPTIONS from the default config
as we don't want propagate exceptions set to True in the
Production environment. Dev and Test env will have this set to True by
default, because the TESTING and DEBUG flags are turned on.

Fixes codeforpdx/dwellingly-app/issues/252
Fixes codeforpdx/dwellingly-app/issues/307

### 🆘Tips on checking out this branch.🆘

Not necessary, but probably preferred to first drop your database, as db names have changed, then checkout the branch.
run `pipenv install -d`
run `cp .env.example .env`
Then you can create the DB and do whatever you would like.

- [x] Code is readable and formatted (Probably)

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,0 +1,1 @@
+SECRET_KEY='dwellingly'
--- a/.flaskenv
+++ b/.flaskenv
@@ -1,0 +1,1 @@
+FLASK_ENV=development
--- a/.gitignore
+++ b/.gitignore
@@ -65,6 +65,9 @@ db.sqlite3-journal
 instance/
 .webassets-cache
 
+# SQL Lite DB's
+data-dev.sqlite
+
 # Scrapy stuff:
 .scrapy
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,3 +1,4 @@
+# Contributing
 ## Installation
 Set up Dwelling Flask Testing Backend (for the first time)
 NOTE: Database is SQLite3 via SQLAlchemy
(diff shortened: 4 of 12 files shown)


yunnie (approved):
Good refactoring of the application configurations.

PR #104, opened 2020-09-19
Hash Passwords

cecillesalazar:

Please post a link to the issue (on the front end repo) that this PR is resolving.

https://github.com/codeforpdx/dwellingly-app/issues/280

Previously we had been storing passwords in plain text - no more with bcrypt! More notable changes is that the password hash being stored in the db has now changed in type from string to large binary. 

_Is a re-seeding of the database necessary?_

Yup, there was some changes to the seed file. 

_Any new dependencies to install?_

Yup, bcrypt. It's added to the Pipfile. 

_Any special requirements to test?_

I've only ever pulled the backend - and never setup the frontend. So whoever tests this DEFINITELY should pull both down and make sure they can log-in properly through the whole flow. 

### DIFF
--- a/resources/user.py
+++ b/resources/user.py
@@ -8,6 +8,7 @@ from models.revoked_tokens import RevokedTokensModel
 import json
 from werkzeug.security import safe_str_cmp
 from flask_jwt_extended import create_access_token, create_refresh_token, jwt_required, get_jwt_claims, get_raw_jwt, get_jwt_identity, jwt_refresh_token_required
+import bcrypt
 
 
 class UserRoles(Resource):
@@ -31,13 +32,31 @@ class UserRegister(Resource):
     def post(self):
         data = UserRegister.parser.parse_args()
 
+        # I think the above print statement will ALWAYS be false because upon saving - we're always hashing
+        # So we're NEVER seeing the plaintext password anymore
+        # I think there was a lot of learning here...but ultimately...we're not doing it right
+        # When we register
+        # We need to take the plain text password
+        # hash that
+        # store it in the db
+        # then when we login -
+        # we need to take that same plain text password
+        # hash it again --> don't think this needs to happen - because checkpw function can make the comparison
+        # then compare it to the db
+        # start here when you come back
+
+        # Let's start by hashing the password
+        # We'll convert the string into bytes for bcrypt to be able to process
+        hashed_password = bcrypt.hashpw(bytes(data['password'], 'utf-8'), bcrypt.gensalt())
+
         if UserModel.find_by_email(data['email']):
             return {"message": "A user with that email already exists"}, 400
 
         user = UserModel(firstName=data['firstName'],
                          lastName=data['lastName'], email=data['email'],
-                         password=data['password'], phone=data['phone'],
+                         password=hashed_password, phone=data['phone'],
                          role=RoleEnum(data['role']) if data['role'] else None, archived=data['archived'])
+        # And we'll store it into the db as bytes
         user.save_to_db()
 
         return {"message": "User created successfully."}, 201
@@ -152,6 +171,7 @@ class UserLogin(Resource):
     parser.add_argument('password', type=str, required=True, help="This field cannot be blank.")
 
     def post(self):
+        print('IN THE POST ENDPOINT')
         data = UserLogin.parser.parse_args()
 
         user = UserModel.find_by_email(data['email'])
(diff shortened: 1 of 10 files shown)


resources/user.py
```diff
@@ -30,14 +31,16 @@ class UserRegister(Resource):
 
     def post(self):
         data = UserRegister.parser.parse_args()
+        hashed_password = bcrypt.hashpw(bytes(data['password'], 'utf-8'), bcrypt.gensalt())
```
NickSchimek: I see this algorithm in a few places throughout the codebase. Is there a way we can dry this up a bit?

tests/integration/test_users.py
```diff
@@ -4,10 +4,14 @@
 from models.user import RoleEnum
 from flask_jwt_extended import create_access_token, create_refresh_token
 
+plaintext_password = "1234"
+
 def test_user_auth(client, test_database, admin_user):
+    print(admin_user.email)
+    print(admin_user.password)
```
NickSchimek: Let's remove these print statements.

NickSchimek (changes requested):
This looks great! Thanks @cecillesalazar . I have a couple concerns. One is based on design/architecture. Although it is not an issue that would stop this PR from being merged, however it would be nice to dry up the codebase in this PR.

The other concern is over bcrypt's handling of passwords greater than 72 characters. What's your thoughts on the [Maximum password length](https://github.com/pyca/bcrypt#maximum-password-length) issue when using bcrypt?

Also, I hope the frontend is not sending plaintext passwords over the wires that can be read by any eavesdropper. We may have to create a separate issue for this if that is the case.

resources/user.py
cecillesalazar: Good catch 👍 I created a /utils folder. I figured if we need to make future helper methods - might be good for everything to be in one place~

cecillesalazar:
@NickSchimek in terms of the max password length - I think it might be easier to sanitize the input from the frontend and limit the password length for registering users to less than the threshold where we would have to do backend finagling. Also - I never work in the frontend repo so I'm not sure if there's any protections for sending the plaintext password. Might be good to make a ticket to check/implement that just in case! 😅

EDIT: I think @Chris-Boe already beat us to the punch in thinking about it - as there's a ticket - https://github.com/codeforpdx/dwellingly-app/issues/319.

NickSchimek:
@cecillesalazar These updates are a big improvement. Thank you! There is one flaw I didn't quite catch the other day, it might have not been an issue depending on how the code to hash a password was reorganized. This also means we're missing a test. The issue is with the patch action, we're not hashing passwords when users update their password. There is a way we could implement a solution so that hashing passwords are done automatically which would reduce the context a developer needs to have to work with users, and has the benefit of preventing bugs of this nature slipping through to production. I'm also fine with the current implementation you have if you want to continue that way. If you would like to talk more about other design possibilities, then reach out to me on slack or we can also talk at the team meeting.

utils/auth.py
```diff
@@ -0,0 +1,9 @@
+import bcrypt
+
+def hash_pw(plaintext_password):
+    if isinstance(plaintext_password, str):
+        return bcrypt.hashpw(bytes(plaintext_password, 'utf-8'), bcrypt.gensalt())
+
+def check_pw(plaintext_password, db_password):
+    if isinstance(plaintext_password, str):
```
NickSchimek: Is this type check necessary? What should we do if it is not a string?

NickSchimek:
@cecillesalazar I just noticed that the test suite runs very slow on this branch. bcrypt consumes a lot of computing time for hashing passwords. For me, having fast tests is of the utmost importance. If you could commit a fix for this that would be great, if not we can create a separate issue.

cecillesalazar:
Thanks for taking another look! I'll have some time to push a fix for the tests tmrw and I should be able to add to the patch endpoint as well~

utils/auth.py
cecillesalazar: Nope! I added it initially as a safeguard in case we had the case of "1234"(str) vs 1234(int). I tested the latter, and it still parsed as a string - so I removed the check~

cecillesalazar:
@NickSchimek - I added the hashing to the patch endpoint. Sadly, because we do not include `password` in the returned json for the user model (probably best not to, tbh), I couldn't add the password comparison in `test_patch_user`. I did check for sure that it hashes and sets the new password properly just by hitting the endpoint and using prints statements~

I played a little bit in optimizing the tests - but didn't have a ton of time to do any real benchmarking - so yep probably best on another ticket. I'll make the issue 👍

NickSchimek:
> Sadly, because we do not include password in the returned json for the user model (probably best not to, tbh), I couldn't add the password comparison in test_patch_user.

That's right, we should never return password in JSON response. There is a simple way to test this without relying on the JSON response. Although, I'm not too concerned about a test here as I plan to simplify the code base which should make this kind of test irrelevant. 

> I played a little bit in optimizing the tests - but didn't have a ton of time to do any real benchmarking

There is no need for benchmarking here. With the right test setup it will be blazingly fast... well as fast as it can be.

NickSchimek (approved):
Thanks @cecillesalazar 💫

I believe this is ready to go. I'm going to hold off on merging this just for a bit.

cecillesalazar:
Awesome 👍 Just ping me when you're ready - happy to rebase the branch~

PR #106, opened 2020-09-20
Iss336 response status

yunnie:
[Issue 336:](https://github.com/codeforpdx/dwellingly-app/issues/336) Update response statuses for all lease endpoints.

No additional libraries, or reseeding of the database required.

- [ X] Code builds successfully
- [ X] Code has been tested and does not produce errors
- [ X] Code is readable and formatted
- [ X] There isn't any unnecessary commented-out code

### DIFF
--- a/tests/integration/test_leases.py
+++ b/tests/integration/test_leases.py
@@ -85,7 +85,7 @@ class TestCreateLease:
         num_leases = len(LeaseModel.query.all())
         response = self.client.post(self.endpoint, json=self.valid_payload, headers=auth_headers["pm"])
 
-        assert is_valid(response, 200)
+        assert is_valid(response, 201)
         assert response.json == 201
         assert num_leases + 1 == len(LeaseModel.query.all())
 
@@ -118,8 +118,8 @@ class TestDeleteLease:
         num_leases = len(LeaseModel.query.all())
         response = self.client.delete(f'{self.endpoint}/504', headers=auth_headers["pm"])
 
-        assert is_valid(response, 200)
-        assert response.json == {'Message': 'Lease Removed from Database'}
+        assert is_valid(response, 422) 
+        assert response.json == {'Message': 'Lease Not In Database'}
         assert num_leases == len(LeaseModel.query.all())
 
 
@@ -132,7 +132,7 @@ class TestUpdateLease:
     def test_valid_lease_id(self, auth_headers):
         response = self.client.put(f'{self.endpoint}/{self.lease.id}', headers=auth_headers["pm"])
 
-        assert is_valid(response, 201)
+        assert is_valid(response, 200)
         assert response.json == self.lease.json()
 
     def test_invalid_lease_id(self, auth_headers):
@@ -148,7 +148,7 @@ class TestUpdateLease:
         with freeze_time(Time.one_year_from_now()):
             response = self.client.put(f'{self.endpoint}/{self.lease.id}', json=payload, headers=auth_headers["pm"])
             
-            assert is_valid(response, 201)
+            assert is_valid(response, 200)
             assert response.json['dateUpdated'] != old_date
 
     def test_updatedDate_is_left_unchanged_when_given_invalid_params(self, auth_headers):
@@ -157,7 +157,7 @@ class TestUpdateLease:
         with freeze_time(Time.one_year_from_now()):
             response = self.client.put(f'{self.endpoint}/{self.lease.id}', json={}, headers=auth_headers["pm"])
 
-        assert is_valid(response, 201)
+        assert is_valid(response, 200)
         assert response.json['dateUpdated'] == old_date
         
     def test_invalid_attribute_ids(self, auth_headers):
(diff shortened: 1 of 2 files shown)


tests/integration/test_leases.py
```diff
@@ -251,14 +251,16 @@ def test_authorized_create_request(self, auth_headers):
...
-            assert is_valid(response, 200)
+            assert is_valid(response, 201)
 
     def test_authorized_delete_request(self, auth_headers):
+        id = 1
         for _, role in auth_headers.items():
-            response = self.client.delete('/api/lease/1', headers=role)
+            response = self.client.delete('/api/lease/{}'.format(id), headers=role)
             assert is_valid(response, 200)
+            id += 1
```
NickSchimek, line 255: Thanks for catching. I suppose this wont be an issue as long as we have more seeds then roles. I winced when I wrote these for loops here.

NickSchimek: FYI: I often see a 204 returned on a delete. However, we are returning a json message so this does seem more appropriate. Thanks!

tests/integration/test_leases.py
```diff
@@ -157,7 +157,7 @@ def test_updatedDate_is_left_unchanged_when_given_invalid_params(self, auth_head
         with freeze_time(Time.one_year_from_now()):
             response = self.client.put(f'{self.endpoint}/{self.lease.id}', json={}, headers=auth_headers["pm"])
 
-        assert is_valid(response, 201)
+        assert is_valid(response, 200)
```
NickSchimek: Should we be returning 200 here? I suppose depending on the perspective a 200 status is OK.

tests/integration/test_leases.py
```diff
@@ -118,8 +118,8 @@ def test_authorized_request_with_invalid_lease_id(self, auth_headers):
         num_leases = len(LeaseModel.query.all())
         response = self.client.delete(f'{self.endpoint}/504', headers=auth_headers["pm"])
 
-        assert is_valid(response, 200)
-        assert response.json == {'Message': 'Lease Removed from Database'}
+        assert is_valid(response, 422)
```
NickSchimek: I know I said to use 422 here, however it looks like we are already using 404 throughout the app, and I believe it's best to stay consistent. In regards to the 404 status the Mozilla docs say: 
> In an API, this can also mean that the endpoint is valid but the resource itself does not exist

So I guess it is not terribly incorrect to use a 404 here. What do you think?

NickSchimek (commented):
Thank you! Just one question for you.

tests/integration/test_leases.py
yunnie: Using 404 in this context makes sense.

tests/integration/test_leases.py
yunnie: Good catch. Since this is a failed update I think the status code should be:
422 Unprocessable Entity: The request was well-formed but was unable to be followed due to semantic errors.
or possibly 
400 Bad Request:  The server could not understand the request due to invalid syntax.

NickSchimek: Yeah, I think sending an empty payload to a patch request should automatically be rejected as a 400 bad request. Because at this point the entity in question could very well be processable, but there was no data to process. Hence, a bad request.

yunnie:
Should I close this pull request and then resubmit after making the two changes discussed?

NickSchimek:
> Should I close this pull request and then resubmit after making the two changes discussed?

It is not necessary, I'd prefer to push a new commit on the branch here. Especially for bigger PR's as it makes it easier to re-review the new changes.

tests/integration/test_leases.py
```diff
@@ -132,7 +132,7 @@ def setup(self):
     def test_valid_lease_id(self, auth_headers):
```
NickSchimek: These are excellent changes, However, the name of this test is a bit confusing now. We're testing something that should be valid, but asserting a 400 error.

I agree with the change in the logic to say that an empty payload should be a bad request, and since new logic was added we should update our tests accordingly. I'm fine to do that in this ticket or we can create a new ticket for it. What would you like to do?

yunnie: I think test is not necessary. test_date_is_updated_with_valid_payload tests both that the lease id is valid and that an update is made with a valid payload. And test_updatedData_is_left_unchanged_when_given_invalid_parameters tests for a valid lease id with an empty payload, which is essentially the same test as test_valid_lease_id. 

test_invalid_lease_id tests the case where the lease id is not valid. 

I recommend removing test_valid_lease_id.

yunnie: If you agree, I'll just make the change.

NickSchimek:
Go for it

tests/integration/test_leases.py
yunnie: Or we can rename the test. test_valid_id_empty_payload.

NickSchimek: Yes, I agree. Let's remove the test.

NickSchimek: or perhaps we can rename it to `test_valid_id_with_invalid_payload` - I don't think we have any tests that check what happens when there is an invalid payload.

NickSchimek: nevermind. I need to learn how to read better. 
> And test_updatedData_is_left_unchanged_when_given_invalid_parameters tests for a valid lease id with an empty

Yes, agreed. Let's remove it.

NickSchimek (approved):
Thank you !!!

PR #110, opened 2020-09-24
Add get user by role endpoint

jls-github:

codeforpdx/dwellingly-app/issues/296 (should not close - frontend work required)

Supplies backend api endpoint to display user data based on queried role (formatted '/user?r=<int>'). Endpoint is locked behind admin permissions. This will allow us to populate the JOIN staff page on the front end.

### DIFF
--- a/resources/user.py
+++ b/resources/user.py
@@ -43,6 +43,11 @@ class UserRegister(Resource):
         return {"message": "User created successfully."}, 201
 
 class User(Resource):
+
+    @admin_required
+    def get(self):
+        return {}, 200
+
     @admin_required
     def get(self, user_id):
         user = UserModel.find_by_id(user_id)
--- a/resources/user.py
+++ b/resources/user.py
@@ -43,11 +43,6 @@ class UserRegister(Resource):
         return {"message": "User created successfully."}, 201
 
 class User(Resource):
-
-    @admin_required
-    def get(self):
-        return {}, 200
-
     @admin_required
     def get(self, user_id):
         user = UserModel.find_by_id(user_id)
@@ -209,3 +204,8 @@ class UserAccessRefresh(Resource):
             'access_token': create_access_token(identity=current_user)
         }
         return ret, 200
+
+class Users(Resource):
+    @admin_required
+    def get(self):
+        return {}, 200
--- a/resources/user.py
+++ b/resources/user.py
@@ -1,4 +1,5 @@
 from flask_restful import Resource, reqparse
+from flask import request
 
 from models.property import PropertyModel
 from models.tenant import TenantModel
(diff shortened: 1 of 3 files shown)


tests/integration/test_users.py
```diff
@@ -197,4 +197,26 @@ def test_delete_user(client, auth_headers, new_user):
...
+def test_get_user(client, auth_headers, new_user):
+    """GET '/user' returns a list of all users queried by role"""
+
+    staff_user_response = client.get('/api/user?r=3', headers=auth_headers["admin"])
+    admin_user_response = client.get('/api/user?r=4', headers=auth_headers["admin"])
+
+    assert is_valid(staff_user_response, 200)
+    assert is_valid(admin_user_response, 200)
+    assert all(staff["role"] == 3 for staff in staff_user_response.json["users"])
+    assert all(admin["role"] == 4 for admin in admin_user_response.json["users"])
```
NickSchimek:
```suggestion
    assert all(staff["role"] == RoleEnum.STAFF.value for staff in staff_user_response.json["users"])
    assert all(admin["role"] == RoleEnum.ADMIN.value for admin in admin_user_response.json["users"])
```

tests/integration/test_users.py
```diff
@@ -197,4 +197,26 @@ def test_delete_user(client, auth_headers, new_user):
...
-    assert is_valid(response, 400) # BAD REQUEST
\ No newline at end of file
+    assert is_valid(response, 400) # BAD REQUEST
+
+
+def test_get_user(client, auth_headers, new_user):
+    """GET '/user' returns a list of all users queried by role"""
+
+    staff_user_response = client.get('/api/user?r=3', headers=auth_headers["admin"])
+    admin_user_response = client.get('/api/user?r=4', headers=auth_headers["admin"])
```
NickSchimek:
```suggestion
    staff_user_response = client.get(f'/api/user?r={RoleEnum.STAFF.value}', headers=auth_headers["admin"])
    admin_user_response = client.get(f'/api/user?r={RoleEnum.ADMIN.value}', headers=auth_headers["admin"])
```

resources/user.py
```diff
@@ -204,3 +205,27 @@ def post(self):
...
         return ret, 200
+
+class Users(Resource):
+    @admin_required
+    def get(self):
+
+        role_query = int(request.args["r"])
+
+        if role_query not in range(0, 5):
+            return {"message": "Invalid role"}, 400
```
NickSchimek, line 216: OK for now, but will break when things change. And the RoleEnum will change soon.

NickSchimek (changes requested):
Thanks for getting this done so fast. 

This assertion `assert all(staff["role"] == 3 for staff in staff_user_response.json["users"])` is technically not asserting anything (replace the 3 with 'anything' to test yourself.), because the query returns an empty response as there is currently no staff seeds. We can remove this and leave the admin tests. The admin tests give me enough confidence that it works correctly. 

Could we replace all the magic numbers with the RoleEnum.

resources/user.py
jls-github: Ahh okay. How is the RoleEnum changing? Should we just prepare to change this now?

NickSchimek: No, let's not take care of it now. I'll create a new issue for that in due time.

NickSchimek (approved):

PR #113, opened 2020-09-26
Small step forward in revamping the test suite

NickSchimek:
## Description
Improved the robustness and speed of the `lease` integration tests by a factor of 2 to 3x. This improvement is based on results before bcrypt was installed.

- Replaced the auth_headers with new ones that do not hit the login endpoint at each run.
- Implemented an empty test db - This is being done to improve robustness so that tests will not fail when the seed file changes, and of course initializing all the seeds for each test is the main source of the slow tests.
- Created factories for the lease endpoints, and implemented the new factories into the lease file.

Run times for the lease integration test file:

- with bcrypt ~19 seconds
- without bcrypt ~5.5 seconds
- This branch w/bcrypt ~2.3 seconds

This is a small step forward. More improvements to come which will allow us to type less while writing better and faster tests.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -6,11 +6,8 @@ from db import db
 from data.seedData import seedData
 from models.user import UserModel, RoleEnum
 from models.property import PropertyModel
-from models.lease import LeaseModel
-from models.tenant import TenantModel
 from utils.auth import hash_pw
 import jwt
-from datetime import datetime
 
 newPropertyName = "test1"
 newPropertyAddress = "123 NE FLanders St"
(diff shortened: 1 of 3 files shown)


yunnie (approved):
Changes look good. Should make it easier to create new entries to test creates and updates.

jls-github (commented):
Thanks Nick. I'm gathering we're making three changes to each of the test files.

1. We're declaring our test data at the top of the file with the @pytest.fixture decorator
2. We decorate each test class with pytest.mark.usefixtures('client_class', 'empty_test_db') - which causes the test to use the empty db instead of a real db with seeds
3. We've changed our headers to valid_headers and auth_headers

NickSchimek:
1. Not quite, or rather don't assume that just yet. It is just what I've done for now. I've got a cleaner solution in mind. I just need to try it out when the opportunity presents itself.
2. Yes, that is one way to do it. <- I will explain more 
3. Kinda, I want to remove usage of the auth_headers completely. It will be replaced with the `valid_header` . The auth_headers make a post request to the login endpoint. so every endpoint test that uses the auth_headers is also testing the login endpoint. This slows down the tests and makes them more vulnerable in the long run. As far as the admin, staff, and pm_header - You could say they're still in beta. We'll probably need to improve these a bit in order to ensure our app's authorization implementation is correct.

NickSchimek:
Once I got everything ready to go, it will be clear what to do. Thanks for asking questions

PR #114, opened 2020-09-27
Create base interface test

NickSchimek:
## Description

Created unit tests for the base model. This was implemented in a
separate file so that it can be used to test the inherited classes
conform to the interface of the base model. This is needed to ensure
nothing is done to override these methods.

Created a new method find. This will eventually replace the find_by_id
method since the find method will automatically return a 404 status
code and custom json message. This will reduce the number of
conditionals needed in the endpoints and the number of integration tests
needed.

Implemented the Base Interface Test into the lease model test.

Update: Added a delete method on the base class. This is a class method that is planned to replace the delete_from_db instance method. The class method will automatically return a 404 status code and custom message when the object being deleted cannot be found in the db.

### DIFF
--- a/models/base_model.py
+++ b/models/base_model.py
@@ -8,6 +8,10 @@ class BaseModel(db.Model):
     def find_by_id(cls, id):
         return cls.query.filter_by(id=id).first()
 
+    @classmethod
+    def find(cls, id):
+        return cls.query.get_or_404(id, f'{cls.__name__} not found')
+
     def save_to_db(self):
         db.session.add(self)
         db.session.commit()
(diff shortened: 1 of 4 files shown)


yunnie (approved):
Looks good.

jls-github (approved):
To make sure I've got this right - We're inheriting from the BaseInterfaceTest, we won't have to rewrite any of those methods in the unit tests unless something is unique to that method. We'll just have to write the setup method and whatever unique tests we'll need for each model. Does that sound right?

NickSchimek:
> To make sure I've got this right - We're inheriting from the BaseInterfaceTest, we won't have to rewrite any of those methods in the unit tests unless something is unique to that method. We'll just have to write the setup method and whatever unique tests we'll need for each model. Does that sound right?

That's correct. You will only need to write the setup method. And then if the model implements any other methods then those will need to be tested. For example the `json` method would need to be tested, because it is a unique method.

PR #115, opened 2020-09-27
Fix typo in email templates: Dwellinlgy -> Dwellingly

langston-barrett:
I left out the MR template info because this is such a small change.


NickSchimek (approved):
Thank you for catching this!

PR #117, opened 2020-09-28
Refactor lease endpoint/tests example

### LINKED ISSUE #359: Update action for all resources
Need to look into how the FE is implementing the updates for resources. This is to help make the design of the app simpler.
For example: Should we allow empty payloads when updating a resource. Should we enforce the entire resource be sent?

Proposal: 
 - Automatically reject request as bad request when there is an empty payload to any update action. 
 - Multi-line if conditionals in the update actions take time to read. And code is read more often than it is written.
     - Implement one of the following:
         -  Enforce FE to send all attributes, then we can simply update resource in one go
         - Allow FE to send one or more attributes and only update those attributes without using any conditionals in the update action. (Perhaps there is an easy way to do this. If not I propose we engineer a way to do this)

### LINKED ISSUE #360: Refactor lease endpoint
**Purpose**
The lease endpoint could probably be improved

On hold - Pending completion of #336 #337 #338 and #359

NickSchimek:
Refactor the lease endpoint.
Removes dependency on the deprecated Reqparse and uses Marshmallow for validation, serialization, and deserialization
Makes use of new abstractions, The `find`, `create`, `update`, and `delete` methods to remove logic out of the endpoints.

Implements validation on foreign keys which was previously not being done, and introduces validation tests.

This is only an example, and making use of this example and the new methods for interacting with the db should reduce complexity, and bugs. This is by no means a perfect example of what should always be done, and can most likely be improved and refactored in the future. There is probably a better a way to do this so please let me know your thoughts or else.

Resolves codeforpdx/dwellingly-app/issues/359
Resolves codeforpdx/dwellingly-app/issues/360

### DIFF
--- a/data/seedData.py
+++ b/data/seedData.py
@@ -222,7 +222,6 @@ def seedData():
                          tenantID=tenant_renty_mcrenter.id,
                          dateTimeStart=now,
                          dateTimeEnd=future,
-                         dateUpdated=now,
                          occupants=3)
     lease_1.save_to_db()
     lease_2 = LeaseModel(name="Lease 2",
@@ -230,7 +229,6 @@ def seedData():
                          tenantID=tenant_soho_muless.id,
                          dateTimeStart=now,
                          dateTimeEnd=future,
-                         dateUpdated=now,
                          occupants=2)
     lease_2.save_to_db()
     lease_3 = LeaseModel(name="Lease 3",
@@ -238,7 +236,6 @@ def seedData():
                          tenantID=tenant_starvin_artist.id,
                          dateTimeStart=now,
                          dateTimeEnd=future,
-                         dateUpdated=now,
                          occupants=1)
     lease_3.save_to_db()
 
--- a/models/base_model.py
+++ b/models/base_model.py
@@ -30,11 +30,8 @@ class BaseModel(db.Model):
         db.session.delete(self)
         db.session.commit()
 
-    def json(self):
-        raise NotImplementedError(f"This {self.__class__} must implement a <json> method")
-
     def __repr__(self):
-        return f'<{type(self).__name__} {self.json()}>'
+        return "<{!r} {!r}>".format(type(self).__name__, self.__dict__)
 
     @classmethod
     def _name(cls):
--- a/models/base_model.py
+++ b/models/base_model.py
@@ -34,6 +34,22 @@ class BaseModel(db.Model):
         db.session.add(cls(**attrs))
         db.session.commit()
 
+    @classmethod
+    def update(cls, schema, id, attributes):
+        obj = cls.find(id)
+        try:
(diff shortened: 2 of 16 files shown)


jls-github (commented):
The controller looks so much cleaner and more readable. I very much appreciate that we're moving the logic out of there.

sokampdx:
So I saw a few unused import:
models/property.py:1
tests/integration/test_leases.py:2
tests/unit/test_lease.py:1,5,6

sokampdx:
I like how you have a generic create and update in the base_model which made the resource/lease.py much cleaner.  At the same time, we added a lot of indirection when we abstracted away the return json. I also trying to understand the purpose for the added folder structure in the serializers.

Have you run into these stackoverflow questions regarding auto generated schema?
https://stackoverflow.com/questions/42891152/how-to-dynamically-generate-marshmallow-schemas-for-sqlalchemy-models
https://stackoverflow.com/questions/63697476/generating-marshmallow-schema-automatically-with-json-serializable-enums

NickSchimek:
Thanks for the feedback @sokampdx - 
I removed all the unused imports you pointed. Thanks for that!
> I also trying to understand the purpose for the added folder structure in the serializers

I'm not completely sold on this. As I mentioned earlier I may have abstracted too soon. I did this because I've worked with this structure in the past and I found it very convenient. However with Marshmallow it might be better to build different schemas for a resource rather than relying on the different options that can be passed in, but I'm not sure yet, only time will tell. 
The main reason why I created the serializer file is because the schema can take a lot of options upon instantiation that affects its behavior. So yes, I leave that up to another method so that the serializer can be tested and relied upon independently from the calling method.

> At the same time, we added a lot of indirection when we abstracted away the return json

I wouldn't say there is a lot of indirection. There is one level of indirection. The serializer is called which calls the schema.

With the JSON method there was list comprehensions, and sometimes list comprehensions within list comprehensions. We could add the json method back in and call the schema from there instead of having the serializer file doing that. - It would still be one level of indirection, but at least we could test the response separately from the action. I'm not opposed to doing it that way at this point in time. I think it will become more clear what should be done once we start figuring out how to deal with authorizations for our resources to ensure that only authorized data is sent for the request. Especially regarding the property manager role. We need to ensure that we do not send back more information than absolutely necessary when a property manager hits an endpoint.

So yes, we could move the code in the serializer method to a json method on the object instance. Is that what you have in mind or are you thinking of something else?

NickSchimek:
Regarding the stack overflow questions. I had thought about dynamically generating the schemas. It could be done. However, we also want to use the schemas for validating input which may lead to additional entries in the schema that is not common to other schemas. I think it is best to hold off on doing that until we have more information on how we're using the schemas.

htharker42 (approved):
gtg

PR #118, opened 2020-10-01
Add timestamps to all models

### LINKED ISSUE #346: Create timestamp columns for all db tables
## Description
This feature is a sub-task of issue #310 

The end goal is to create timestamps for all existing and future db objects so that each object automatically has access to them, if needed, without having to write any implementation logic.

## Acceptance criteria:
 - [ ] Every db should have a `created_at` and `updated_at` column
 - [ ] The new columns should be a required `Date` field
 - [ ] The `created_at` timestamps should be updated at object creation
 - [ ] The `updated_at` timestamp should be updated each time a db transaction is performed

**Please make use of the existing base_model as much as possible to reduce code duplication.**

## Stretch goal (not necessary, but would be nice)
 - [ ] Create a test (possibly using a double) to check the timestamps work as expected.
 - [ ] Ideally when a new object is created the `created_at` and `updated_at` timestamp should be exactly the same. Ensure timestamps have same datetime for new records, and only the updated_at is updated for existing records.

## Scope definition
Editing any existing timestamps on any of the db models is out of the scope of this issue. This will be handled in another issue.

NickSchimek:

This happened to be easier than I thought it would.

Closes codeforpdx/dwellingly-app/issues/346

### DIFF
--- a/models/base_model.py
+++ b/models/base_model.py
@@ -1,8 +1,12 @@
 from db import db
+from datetime import datetime
 
 
 class BaseModel(db.Model):
     __abstract__ = True
+
+    created_at = db.Column(db.DateTime, nullable=False, default=datetime.utcnow)
+    updated_at = db.Column(db.DateTime, onupdate=datetime.utcnow)
 
     @classmethod
     def find_by_id(cls, id):


htharker42 (approved):
looks good, we will need to update existing models to removed redundant code.

PR #120, opened 2020-10-01
audited and harmonized return messages

### LINKED ISSUE #332: Audit and update json response keys to all use the same convention
## Description

In some cases we return json responses with the a few different names for the same key.

For example in some cases we return 'msg', 'message', and sometimes 'Message'

I believe we should choose one style and use the same convention across the app. My reasoning for doing this is that I would assume it would make it easier for the front end developers to extract the value from the key, and decrease bugs in potential misspellings. 

It appears the reason why we have different keys is that the pyjwt library returns 'msg' for all of its messages, and the backend developers are returning 'message' for all other messages. 

If we want to go with 'message' then pyjwt has an option for that via a custom config.

An update to the README describing conventions used in the app would also be helpful.

Tasks to be done:

- [ ] Team to decide on convention
- [ ] Make necessary changes in backend
- [ ] Make necessary changes in frontend
- [ ] Update readme

yunnie:

Closes codeforpdx/dwellingly-app/issues/332

Sorry, but this change hits a lot of files. Changed the various version of "message", "Message", "msg" to "message". Made return message format more consistent, for example changed "Admin Access Required" to "Admin access required", and "Missing Authorization Header" to "Missing authorization header". This should make the error messages seem more consistent. 


models/emergency_contact.py
```diff
@@ -32,7 +32,7 @@ def json(self):
             'id': self.id,
             'name':self.name,
             'description': self.description,
-            'contact_numbers': [number.json() for number in self.contact_numbers]
+            'contactNumbers': [number.json() for number in self.contact_numbers]
```
NickSchimek: This could impact the FE. I think they like receiving the json keys in camelCase though. I know they're still working on integrating a lot of the pages so it may not have any impact at all.

NickSchimek:
I like the PR title. "Harmonized" 😄 It shouldn't be an issue that the PR touches a lot of files, although I will have probably have to rebase all my active PR's 😢 unless they get merged before this 🐎 

At first glance the PR looks good. I have one comment where I noticed the JSON response format changed. I'm not sure if it is an issue or not. Is it?

NickSchimek:
I updated the PR description so that it will link and close the related issue when merged

models/emergency_contact.py
yunnie: All of the other json keys use camel case, so it would be more consistent. But if FE developers have already realized that contact_numbers is the exception, then it makes sense to just keep it as contact_numbers.

yunnie:
The error messages themselves are not a problem in their current state. Some use title case, most use capitalized first word. It's really just a matter of consistent style to give the application a more unified feel. And if the error messages are never displayed, then the inconsistency in the style doesn't really matter.

NickSchimek:
> It's really just a matter of consistent style to give the application a more unified feel.

Yeah, the inconsistency really bothers me. I'm working to abstract this away so that it wont be a problem anymore.

models/emergency_contact.py
NickSchimek: Personally I prefer snake case. :-) For the json response perhaps camel case is preferred by the FE. I noticed in some other places we are doing some weird stuff with the json responses, and after speaking with @Chris-Boe I'm not to worried about breaking the FE, however we still need to take note of it so that people who work on the FE don't waste time debugging. 

It looks like they're dependent on `contact_numbers` in two files:

- codeforpdx/dwellingly-app/src/views/emergencyContacts.js
- codeforpdx/dwellingly-app/src/views/addEmergencyContact.js

NickSchimek: We can either revert this and ignore the inconsistency in style, or get an issue created and/or a PR submitted to update the FE.

yunnie: I'll revert back to the snake case. If we feel that consistency is really an issue, we can create another issue.

NickSchimek: OK. sounds good.

resources/user.py
```diff
@@ -149,31 +149,31 @@ def post(self, user_id):
...
+    parser.add_argument('password', type=str, required=True, help="This field cannot be blank")
 
     def post(self):
         data = UserLogin.parser.parse_args()
 
         user = UserModel.find_by_email(data['email'])
 
         if user and user.archived:
-            return {"message": "Not a valid user"}, 403
+            return {"message": "Invalid user"}, 403
```
NickSchimek, line 161: Note. I think if a user is archived then perhaps a "User not found" response is more appropriate. Simply making a note of this. (No action needed)

tests/integration/test_emergency_contacts.py
```diff
@@ -47,16 +48,22 @@ def test_emergency_contacts_POST(client, auth_headers):
   }
 
   response = client.post(endpoint, json=newContact, headers=auth_headers["admin"])
-  assert is_valid(response, 401) # UNAUTHORIZED - An emergency contact with this name already exists
+  assert is_valid(response, 401) # UNAUTHORIZED - Emergency Contact With This Name Already Exists
+  assert response.json == \
+          {'message':
+           'An emergency contact with this name already exists'}
```
NickSchimek, line 54: This sounds like a validation issue rather than an authorization issue. Created issue codeforpdx/dwellingly-app/issues/374 to address this

tests/integration/test_emergency_contacts.py
```diff
@@ -47,16 +48,22 @@ def test_emergency_contacts_POST(client, auth_headers):
...
   assert is_valid(response, 401) # UNAUTHORIZED - Admin Access Required
+  assert response.json == {'message': 'Admin access required'}
 
   response = client.post(endpoint, json=newContact, headers=auth_headers["pending"]) 
   assert is_valid(response, 401) # UNAUTHORIZED - Admin Access Required
+  assert response.json == {'message': 'Admin access required'}
 
   response = client.post(endpoint, json=newContact)
   assert is_valid(response, 401) # UNAUTHORIZED - Missing Authorization Header
+  assert response.json == {'message': 'Missing authorization header'}
```
NickSchimek, line 66: These extra tests don't really add much value. I will be talking about this topic in the next meeting, and how the test suite is being refactored to avoid redundant tests. They're fine for now, however they will most likely be deleted in the coming months.

NickSchimek (dismissed):
Thanks for doing this!

NickSchimek:
@yunnie The PR looks good. Are you comfortable resolving the conflicts?

yunnie:
@NickSchimek I haven't resolved conflicts before with git, but I'm happy to give it a try. Let me know if you're OK with that.

NickSchimek:
Yeah, you can’t harm anything by giving it a try. If you get stuck then
please reach out to me, or I can also walk you through it too.

NickSchimek (dismissed):

NickSchimek (approved):
Looks good. Thank you!

PR #123, opened 2020-10-06
Install Marshmallow and created a factory module

### LINKED ISSUE #373: Install Marshmallow
## Description

Flask-Restful's request parser is being deprecated, and will not be replaced. Marshmallow is a popular choice for validating input and serializing/deserializing data. 

Marshmallow can return detailed validation errors, and allows us to build custom validations.
The serializer supports supports scoped queries as input and can serialize all relationships for the objects.

Compared to the other libraries I researched. Marshmallow is not only the most widely used, but has some very convenient features in regards to validating input data that the others lacked. And flask-marshmallow has a thin wrapper around marshmallow that allows to automatically generate schemas for validating and serializing data based off our db column specified in the models.

NickSchimek:
## Description

Implemented marshmallow and implemented some basic auto schemas for the lease, tenant, and user. Marshmallow can be used for input validation and serialization/deserialization. It can take many options when serializing, and so testing the expected output is necessary. Implemented a design that will allow any number of unique serializers (tested serializers) to be created and invoked while keeping each individual file small.

Created a factory module that automatically includes any factories defined into the conftest so that all tests will have access to the factories defined within that module. This is ultimately going to be replacing the seed data file that is currently being used for tests. <-- @jls-github Here is the answer to one of your questions you asked recently.

Also took the liberty to remove the landlordID from the lease Table. It is not being used in the FE or the BE, and I don't see a logical reason to have it. A Lease belongs to a property, and a property has many property managers. Therefore I'm not sure what the significance of a landlordID on a lease would be. If we need to have signatories on a. lease in the future then that may be reason to add it back in, but perhaps under a more applicable name.

This is the basic structure for an implementation. It is not being used yet. If this is approved then I will start to integrate it in PR #117 

Closes codeforpdx/dwellingly-app/issues/373

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+.DS_Store
 # Byte-compiled / optimized / DLL files
 __pycache__/
 *.py[cod]
--- a/Pipfile
+++ b/Pipfile
@@ -37,6 +37,8 @@ pytest = "*"
 "zope.interface" = "*"
 freezegun = "*"
 bcrypt = "==3.1.7"
+flask-marshmallow = "*"
+marshmallow-sqlalchemy = "*"
 
 [requires]
 python_version = "3.8"
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "a59da4beaaa74c5ac012f8883a77bfcbd763b119a602f65276e6e45cea306a4b"
+            "sha256": "abd5adc7e3c37aee9151eb37578e4f5603de0f0681f23fb523c107abbf7da991"
         },
         "pipfile-spec": 6,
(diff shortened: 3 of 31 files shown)


jls-github (approved):
Thanks for answering my question about the schema, @NickSchimek

htharker42 (approved):
I am pretty sure we can get rid of the landlordID, we might have to update the Properties model if retrieving it quickly is an issue.

NickSchimek:
OK. Let's see what the requirements will be and work off of that. Currently we're not doing any eager loading anywhere, if that's what you mean. It's something that is on my mind to watch and tackle in the future. Unfortunately Flask-sqlalchemy does not make it easy, but I think we can abstract it to make it easy, and by easy I mean less implementation details sprinkled everywhere.

PR #125, opened 2020-10-08
Cp 375 bare except

### LINKED ISSUE #375: Remove all bare except statements across the app
## Description

There should not be any bare except statements in the app for many reasons. Here are two:

- Bare except statements catch all exceptions even ones that a dev should not usually be catching
- Flask has built-in error handling that will return an appropriate response when something bad happens

There are a few cases across the app where there is a bare except clause that is returning a 500 response code.
Simply remove all of these. see codeforpdx/dwellinglybackend/pull/121 as an example of what should be done.

cphpitts:

Closes codeforpdx/dwellingly-app/issues/375

- [X ] Code builds successfully

### DIFF
--- a/resources/emergency_contacts.py
+++ b/resources/emergency_contacts.py
@@ -52,10 +52,9 @@ class EmergencyContacts(Resource):
         data["contact_numbers"] = numbersData
 
         contactEntry = EmergencyContactModel(**data)
-        try:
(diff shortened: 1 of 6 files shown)


NickSchimek (approved):
Awesome! Thanks for doing this.

PR #126, opened 2020-10-09
Replace addedOn by created_at

### LINKED ISSUE #380: Update tenant JSON response
## Description

Add the `created_at` field to the JSON response in the tenant model. It should return the same time format as the `addedOn` field.

The `created_at` field is already defined in the base model. The `addedOn` field should be removed in both the JSON response and the Table declaration if it is not being utilized in the front-end.

**Purpose**

The created_at field will eventually replace the addedOn field.

**Acceptance criteria**

 - The tenant JSON response is returning the `created_at` timestamp

Related to issue #310

Chris-Boe:
Closes codeforpdx/dwellingly-app/issues/380

Is a re-seeding of the database necessary? nope
Any new dependencies to install? nope
Any special requirements to test? nope, it's included in the /serializer/tenant/ test here which passes or you can look at the network tab in the front end on the /manage/tenants page to see it come through

### DIFF
--- a/models/tenant.py
+++ b/models/tenant.py
@@ -13,7 +13,7 @@ class TenantModel(BaseModel):
     lastName = db.Column(db.String(100))
     phone = db.Column(db.String(20))
     propertyID = db.Column(db.Integer, db.ForeignKey('properties.id'))
-    addedOn = db.Column(db.Date())
+    created_at = db.Column(db.Date())
     unitNum = db.Column(db.String(20))
 
     # relationships
@@ -31,7 +31,7 @@ class TenantModel(BaseModel):
             user = UserModel.find_by_id(id)
             if user: self.staff.append(user)
         self.unitNum = unitNum
-        self.addedOn = datetime.date(datetime.now())
+        self.created_at = datetime.date(datetime.now())
 
 
     def json(self):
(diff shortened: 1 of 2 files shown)


models/tenant.py
```diff
@@ -13,7 +13,7 @@ class TenantModel(BaseModel):
     lastName = db.Column(db.String(100))
     phone = db.Column(db.String(20))
     propertyID = db.Column(db.Integer, db.ForeignKey('properties.id'))
-    addedOn = db.Column(db.Date())
+    created_at = db.Column(db.Date())
```
NickSchimek: This should not be needed. `created_at` is already defined for all models in the db on the base class.

models/tenant.py
```diff
@@ -31,7 +31,7 @@ def __init__(self, firstName, lastName, phone, propertyID, staffIDs, unitNum):
             user = UserModel.find_by_id(id)
             if user: self.staff.append(user)
         self.unitNum = unitNum
-        self.addedOn = datetime.date(datetime.now())
+        self.created_at = datetime.date(datetime.now())
```
NickSchimek: This should also not be needed. `created_at` defaults to the current UTC time when the object is created.

NickSchimek (changes requested):
Thanks for doing this @Chris-Boe - Lines 16 and 34 are probably safe to delete as they re-implement what the base model already has setup. https://github.com/codeforpdx/dwellinglybackend/blob/development/models/base_model.py#L5

NickSchimek (approved):
Looks good! Thanks

PR #127, opened 2020-10-10
Added unique constraint @ db level to User email

### LINKED ISSUE #333: Audit and set db level constraints
## Description
Audit and set db level constraints as necessary.

For example. User's cannot have the same email address. However, the current setup allows for the same email address to be saved in the database. Regardless if there is validation in the backend for this. There should also also be one set at the database level. Not every validation needs to be set at db level, but in the case of unique fields it is good practice to have that set at the db level.

yunnie:

Closes codeforpdx/dwellingly-app/issues/333 

Attempting to add a duplicate email in the User model now results in an exception. I didn't see any other field that required a constraint. 

### DIFF
--- a/models/user.py
+++ b/models/user.py
@@ -20,7 +20,7 @@ class UserModel(BaseModel):
     __tablename__ = 'users'
 
     id = db.Column(db.Integer, primary_key=True)
-    email = db.Column(db.String(100), unique=True)
+    email = db.Column(db.String(100), unique=True, nullable=False)
     role = db.Column(db.Enum(RoleEnum), default=RoleEnum.PENDING)
     firstName = db.Column(db.String(80))
     lastName = db.Column(db.String(80))
--- a/models/user.py
+++ b/models/user.py
@@ -20,7 +20,7 @@ class UserModel(BaseModel):
     __tablename__ = 'users'
 
     id = db.Column(db.Integer, primary_key=True)
-    email = db.Column(db.String(100))
+    email = db.Column(db.String(100), unique=True)
     role = db.Column(db.Enum(RoleEnum), default=RoleEnum.PENDING)
     firstName = db.Column(db.String(80))
     lastName = db.Column(db.String(80))
--- a/tests/integration/test_users.py
+++ b/tests/integration/test_users.py
@@ -3,6 +3,7 @@ from conftest import is_valid, log
 from freezegun import freeze_time
 from models.user import RoleEnum
 from flask_jwt_extended import create_access_token, create_refresh_token
+import pytest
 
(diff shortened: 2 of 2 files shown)


NickSchimek:
@yunnie Can you update the PR description so that it automatically closes the issue? Adding the extra brackets, parens and link does not work to automatically close the PR.

tests/integration/test_users.py
```diff
@@ -200,7 +201,13 @@ def test_patch_user(client, auth_headers, new_user):
...
             {'message': 'Only admins can change roles'}
 
-
+def test_unique_user_constraint(client, auth_headers, new_user):
+    """Emails must be unique, otherwise an Exception is thrown"""
+    with pytest.raises(Exception) as e_info:
+        userToPatch = UserModel.find_by_email(new_user.email)
+        response = client.patch(f"/api/user/{userToPatch.id}",
+                           json = {"email": "user1@dwellingly.org"},
+                           headers=auth_headers["admin"])
```
NickSchimek, line 210: Writing validation tests will hopefully be a lot simpler with marshmallow. Usually I wouldn't test this type of database configuration, because the constraint would be repeated on the application side, and I would test that. This test is good in that it highlights we will send an uncaught 500 internal server error to the frontend. Do you think we should add the application side validation check now to mitigate this or should we wait and do it with the future refactorings?

models/user.py
```diff
@@ -20,7 +20,7 @@ class UserModel(BaseModel):
     __tablename__ = 'users'
 
     id = db.Column(db.Integer, primary_key=True)
-    email = db.Column(db.String(100))
+    email = db.Column(db.String(100), unique=True)
```
NickSchimek: should probably also add this too:
```suggestion
    email = db.Column(db.String(100), unique=True, nullable=False)
```

NickSchimek (dismissed):
I was a bit vague in that issue I wrote up. There are probably some more constraints we could add here, but I'm not sure right now. Thanks for doing this. This looks good. We could add the app side validation check now, however we will eventually do that with marshmallow so it can wait too.

tests/integration/test_users.py
yunnie: If the uncaught exception is going to cause problems for the FE development, then we should add the application side validation check now. For example, if the FE expects a failure status code, or a failure message in cases where the email is a duplicate, then leaving the exception will be a problem. Otherwise, I think we can wait until we refactor.

NickSchimek: I don't think the FE is handling that scenario. It can probably wait without any issues.

NickSchimek (approved):

PR #130, opened 2020-10-18
Issue310 Use updated_at and created_at throughout backend

### LINKED ISSUE #310: Use created_at and updated_at timestamps for all db entries
## Description

A couple of the db tables have their own variation of updated_at and created_at timestamps. It would be less prone to error to uniformly apply these timestamps to all tables, and update them behind the scenes so that the individual models/controllers are not responsible for implementing logic that should be able to be abstracted away.

Terminology such as `created_at` and `updated_at` could be used across all Tables. This would mean renaming the few columns that are already implemented, and would possibly require changes on the frontend. I'm open to using other terminology, but it should be consistent across all Tables so that we can easily abstract the behavior away from the business logic.

This issue will be broken down into several tasks:

- [x] Create timestamp columns for all db tables. Issue #346
- [ ] Return timestamps in json responses where appropriate
- [ ] Update Frontend to use new timestamps
- [ ] Remove old timestamps in BE

Chris-Boe:
Closes codeforpdx/dwellingly-app/issues/310

Is a re-seeding of the database necessary? Not really
Any new dependencies to install? Nope
Any special requirements to test? Nope

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -27,6 +27,7 @@ psycopg2-binary = "*"
 bcrypt = "==3.1.7"
 flask-marshmallow = "*"
 marshmallow-sqlalchemy = "*"
+atomicwrites = "*"
 
 [requires]
 python_version = "3.8"
--- a/Pipfile
+++ b/Pipfile
@@ -12,6 +12,7 @@ freezegun = "*"
 pytest-watch = "*"
 python-dotenv = "*"
 pytest = "*"
+atomicwrites = "*"
 
 [packages]
 Flask = "==1.1.1"
@@ -27,7 +28,6 @@ psycopg2-binary = "*"
 bcrypt = "==3.1.7"
 flask-marshmallow = "*"
 marshmallow-sqlalchemy = "*"
-atomicwrites = "*"
 
 [requires]
 python_version = "3.8"
--- a/models/base_model.py
+++ b/models/base_model.py
@@ -8,7 +8,7 @@ class BaseModel(db.Model):
     __abstract__ = True
 
     created_at = db.Column(db.DateTime, nullable=False, default=datetime.utcnow)
-    updated_at = db.Column(db.DateTime, onupdate=datetime.utcnow)
+    updated_at = db.Column(db.DateTime, onupdate=datetime.utcnow, default=datetime.utcnow, nullable=False)
 
     @classmethod
     def find_by_id(cls, id):
--- a/models/lease.py
+++ b/models/lease.py
@@ -12,5 +12,5 @@ class LeaseModel(BaseModel):
     propertyID = db.Column(db.Integer, db.ForeignKey('properties.id'), nullable=False)
     tenantID = db.Column(db.Integer, db.ForeignKey('tenants.id'), nullable=False)
     occupants = db.Column(db.Integer)
-    dateTimeStart = db.Column(db.DateTime, nullable=False, default=datetime.utcnow)
-    dateTimeEnd = db.Column(db.DateTime, nullable=False, default=datetime.utcnow)
+    dateTimeStart = db.Column(db.DateTime, nullable=False)
+    dateTimeEnd = db.Column(db.DateTime, nullable=False)
--- a/models/lease.py
+++ b/models/lease.py
@@ -1,5 +1,6 @@
 from db import db
 from models.base_model import BaseModel
+from utils.time import Time
 
 
 class LeaseModel(BaseModel):
@@ -10,5 +11,5 @@ class LeaseModel(BaseModel):
     propertyID = db.Column(db.Integer, db.ForeignKey('properties.id'), nullable=False)
     tenantID = db.Column(db.Integer, db.ForeignKey('tenants.id'), nullable=False)
     occupants = db.Column(db.Integer)
-    dateTimeStart = db.Column(db.DateTime, nullable=False)
-    dateTimeEnd = db.Column(db.DateTime, nullable=False)
+    dateTimeStart = db.Column(db.DateTime, nullable=False, default=datetime.datetime.utcnow)
+    dateTimeEnd = db.Column(db.DateTime, nullable=False, default=datetime.datetime.utcnow)
--- a/models/lease.py
+++ b/models/lease.py
@@ -1,6 +1,7 @@
 from db import db
 from models.base_model import BaseModel
 from utils.time import Time
+from datetime import datetime
 
 
 class LeaseModel(BaseModel):
@@ -11,5 +12,5 @@ class LeaseModel(BaseModel):
     propertyID = db.Column(db.Integer, db.ForeignKey('properties.id'), nullable=False)
     tenantID = db.Column(db.Integer, db.ForeignKey('tenants.id'), nullable=False)
     occupants = db.Column(db.Integer)
-    dateTimeStart = db.Column(db.DateTime, nullable=False, default=datetime.datetime.utcnow)
-    dateTimeEnd = db.Column(db.DateTime, nullable=False, default=datetime.datetime.utcnow)
+    dateTimeStart = db.Column(db.DateTime, nullable=False, default=datetime.utcnow)
+    dateTimeEnd = db.Column(db.DateTime, nullable=False, default=datetime.utcnow)
--- a/models/tickets.py
+++ b/models/tickets.py
@@ -6,6 +6,7 @@ from models.notes import NotesModel
 from models.tenant import TenantModel
 from datetime import datetime, timedelta
 from models.base_model import BaseModel
+from utils.time import Time
 
 
 class TicketModel(BaseModel):
@@ -67,8 +68,8 @@ class TicketModel(BaseModel):
             'minsPastUpdate': minsPastUpdate,
             'urgency': self.urgency,
             'notes': message_notes,
-            'created_at': self.created_at.strftime("%m/%d/%Y %H:%M:%S") if self.created_at else None,
-            'updated_at': self.updated_at.strftime("%m/%d/%Y %H:%M:%S") if self.updated_at else None
+            'created_at': Time.format_date(self.created_at) if self.created_at else None,
+            'updated_at': Time.format_date(self.updated_at) if self.updated_at else None
         }
 
     @classmethod
--- a/models/tickets.py
+++ b/models/tickets.py
@@ -68,8 +68,8 @@ class TicketModel(BaseModel):
             'minsPastUpdate': minsPastUpdate,
             'urgency': self.urgency,
             'notes': message_notes,
-            'created_at': Time.format_date(self.created_at) if self.created_at else None,
-            'updated_at': Time.format_date(self.updated_at) if self.updated_at else None
+            'created_at': Time.format_date(self.created_at),
+            'updated_at': Time.format_date(self.updated_at)
         }
 
     @classmethod
--- a/models/tickets.py
+++ b/models/tickets.py
@@ -48,12 +48,7 @@ class TicketModel(BaseModel):
 
         assignedUserData = UserModel.find_by_id(self.assignedUser)
         assignedUser = "{} {}".format(assignedUserData.firstName, assignedUserData.lastName)
-
-        if self.updated_at is None:
-            minsPastUpdate = 0
-        else:
-            dateTimeNow = datetime.utcnow()
-            minsPastUpdate = int((datetime.utcnow() - self.updated_at).total_seconds() / 60)
+        minsPastUpdate = int((datetime.utcnow() - self.updated_at).total_seconds() / 60)
         
         return {
             'id': self.id,
--- a/models/tickets.py
+++ b/models/tickets.py
@@ -52,8 +52,8 @@ class TicketModel(BaseModel):
         if self.updated_at is None:
             minsPastUpdate = 0
         else:
-            dateTimeNow = datetime.now()
-            minsPastUpdate = int((datetime.now() - self.updated_at).total_seconds() / 60)
+            dateTimeNow = datetime.utcnow()
+            minsPastUpdate = int((datetime.utcnow() - self.updated_at).total_seconds() / 60)
         
         return {
             'id': self.id,
@@ -79,7 +79,7 @@ class TicketModel(BaseModel):
     @classmethod
     def find_count_by_age_status(cls, status, minutes):
         #calculated in minutes: 1 day = 1440, 1 week = 10080
-        dateTime = datetime.now() - timedelta(minutes = minutes)
+        dateTime = datetime.utcnow() - timedelta(minutes = minutes)
         return db.session.query(TicketModel).filter(TicketModel.updated_at >= dateTime).filter(TicketModel.status == status).count()
     
     @classmethod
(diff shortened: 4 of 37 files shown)


Chris-Boe:
Relates to https://github.com/codeforpdx/dwellingly-app/pull/393

Pipfile
```diff
@@ -27,6 +27,7 @@ psycopg2-binary = "*"
 bcrypt = "==3.1.7"
 flask-marshmallow = "*"
 marshmallow-sqlalchemy = "*"
+atomicwrites = "*"
```
NickSchimek: What is this being used for?

models/tickets.py
```diff
@@ -80,7 +80,7 @@ def find_count_by_status(cls, status):
     def find_count_by_age_status(cls, status, minutes):
         #calculated in minutes: 1 day = 1440, 1 week = 10080
         dateTime = datetime.now() - timedelta(minutes = minutes)
-        return db.session.query(TicketModel).filter(TicketModel.updated >= dateTime).filter(TicketModel.status == status).count()
+        return db.session.query(TicketModel).filter(TicketModel.updated_at >= dateTime).filter(TicketModel.status == status).count()
```
NickSchimek, line 80: Do you know what this method is for? I'm curious how we are using it.

utils/time.py
```diff
@@ -1,11 +1,16 @@
...
 
 class Time:
     @staticmethod
     def format_date(date):
-        return date.strftime("%m/%d/%Y %H:%M:%S") if date else None
+        return date.strftime(time_format) if date else None
+    
+    @staticmethod
+    def serialized_date_format(date):
+        return date.strftime("%Y-%m-%dT%H:%M:%S.%f")
```
NickSchimek: Perhaps we can do this, and remove all of the conditional logic that is wrapped around calling this method?
```suggestion
        return date.strftime("%Y-%m-%dT%H:%M:%S.%f") if date else None
```

tests/serializers/tenant/test_tenant_serializer.py
```diff
@@ -9,11 +9,11 @@ def test_serializer(self, create_tenant):
...
-                'id': tenant.id,
                 'firstName':tenant.firstName,
+                'id': tenant.id,
                 'lastName':tenant.lastName,
                 'phone': tenant.phone,
                 'unitNum': tenant.unitNum,
-                'created_at': Time.format_date(tenant.created_at),
-                'updated_at': Time.format_date(tenant.updated_at)
+                'created_at': Time.serialized_date_format(tenant.created_at),
+                'updated_at': None
```
NickSchimek:
```suggestion
                'updated_at': Time.serialized_date_format(tenant.updated_at)
```

tests/serializers/property/test_property_serializer.py
```diff
@@ -16,8 +16,7 @@ def test_serializer(self, create_property):
             'city': property.city,
             'state': property.state,
             'zipcode': property.zipcode,
-            'dateAdded': property.dateAdded,
             'archived': property.archived,
-            'created_at': Time.format_date(property.created_at),
-            'updated_at': Time.format_date(property.updated_at)
+            'created_at': Time.serialized_date_format(property.created_at),
+            'updated_at': None
```
NickSchimek:
```suggestion
            'updated_at': Time.serialized_date_format(property.updated_at)
```

tests/serializers/lease/test_lease_serializer.py
```diff
@@ -13,8 +13,7 @@ def test_serializer(self, create_lease):
                 'name': lease.name,
                 'dateTimeStart': Time.format_date(lease.dateTimeStart),
                 'dateTimeEnd': Time.format_date(lease.dateTimeEnd),
-                'dateUpdated': Time.format_date(lease.dateUpdated),
+                'updated_at': None,
```
NickSchimek: I'm not sure if we will always want to keep updated_at defaulted to None.
None is technically an implementation detail. Perhaps we can do this instead:
```suggestion
                'updated_at': Time.serialized_date_format(lease.updated_at),
```

NickSchimek (changes requested):
Very nice! Thanks for doing this. A couple small suggestions and questions. 

It looks like `serialized_time_format`, the default db structure, will be used for sending date/time to the frontend. And `time_format` will be used for deserializing the date/time received from the frontend?

Pipfile
Chris-Boe: I wasn't able to run the tests until I manually installed this, since it wasn't in the pipfile

models/tickets.py
Chris-Boe: Looks like find_count_by_age_status is used for the dashboard widgets

tests/serializers/lease/test_lease_serializer.py
Chris-Boe: > Very nice! Thanks for doing this. A couple small suggestions and questions.
> 
> It looks like `serialized_time_format`, the default db structure, will be used for sending date/time to the frontend. And `time_format` will be used for deserializing the date/time received from the frontend?

Not quite, that serialized_date_format method is just for the serializer tests that are doing a "dump" of the object from the db, while our resources are using the object model's json() method to explicitly send back the different properties to the frontend. Ex from issue 390 on the frontend: (emergency contacts)
image

tests/serializers/property/test_property_serializer.py
Chris-Boe: "None" is just for these tests because we are immediately creating the object and then serializing it, which means there won't be an updated_at value since there isn't a default for it.

tests/serializers/tenant/test_tenant_serializer.py
Chris-Boe: "None" is just for these tests because we are immediately creating the object and then serializing it, which means there won't be an updated_at value since there isn't a default for it.

utils/time.py
Chris-Boe: Oh duh! Good idea

Pipfile
NickSchimek: hmmm. that seems odd. Perhaps we can move this to the dev/test packages then?

models/tickets.py
NickSchimek: OK. I will ask Hugh about this the next time we meet. Looking at this method I don't think it works as intended, and there are no tests to refer to.

tests/serializers/lease/test_lease_serializer.py
NickSchimek: As of yet the serializer is not hooked up to anything. Prior to this PR the serializer behavior matched the json() behavior. However, this PR is suggesting we change the behavior, and I'm trying to understand why.

From what your saying, It sounds like the serializer behavior should be reverted back to match the behavior of the json() method, because, regardless if we use the serializer module or not, the "dump" of the object from the db is what the FE will be receiving.

The output of the serializer should match the output of the json(); otherwise it is going to cause confusion going forward.

Chris-Boe: This PR isn't changing the behavior of the serializer, it's updating the tests to match the current behavior. We can talk about it tomorrow

NickSchimek: > This PR isn't changing the behavior of the serializer,

If the behavior was not changed then the tests would not need to be changed.

The deleted lines containing the `created_at` and `updated_at` fields in the schema files changed the behavior.
Screen Shot 2020-10-21 at 2 18 17 PM

Those lines specify which format to use when dumping the data. Was there another reason to delete those lines, if not to change the behavior?

Chris-Boe: Ahh okay, I was under the impression they were repetitive from the BaseModel already containing them. I can put them back! Or is there a way to just define the format in the BaseModel for all classes? I'm very new to flask/python in general 😉

NickSchimek: The Model's define the schema at the database level. These schemas are for marshmallow and define how we dump/load and validate data. As far as I know there is not a way to define that at the database level. That is not really the job of the database.

Chris-Boe: Interesting. I'll fix this tomorrow!

sokampdx:
We should create a corresponding frontend ticket that uses these fields 
e.g. 
- https://github.com/codeforpdx/dwellingly-app/blob/7529eeac94d8f66d5deb23f4b13ad7ee3df5ffd7/src/views/tickets.js#L37
- https://github.com/codeforpdx/dwellingly-app/blob/7529eeac94d8f66d5deb23f4b13ad7ee3df5ffd7/src/views/properties.js#L37

Pipfile
Chris-Boe: Sounds like it's a Windows specific required packages for the testing commands. I've moved it to the dev-packages section

Chris-Boe:
> We should create a corresponding frontend ticket that uses these fields
> e.g.
> 
> * https://github.com/codeforpdx/dwellingly-app/blob/7529eeac94d8f66d5deb23f4b13ad7ee3df5ffd7/src/views/tickets.js#L37
> * https://github.com/codeforpdx/dwellingly-app/blob/7529eeac94d8f66d5deb23f4b13ad7ee3df5ffd7/src/views/properties.js#L37

@sokampdx I already have a PR up that updates those fields https://github.com/codeforpdx/dwellingly-app/pull/393

As soon as other places have a need for those common dates, we can update the front end to use them

models/base_model.py
```diff
@@ -8,7 +8,7 @@ class BaseModel(db.Model):
     __abstract__ = True
 
     created_at = db.Column(db.DateTime, nullable=False, default=datetime.utcnow)
-    updated_at = db.Column(db.DateTime, onupdate=datetime.utcnow)
+    updated_at = db.Column(db.DateTime, onupdate=datetime.utcnow, default=datetime.utcnow, nullable=False)
```
NickSchimek, line 11: Yay! Thank you for doing this.

models/lease.py
```diff
@@ -10,5 +12,5 @@ class LeaseModel(BaseModel):
     propertyID = db.Column(db.Integer, db.ForeignKey('properties.id'), nullable=False)
     tenantID = db.Column(db.Integer, db.ForeignKey('tenants.id'), nullable=False)
     occupants = db.Column(db.Integer)
-    dateTimeStart = db.Column(db.DateTime, nullable=False)
-    dateTimeEnd = db.Column(db.DateTime, nullable=False)
+    dateTimeStart = db.Column(db.DateTime, nullable=False, default=datetime.utcnow)
+    dateTimeEnd = db.Column(db.DateTime, nullable=False, default=datetime.utcnow)
```
NickSchimek: I believe I recently removed these defaults after talking to Hugh about them. What is the reason for adding them back in? I believe dateTimeStart and dateTimeEnd should be required fields when creating a lease, when there is a default assigned then we are essentially saying they're optional fields. If the requirement is for them to be optional then this is fine; however I can't imagine that be the case.

models/tickets.py
```diff
@@ -52,8 +49,11 @@ def json(self):
         assignedUserData = UserModel.find_by_id(self.assignedUser)
         assignedUser = "{} {}".format(assignedUserData.firstName, assignedUserData.lastName)
 
-        dateTimeNow = datetime.now()
-        minsPastUpdate = int((datetime.now() - self.updated).total_seconds() / 60)
+        if self.updated_at is None:
+            minsPastUpdate = 0
```
NickSchimek: This branch will never trigger. Now that `updated_at` is defaulted to the time of initialization.

schemas/property.py
```diff
@@ -1,7 +1,7 @@
 from ma import ma
 from models.property import PropertyModel
-from schemas.time_format import time_format
 from marshmallow import fields
+from utils.time import time_format
 
 class PropertySchema(ma.SQLAlchemyAutoSchema):
```
NickSchimek: picky: pep8 calls for 2 blank lines above a class declaration

schemas/tenant.py
```diff
@@ -1,8 +1,7 @@
 from ma import ma
 from models.tenant import TenantModel
-from schemas.time_format import time_format
 from marshmallow import fields
-
+from utils.time import time_format
 
 class TenantSchema(ma.SQLAlchemyAutoSchema):
```
NickSchimek: picky: pep8 calls for 2 blank lines above a class declaration

tests/integration/test_base_model.py
```diff
@@ -52,7 +52,7 @@ def test_no_update(self):
 
         assert obj.first_name == 'Bye'
         assert obj.last_name == None
-        assert obj.updated_at == None
+        assert obj.updated_at != None
```
NickSchimek: This test can be deleted. It's not testing any behavior that is not already under test, and failed because it was testing an implementation setup.

utils/time.py
```diff
@@ -1,11 +1,12 @@
 from datetime import datetime
 from dateutil.relativedelta import relativedelta
 
+time_format = "%m/%d/%Y %H:%M:%S"
 
 class Time:
```
NickSchimek, line 6: This class is only used by tests, hence the reason for it being in the test folder. I don't think it's appropriate to move it into the application code. This change will not block merging though.

NickSchimek (commented):
@Chris-Boe This looks pretty good. I just have a few small comments

tests/integration/test_base_model.py
NickSchimek: Also if would be awesome if you could just delete `assert obj.updated_at != None` on lines 39 and 47 for me as well.

utils/time.py
Chris-Boe: Having all the time components in one place helps with the organization I think. Plus we can use that class in other places like the Widgets functions when we need to

tests/integration/test_base_model.py
Chris-Boe: Thanks for finding these!

models/lease.py
Chris-Boe: Oh you're right, I think I misheard during the meeting. I'll fix it!

models/lease.py
```diff
@@ -1,5 +1,7 @@
 from db import db
 from models.base_model import BaseModel
+from utils.time import Time
+from datetime import datetime
```
NickSchimek, line 4: Do we need these imports? I don't see them being used anywhere.

resources/lease.py
```diff
@@ -4,6 +4,7 @@
 from schemas.lease import LeaseSchema
 from serializers.lease import LeaseSerializer
 from flask_jwt_extended import jwt_required
+from utils.time import time_format
```
NickSchimek, line 7: I don't see this import being used anywhere.

resources/widgets.py
```diff
@@ -70,7 +70,7 @@ def get(self):
                     "desc": 'In Progress'
                 },
                 {
-                    "stat": TicketModel.find_count_by_age_status("In Progress", 10080),
+                    "stat": TicketModel.find_count_by_update_status("In Progress", 10080),
```
NickSchimek, line 73: Looks like we're tracking state with hard-code strings. FYI: Created issue codeforpdx/dwellingly-app/issues/404 to address this.

utils/time.py
NickSchimek: Then we should probably at some point create unit tests for the class.

NickSchimek (approved):
Thank you @Chris-Boe ! This looks great, thank you for working on this.

PR #131, opened 2020-10-21
Aks 370 manager property link

### LINKED ISSUE #370: Create many to many relationship
**Purpose**

Create a many to many relationship  between Properties table and the users table

Possibly name the new table `PropertyAssignments`

**Acceptance Criteria**

- [x] users do not have to be assigned a property
- [x] appropriate Tests added

- any other rules you an encounter while building this out that makes logical sense to have in here is probably OK to add.

related issue #369 and #353 
moved other criteria to #394

sokampdx:

Closes codeforpdx/dwellingly-app/issues/370

Is a re-seeding of the database necessary? Yes

### DIFF
--- a/models/property.py
+++ b/models/property.py
@@ -1,8 +1,11 @@
 from db import db
+from utils.time import Time
+from marshmallow import ValidationError
+
 from models.tenant import TenantModel
 from models.base_model import BaseModel
-from models.user import UserModel
-from utils.time import Time
+from models.user import UserModel, RoleEnum
+from models.property_assignment import PropertyAssignment
 
 
 class PropertyModel(BaseModel):
@@ -15,21 +18,21 @@ class PropertyModel(BaseModel):
     city = db.Column(db.String(50))
     state = db.Column(db.String(50))
     zipcode = db.Column(db.String(20))
-    propertyManager = db.Column(db.Integer(), db.ForeignKey('users.id'))
     archived = db.Column(db.Boolean)
 
     tenants = db.relationship(TenantModel, backref="property")
     leases = db.relationship('LeaseModel',
         backref='property', lazy=True, cascade="all, delete-orphan")
+    managers = db.relationship(UserModel, secondary='property_assignments', backref='properties')
 
-    def __init__(self, name, address, unit, city, state, zipcode, propertyManager, archived):
+    def __init__(self, name, address, unit, city, state, zipcode, propertyManagerIDs, archived):
         self.name = name
         self.address = address
         self.unit = unit
         self.city = city
         self.state = state
         self.zipcode = zipcode
-        self.propertyManager = propertyManager
+        self.managers = self.set_property_managers(propertyManagerIDs)
         self.archived = False
 
     def json(self):
@@ -37,7 +40,7 @@ class PropertyModel(BaseModel):
         for tenant in self.tenants:
             property_tenants.append(tenant.id)
 
-        property_manager = UserModel.find_by_id(self.propertyManager)
+        managers_name = [manager.fullName for manager in self.managers]
 
         return {
             'id': self.id,
@@ -47,8 +50,8 @@ class PropertyModel(BaseModel):
             'city': self.city,
             'state': self.state,
             'zipcode': self.zipcode,
-            'propertyManager': self.propertyManager,
-            'propertyManagerName': property_manager.full_name(),
+            'propertyManager': [user.json() for user in self.managers] if self.managers else None,
+            'propertyManagerName': managers_name if managers_name else None,
             'tenantIDs': property_tenants,
             'archived': self.archived,
             'created_at': Time.format_date(self.created_at),
@@ -61,4 +64,18 @@ class PropertyModel(BaseModel):
 
     @classmethod
     def find_by_manager(cls, manager_id):
-        return cls.query.filter_by(propertyManager=manager_id).all()
+        return cls.query.filter(cls.managers.any(UserModel.id == manager_id)).all()
+
+    @classmethod
+    def set_property_managers(cls, ids):
+        managers = []
+        if ids:
+            for id in ids:
+                user = UserModel.find_by_id(id)
+                if user and user.role == RoleEnum.PROPERTY_MANAGER:
+                    managers.append(user)
+                elif user and user.role != RoleEnum.PROPERTY_MANAGER:
+                    raise ValidationError(f'{user.fullName} is not a property manager')
+                else:
+                    raise ValidationError(f'{id} is not a valid user id')
+        return managers
--- a/tests/integration/test_properties.py
+++ b/tests/integration/test_properties.py
@@ -1,4 +1,5 @@
 from models.property import PropertyModel
+from models.user import UserModel
 import json
 
 def test_get_properties(client, test_database):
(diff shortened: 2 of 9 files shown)


models/property.py
```diff
@@ -61,4 +62,14 @@ def find_by_name(cls, name):
 
     @classmethod
     def find_by_manager(cls, manager_id):
-        return cls.query.filter_by(propertyManager=manager_id).all()
+        return cls.query.filter(cls.managers.any(UserModel.id == manager_id)).all()
+
+    @classmethod
+    def set_property_managers(cls, ids):
```
NickSchimek, line 70: Should we create a unit test for this method?

models/property.py
```diff
@@ -61,4 +62,14 @@ def find_by_name(cls, name):
...
+        managers = []
+        if ids:
+            for id in ids:
+                user = UserModel.find_by_id(id)
+                if user:
+                    managers.append(user)
```
NickSchimek: What should happen if a user does not exist?

models/user.py
```diff
@@ -32,14 +32,14 @@ class UserModel(BaseModel):
     created = db.Column(db.DateTime, default=datetime.utcnow)
 
 
-    def __init__(self, firstName, lastName, email, password, phone, role, archived):
+    def __init__(self, firstName, lastName, email, password, phone, role, archived=False):
```
NickSchimek: I think we can remove the parameter archive. Whenever a user is created then archived should always be set to false.

tests/integration/test_properties.py
```diff
@@ -26,19 +26,7 @@ def test_post_property(client, auth_headers, new_property):
...
-    assert property_info['address'] == '123 NE FLanders St'
-    assert property_info['unit'] == '5'
-    assert property_info['city'] == 'Portland'
-    assert property_info['state'] == 'OR'
-    assert property_info['zipcode'] == '97207'
-    assert property_info['propertyManager'] == 5
-    assert property_info['propertyManagerName'] == 'Gray Pouponn'
-    assert property_info['tenantIDs'] == [1]
-    assert property_info['dateAdded'] == '2020-04-12'
-    assert property_info['archived'] == 0
```
NickSchimek, line 40: what's the reason for deleting this?

tests/integration/test_users.py
```diff
@@ -85,7 +85,7 @@ def test_get_user_by_id(client, auth_headers, admin_user):
 
 def test_get_user_by_property_manager_id(client, auth_headers, new_property):
```
NickSchimek: This test seems redundant, and is possibly more focused on implementation rather than behavior. 
Finding a resource by an id should already be under test. I think it is safe to delete this test.

tests/unit/test_property.py
```diff
@@ -1,5 +1,35 @@
...
\ No newline at end of file
+name = "new_property"
+address = "789 Ten St."
+city = "Portland"
+unit = "101"
+state = "OR"
+zipcode = "97207"
+propertyManagerIDs = [5, 6]
+dateAdded = "2020-04-12"
+archived = 0
```
NickSchimek: We should use the boolean `True` value here rather than a truthy value.

tests/unit/test_property.py
```diff
@@ -1,5 +1,35 @@
...
+
+
+def test_new_property(test_database):
```
NickSchimek: FYI: This test doesn't really have much value. It's testing a plain python init function. We wouldn't even need to have the database if it weren't for the set_property_managers function. 

We could set the parameters to:
```
name = 0
address = 0
city = 0
unit = 0
state = 0
zipcode = 0
propertyManagerIDs = []
dateAdded = 0
archived = 0
```

And everything still passes. Hence, the implementation for the init function is being tested. We do need some unit tests for the property model. We can leave this in for now, and I will update the refactoring issues to add the appropriate unit tests on the model.

NickSchimek (commented):
Thanks for doing this @sokampdx - I have some comments/questions. I don't think anything I commented on would block this from being merged.

models/user.py
sokampdx: What happen if we want to create a archived user for test?

tests/integration/test_properties.py
sokampdx: I believe the assert statements are more like unit test like you mentioned in the lease model. We should use the serializer to test the json return value once in unit test later.

tests/integration/test_users.py
sokampdx: Yes. I can remove this test

tests/unit/test_property.py
sokampdx: I can remove this for now.

models/property.py
sokampdx: Currently, we return an empty manager list if none of the id exist. When the fe ticket that list manager in the create new property page is completed, then the ids are pre-validated as existing managers. We can restrict the role as well and then raise an exception for it as invalid manager id shouldn't be happening.

models/property.py
sokampdx: I can add a test for it.

models/property.py
NickSchimek: Thank you. I'm wondering what we should do if partial id's exist? Should we ignore the ones that don't and persist the ones that do, or raise an error?

models/user.py
NickSchimek: Personally, I'm generally against adding logic to make testing simpler. Although if testing is hard then that is a concern too. There is definitely a good reason to adjust application code to make testing simpler, however in this case I don't think it is a valid reason. This is not actually a blocker to get this PR merged, but I would be happy to discuss this more. In future refactorings this init method will probably no longer exist.

tests/integration/test_properties.py
NickSchimek: yes, I agree. However, I think we should at least test that it returns a json response of the requested resource, as that is its main responsibility. And eventually we should unit test the json response. Otherwise these tests do hold some value at this point, because I could replace 'return rentalProperty.json()' with 'return PropertyModel.query.first().json()' and it passes..

models/property.py
NickSchimek: We can always address this later on too when we implement marshmallow for input validation.

sokampdx: I thought about both cases as well. Only include property-manager role user and raise an error. I can do both here which might be redundant later on when fe only allow project manager role be selected when create/update property. We'll deal with that when we come to that road.

tests/integration/test_properties.py
```diff
@@ -66,8 +68,8 @@ def test_archive_property_by_id(client, auth_headers, new_property, test_databas
     assert responseBadPropertyID.get_json() == {'message': 'Property cannot be archived'}
     assert responseBadPropertyID.status_code == 400
 
-def test_delete_property_by_name(client, auth_headers, new_property, test_database):
-    test_property = PropertyModel.find_by_name(new_property.name)
+def test_delete_property_by_name(client, auth_headers, test_database):
+    test_property = PropertyModel.find_by_name("The Reginald")
```
sokampdx, line 71: If the name is 'test1' (which is new_property.name), this test would failed because there are two properties in test database with 'test1'. This is being fix by https://github.com/codeforpdx/dwellingly-app/issues/395

sokampdx:
@NickSchimek Thanks for the review. Made some changes. Ready for you again.

models/property.py
NickSchimek: That's true. Something to think about.

NickSchimek:
Thanks you @sokampdx - Can you rebase one more time?

sokampdx:
I squashed the commits so the rebase is easier. Ready for review again.

NickSchimek (approved):
Yay! Thanks for doing this @sokampdx

PR #132, opened 2020-10-22
Db migrations

### LINKED ISSUE #362: Setup/configure a db migration tool
**Purpose**

Should use a migration tool to simplify db alterations

**Proposed Solution**

Probably just use flask-migrate https://flask-migrate.readthedocs.io/en/latest/

yunnie:

Closes codeforpdx/dwellingly-app/issues/362

Implemented alembic to manage database migrations. Updated CONTRIBUTING.md to include instructions on migrations. Left flask command line functions to create, seed and remove database still function. I did not include functions to use the alembic command line to seed the database. 

Any special requirements to test? Yes. Alembic is a command line tool to manage migrations

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,1 +1,2 @@
 SECRET_KEY='dwellingly'
+PYTHONPATH=${PWD}:${PYTHONPATH}
--- a/.gitignore
+++ b/.gitignore
@@ -133,3 +133,6 @@ data.db
 
 # Jetbrains pyCharm ide
 .idea
+
+# Database migrations
+migrations/versions/*.py
--- a/.gitignore
+++ b/.gitignore
@@ -133,6 +133,3 @@ data.db
 
 # Jetbrains pyCharm ide
 .idea
-
-# Database migrations
-migrations/versions/*.py
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -61,6 +61,8 @@ If you don't see something similar, you may have several versions of Python inst
 
 ### Database migrations.
 
+Database migrations are not used for development. Please ignore do not use migrations during development.  
+
 Database migrations are managed through [Alembic](https://alembic.sqlalchemy.org/en/latest/index.html). After making a change to a model, a database migration is necessary. 
 
 - The first step is to create a revision: `pipenv run alembic revision --autogenerate -m "message"`
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -21,6 +21,7 @@ NOTE: Database is SQLite3 via SQLAlchemy
    - To re-seed the database from scratch run: `pipenv run flask db recreate`
    - Look for the file data.db to be created in the root directory
    - To find other database set-up commands run: `pipenv run flask db --help`
+   - To drop the database run: `pipenv run flask db drop`
 6. Start the server using the flask environment (required every time the project is re-opened):
    - Run: `pipenv run flask run`
    - Run and restart the server on changes: `pipenv run flask run --reload`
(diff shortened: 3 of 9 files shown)


alembic.ini
```diff
@@ -0,0 +1,85 @@
...
+# version location specification; this defaults
+# to migrations/versions.  When using multiple version
+# directories, initial revisions must be specified with --version-path
+# version_locations = %(here)s/bar %(here)s/bat migrations/versions
+
+# the output encoding used when revision files
+# are written from script.py.mako
+# output_encoding = utf-8
+
+sqlalchemy.url =
```
NickSchimek, line 38: This is not confusing at all. [insert sarcasm emoji here]

.env.example
```diff
@@ -1 +1,2 @@
 SECRET_KEY='dwellingly'
+PYTHONPATH=${PWD}:${PYTHONPATH}
```
NickSchimek, line 2: I don't see this being used anywhere. What is this for?

CONTRIBUTING.md
```diff
@@ -58,11 +59,19 @@ If you are still having issues or if your command prompt is throwing an error th
 Check to see where the Python you're running is located (`which python`). You should see something like `/Users/your_account_shortname/.pyenv/shims/python`.
 If you don't see something similar, you may have several versions of Python installed elsewhere (via Anaconda or Homebrew). If (and only if) you'd like to clear out any previous homebrew-based installs, type `brew uninstall --ignore-dependencies python3 && brew uninstall --ignore-dependencies python`.
 
-### Database commands.
+### Database migrations.
```
NickSchimek, line 62: Thank you for updating the docs.

CONTRIBUTING.md
```diff
@@ -58,11 +59,19 @@ If you are still having issues or if your command prompt is throwing an error th
...
 
-- Create and Seed the Database: `pipenv run python manage.py create`
-- Delete, create, Seed the Database: `pipenv run python manage.py recreate`
-- Delete the Database: `pipenv run python manage.py drop`
```
NickSchimek, line 65: I think we will still need to have these commands documented in the readme. We still need to them to create the database right?

alembic.ini
yunnie: Yeah, this was confusing as hell.

.env.example
yunnie: I was having trouble importing 'app' and 'db' as a modules in the alembic "env.py" file. The solution was to include a python path as one of the environmental variables. Apparently this is a common issue. The [recommended solution](https://stackoverflow.com/questions/15648284/alembic-alembic-revision-says-import-error) is to update the PYTHONPATH in the environmental variables.

CONTRIBUTING.md
yunnie: We already replaced "pipenv run python manage.py create" with "pipenv run flask db create". The "flask db create", "flask db recreate" and "flask db drop" are still functional.

NickSchimek: That's right! Thanks.

.env.example
NickSchimek: Thanks, I will check it out.

NickSchimek:
Would you be willing to spin this up and explain to everyone how it works and what it is for in the next meet? I'm not sure what everyones comfort level would be with this. I'm also not sure if @htharker42 wants to start using this now or wait until we go to production. What do you think?

yunnie:
I'd be happy to spin it up and walk people through database migrations.

.gitignore
```diff
@@ -133,3 +133,6 @@ data.db
 
 # Jetbrains pyCharm ide
 .idea
+
+# Database migrations
+migrations/versions/*.py
```
NickSchimek: I don't think we should be gitignoring these files. Otherwise we could end up with devs having different versions of the database.

yunnie: I initially thought contributors would use alembic to manage database migrations, which could create potential merge conflicts as people upgraded, downgraded and made changes, so I added the migrations/versions/*.py to gitignore. But since we're not going to use alembic in the development phase, it makes perfect sense to remove the path from gitignore, so I'll take it out.

CONTRIBUTING.md
```diff
@@ -58,11 +59,19 @@ If you are still having issues or if your command prompt is throwing an error th
...
+Database migrations are managed through [Alembic](https://alembic.sqlalchemy.org/en/latest/index.html). After making a change to a model, a database migration is necessary.
```
NickSchimek, line 66: @yunnie I would like to get this merged, but since we are not going to be using this right away. Should we add a note here, something like "Migrations are currently not being used, Please ignore for now" or simply remove the doc update and put it back in later? What do you think?

NickSchimek (dismissed):
@yunnie I would like to walk through these changes with you before to much time goes by. Please see my comment regarding the doc update, and let me know how you prefer to proceed.

CONTRIBUTING.md
yunnie: Sorry, I forgot to get back to you on this. I think it makes most sense to edit the docs back to back to the original instruction for creating, deleting and seeding the database. And to add a note stating that migrations are currently not being used "please ignore for now". I'll make the changes to the docs, if you haven't already.

yunnie: I just added a line in the CONTRIBUTING.md that instructs users not to use database migrations for development.

NickSchimek (approved):
Thanks for tackling this!

PR #133, opened 2020-10-22
Invite Email endpoint

### LINKED ISSUE #316: Update Email Resource with Invite emails
**Purpose**
Generates and sends an Email with an invitation to create an account on Dwellingly 

**Proposed Solution**
Add feature to the email resource which will populate and send an invitation email. Email should include a link to the create an account page.

Endpoint should accept the email address and send the email. 

**Acceptance Criteria**
- [ ] Add resource /email/invite 
- [ ] accepts email address as a POST request
- [ ] Auth required 
- [ ] creates and send email 
- [ ] 200 confirmation upon success 
- [ ] Endpoint  info added to Readme 

**Additional context**
See password reset email branch.

zfregin:

Closes codeforpdx/dwellingly-app/issues/316

- [X] Code builds successfully


resources/email.py
```diff
@@ -21,8 +21,7 @@ def post(self):
         message.recipients = [user.email]
 
         current_app.mail.send(message)
-        return {"message": "Message sent"}
-
+        return {"Message": "Message Sent"}
```
NickSchimek: This should not be titleized. Let's revert this change

resources/email.py
```diff
@@ -32,3 +31,19 @@ def send_reset_password_msg(user):
...
 
         current_app.mail.send(msg)
+
+        return {"message": "Message sent"}
+
+class InviteEmail(Resource):
+    parser = reqparse.RequestParser()
+    parser.add_argument('email', type=str, required=True, help="This field is not required.")
+
+    def post(self):
```
NickSchimek, line 41: Do you think we should add a test for this?

resources/email.py
```diff
@@ -32,3 +31,19 @@ def send_reset_password_msg(user):
         msg.html = render_template('emails/reset_msg.html', user=user, token=token)
 
         current_app.mail.send(msg)
+
+        return {"message": "Message sent"}
+
+class InviteEmail(Resource):
+    parser = reqparse.RequestParser()
+    parser.add_argument('email', type=str, required=True, help="This field is not required.")
```
NickSchimek:
```suggestion
    parser.add_argument('email', type=str, required=True, help="This field is required.")
```

We should also start using Marshmallow for input validation on any new endpoints. However, we can also create a refactoring ticket for this.

app.py
```diff
@@ -15,7 +15,7 @@
 from resources.tenants import Tenants
 from resources.emergency_contacts import EmergencyContacts
 from flask_mail import Mail
-from resources.email import Email
+from resources.email import Email, InviteEmail
```
NickSchimek, line 18: Technically, email is not a resource. We can put the function on the email class.
```suggestion
from resources.email import Email
```

resources/email.py
```diff
@@ -32,3 +31,19 @@ def send_reset_password_msg(user):
...
+
+    def post(self):
+        data = InviteEmail.parser.parse_args()
+
+        message = Message("You've been invited to join Dwellingly!", sender="noreply@codeforpdx.org", html="<a href=\"http://localhost:3000/signup\">Sign up for Dwellingly</a>")
```
NickSchimek: Let's use a constant instead of the hard-coded string: "http://localhost:3000"  - If we don't have an env var for this then let's create one. 

Also it would probably be good to send both html and text versions of the email, and we probably want to have some kind of message. <- We can always do this in another ticket though. @htharker42 do you have a message that should be used?

NickSchimek (commented):
@zfregin Thanks for doing this. I made some comments. The issue this is for is very similar to the issue @jls-github is working on. If this email is being called by another endpoint then we should not have a separate class for it. If this is being called directly from the frontend then a separate class is fine.

resources/email.py
zfregin: Yes, it should. I can add that once we determine whether the InviteEmail endpoint is needed

config.py
```diff
@@ -15,6 +15,9 @@ class Default(object):
     JWT_BLACKLIST_ENABLED = True
     JWT_BLACKLIST_TOKEN_CHECKS = ['access', 'refresh']
 
+    # Get server domain for Front End
+    FE_SERVER_NAME = os.environ.get('FE_SERVER_NAME')
```
NickSchimek, line 19: Nice! 👍

NickSchimek:
hey @zfregin Thanks for all the work put in here. Unfortunately there were two very similar issues created that caused confusion on what to do. I think PR #139 has everything that is needed to close out both issues. Hence, I'm going to close this PR in favor of #139. Thanks again for all the work put in here.

PR #136, opened 2020-10-29
BUG: User can change password without inputting the correct current password

### LINKED ISSUE #340: User can change password without inputting the correct current password
**Describe the bug**
When the user changes their password on the frontend. The three required parameters are:

- current password
- new password
- new password confirm

Currently the backend grabs the new password and updates, it doesn't seem to be checking if they supplied the correct current password or if they supplied the correct new password confirm

**To Reproduce**
Two ways to reproduce.

1.
Using the UI:
Login
go to settings and select change password
set current password to anything other than the correct password
put in a new password and new password confirm
click save

2.
Create some integration tests to reproduce error

**Expected behavior**
Is that the user should not be able to change the password if the current password is not supplied correctly and if the password confirm does not match password

**Acceptance Criteria**

 [ ] Create tests to reproduce error
 [ ] Implement fix

cecillesalazar:
 
Closes codeforpdx/dwellingly-app/issues/340

Fixes bug where passwords are not checked against the database password on password reset.

I think there needs to be some frontend changes to accompany this PR. I grokked code in the front end - and as far as I can tell - although we make the user provide the current password along with the new password - I don't think the current password is ever sent to the backend for any sort of comparison. 

So the patch /user/:id endpoint only takes the new password - regardless if the current password matches or not. If someone in the frontend can confirm this - the changePassword component will now have to send in both the current password as well as the new password to the patch request and we will need to create a ticket for that work.

No

No

None, unless I missed something
 
### DIFF
--- a/resources/user.py
+++ b/resources/user.py
@@ -74,7 +74,9 @@ class User(Resource):
         parser.add_argument('lastName',type=str, required=False, help="This field is not required")
         parser.add_argument('email',type=str, required=False, help="This field is not required")
         parser.add_argument('phone',type=str, required=False,help="This field is not required")
-        parser.add_argument('password',type=str, required=False,help="This field is not required")
+        parser.add_argument('current_password',type=str, required=False,help="This field is not required")
+        parser.add_argument('new_password', type=str, required=False, help="This field is not required")
+        parser.add_argument('confirm_password', type=str, required=False, help="This field is not required")
 
         data = parser.parse_args()
 
@@ -89,16 +91,28 @@ class User(Resource):
 
         if data['role']:
           user.role = RoleEnum(data['role'])
-        if (data['firstName'] != None):
+        if data['firstName'] is not None:
             user.firstName = data['firstName']
-        if (data['lastName'] != None):
+        if data['lastName'] is not None:
             user.lastName = data['lastName']
         if data['email']:
             user.email = data['email']
         if data['phone']:
             user.phone = data['phone']
-        if data['password']:
-            user.password = data['password']
+
+        # Reset Password
+        if data['current_password'] and data['new_password'] and data['confirm_password']:
+
+          # Step #1: Check if current password matches the one in the db
+          if not user.check_pw(data['current_password']):
+            return {"message": "Password does not match."}, 401
+
+          # Step #2: Check if new password and confirm password match
+          if data['new_password'] != data['confirm_password']:
+            return {"message": "New password does not match."}, 422
+
+          # Step #3: Set the new password
+          user.hash_digest = UserModel.hash_pw(data['new_password'])
 
         user.save_to_db()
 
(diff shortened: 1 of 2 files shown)


NickSchimek:
Yes, you are right. I checked the frontend and here is a pick of the request payload.

Screen Shot 2020-10-29 at 12 38 23 PM

The password field is what the new password should be. Hence, this statement `if data['currentPassword'] and data['newPassword']:` will always evaluate to false.

When I created this ticket I just assumed that all fields were being sent from the frontend. Looks like we will need to create a ticket for the frontend to send all fields in the change password form to the backend.

NickSchimek (commented):
I created issue codeforpdx/dwellingly-app/issues/418 to update the frontend. I will hold off on reviewing/merging until frontend is updated. As we may need to make some small changes once the frontend is updated.

NickSchimek (changes requested):
The frontend PR should be ready to go. You can view it here to see the update. https://github.com/codeforpdx/dwellingly-app/pull/422

It will send the following payload:
```
        current_password: data.current,
        new_password: data.new,
        confirm_password: data.confirm
```

Can you update this PR to use the above payload params. Once this is ready to go then we will get both of them merged at the same time.

cecillesalazar:
@NickSchimek - ready for another look. Not a big fan that we're mixing camel and snake case for the args - but it seems we're also doing that when we return the json with the `created_at` and `updated_at` fields. Maybe we can clean it up at another ticket just for consistency's sake - but doesn't have to happen on this PR~

NickSchimek:
When I came on board there was a mixture of camelCase and snake_case. I've been slowly changing everything to snake case where I could easily. It irritates me too. Pep8 style guide says to use snake case, so we should use snake case. I'm fine if people want to submit prs to change terms to snake case, however that can break front end configuration. So it's just been something I ignore and not worry about to much.

NickSchimek:
I remembered @htharker42 mention we should follow pep8 styling. Here is the paragraph mentioning that snake_case is preferred https://www.python.org/dev/peps/pep-0008/#function-and-variable-names. It would be nice to have some consistency with all the variable names.

resources/user.py
```diff
@@ -89,16 +91,28 @@ def patch(self,user_id):
...
+          # Step #1: Check if current password matches the one in the db
+          if not user.check_pw(data['current_password']):
+            return {"message": "Password does not match."}, 401
+
+          # Step #2: Check if new password and confirm password match
+          if data['new_password'] != data['confirm_password']:
+            return {"message": "New password does not match."}, 422
+
+          # Step #3: Set the new password
+          user.hash_digest = UserModel.hash_pw(data['new_password'])
```
NickSchimek: This can be:
```suggestion
          user.password = data['new_password']
```

It is hashed when the user is saved to db.

NickSchimek (changes requested):
Looks good. one small change.

NickSchimek (approved):
Thanks for getting this done @cecillesalazar

cecillesalazar:
@NickSchimek - thanks for the tweak to get it out~

PR #139, opened 2020-11-05
Create user invite endpoint

### LINKED ISSUE #316: Update Email Resource with Invite emails

### LINKED ISSUE #322: BACKEND: /user/invite Endpoint
**Purpose**
Create endpoint /user/invite that will accept data from the front end (without passwords). 
Should be JWT protected unlike the /user/register endpoint 

**Proposed Solution**
create endpoint which accepts the data from the endpoint
set temporary random password to be overwritten 
then create an account and then trigger an email for the user's account
Email should leverage same features of forgot password feature see ticket #289 
Message will notify user that an account has been created for them and they need to set a password

**Acceptance Criteria**
- [ ]new endpoint which is jwt protected and correctly digests the data
- [ ]endpoint commits user to database
- [ ]endpoint sends new user an email to setup their password

jls-github:
Closes codeforpdx/dwellingly-app/issues/322
Closes codeforpdx/dwellingly-app/issues/316


resources/user.py
```diff
@@ -224,3 +227,23 @@ def get(self):
...
+        user = UserModel(
+            email=data["email"],
+            firstName=data["firstName"],
+            lastName=data["lastName"],
+            phone=data["phone"],
+            password=temp_password,
+            role=RoleEnum.STAFF,
+            archived=False
+            )
+        if user.save_to_db:
```
NickSchimek: Python requires parenthesis on method calls otherwise it returns a function, and a function is truthy. This should be `user.save_to_db()` - Fixing this will cause a test to fail.

resources/user.py
```diff
@@ -224,3 +227,23 @@ def get(self):
...
+            Email.send_user_invite_msg(user)
+            return {"message": "User Invited"}, 200
+        else:
+            return {"message": "Could not create user"}, 400
```
NickSchimek: This branch is untested. If possible I would like to see a test for this scenario. However, If we use the create method for creating a user then this branch is not needed.

templates/emails/invite_user_msg.html
```diff
@@ -0,0 +1,13 @@
+<p>Hello {{ user.firstName }},</p>
+
+<p>You have been invited to the Dwellingly Platform!</p>
+
+<p>To finish creating your account, please reset your password here:</p>
```
NickSchimek: This seems confusing. perhaps we should say. Please setup your password? And we should probably mention that this link expires in ? mins

tests/integration/test_users.py
```diff
@@ -239,3 +239,18 @@ def test_get_user(client, auth_headers, new_user):
     assert is_valid(unauthorized_user_response, 401)
     assert unauthorized_user_response.json == \
             {'message': 'Admin access required'}
+
+@pytest.mark.usefixtures('client_class', 'empty_test_db')
+class TestUser:
+    def setup(self):
+        self.endpoint = '/api/user'
```
NickSchimek: could we change this to `self.endpoint = /api/user/invite`?

tests/integration/test_users.py
```diff
@@ -239,3 +239,18 @@ def test_get_user(client, auth_headers, new_user):
...
+
+    def test_invite_user(self, valid_header):
+        property_manager_json = {
+            "email": "manager@domain.com",
+            "firstName": "Leslie",
+            "lastName": "Knope",
+            "phone": "505-503-4455",
+        }
+        response = self.client.post(self.endpoint + '/invite', headers=valid_header, json=property_manager_json)
+        assert response.json == {'message': 'User Invited'}
```
NickSchimek, line 259: we should probably also assert the status code.

resources/user.py
```diff
@@ -224,3 +227,23 @@ def get(self):
             } for user in users]
 
         return {"users": ret}, 200
+
+class UserInvite(Resource):
```
NickSchimek: I would like to see this in its own file. The User file is getting pretty big and it's possible we end up adding more actions for this resource.

resources/user.py
```diff
@@ -224,3 +227,23 @@ def get(self):
...
+    @jwt_required
+    def post(self):
+        data = request.json
+        temp_password = ''.join(random.choice(string.ascii_lowercase) for i in range(8))
+        user = UserModel(
```
NickSchimek: I would like to see the create method used here; otherwise this endpoint can be prone to bugs. However, this will require a UserSchema which does not exist yet. We could either do this now or create an issue for it.

NickSchimek (changes requested):
This looks good. I requested a few changes.

resources/user.py
NickSchimek: FYI: A basic UserSchema exists now

resources/user.py
```diff
@@ -224,3 +227,23 @@ def get(self):
...
+            email=data["email"],
+            firstName=data["firstName"],
+            lastName=data["lastName"],
+            phone=data["phone"],
+            password=temp_password,
+            role=RoleEnum.STAFF,
```
NickSchimek: Should this be a hard-coded role?

resources/user.py
```diff
@@ -224,3 +227,23 @@ def get(self):
             } for user in users]
 
         return {"users": ret}, 200
+
+class UserInvite(Resource):
+    @jwt_required
```
NickSchimek: I believe this should be `@admin_required` as only admins can create users.

resources/user.py
jls-github: Changed to the create method.

resources/user.py
jls-github: Fixed with the create method. Thanks for the heads up.

templates/emails/invite_user_msg.html
jls-github: Added in those two details. The token is only set to 10 minutes at the moment. Would we ever want to widen that time window?

tests/integration/test_users.py
jls-github: Fixed

tests/integration/test_users.py
jls-github: Testing that now as well.

resources/user.py
jls-github: I moved it to its own file now.

resources/user.py
jls-github: I used the create method now. I had trouble with the user schema at first, but I think I got it working! I was trying to use the User Register Schema which doesn't allow a role.

resources/user.py
jls-github: Changed this to accept whatever value is in the JSON. My confusion was that I thought non-admin users might be inviting other users, so I wanted to set it at a lower level.

resources/user.py
jls-github: Changed this.

jls-github:
Everything should be fixed and good to go! I went ahead and pulled down the dev branch and merged the changes in.

templates/emails/invite_user_msg.html
NickSchimek: Probably not. It is not ideal, or secure for that matter, to have a time window more than a few minutes. If the token happens to expire then they can request a new one by hitting the forgot password endpoint. It might be good to mention that in the email, but I think we can forgo that for now.

resources/user.py
NickSchimek: Yes the UserRegister schema is special. I'm glad you got it figured out.

resources/user.py
NickSchimek: That makes sense, however the requirement is that only admins can invite users.

resources/user.py
```diff
@@ -9,6 +9,9 @@
 import json
 from werkzeug.security import safe_str_cmp
 from flask_jwt_extended import create_access_token, create_refresh_token, jwt_required, get_jwt_claims, get_raw_jwt, get_jwt_identity, jwt_refresh_token_required
+from resources.email import Email
+import string
+import random
```
NickSchimek, line 14: I think these imports are no longer used and can be removed.

resources/user_invite.py
```diff
@@ -0,0 +1,21 @@
...
+from resources.email import Email
+import string
+import random
+
+
+class UserInvite(Resource):
+    @admin_required
+    def post(self):
+        data = request.json
+        temp_password = ''.join(random.choice(string.ascii_lowercase) for i in range(8))
```
NickSchimek, line 15: FYI: I think @htharker42 is currently implementing a password validation to ensure users don't create weak passwords. I'm not sure what the requirements will be, but this may need to be updated. If I get this merged in before he's done then I will simply let him know to update this line if needed.

templates/emails/invite_user_msg.html
```diff
@@ -0,0 +1,13 @@
+<p>Hello {{ user.firstName }},</p>
+
+<p>You have been invited to the Dwellingly Platform!</p>
+
+<p>To finish creating your account, please set up your password here: <a href="{{ url_for('resetpassword', token=token, _method='GET', _external=True) }}"> Reset password link</a></p>
+
+<p>This link will expire in 10 minutes.</p>
+
+<p>Alternatively, you can paste the following link in your browser's address bar: {{ url_for('resetpassword', token=token, _method='GET', _external=True) }}</p>
```
NickSchimek, line 9: FYI: we probably can't use these url_for helper methods. But we will get this updated in a different PR. We will have to direct the user to whatever URL will be used for the React FE.

NickSchimek: created issue codeforpdx/dwellingly-app/issues/461 to address this.

tests/integration/test_users.py
NickSchimek: Thanks, additionally we should assert that this command `Email.send_user_invite_msg(user)` is sent. Because, if that line is inadvertently altered, or removed then the tests still test green. And furthermore it is behavior that should be under test. I think we can either Test that an email is sent or test that the method is called.

tests/integration/test_users.py
```diff
@@ -251,7 +267,6 @@ def setup(self):
             'email': 'faker.unique.email@example.com',
             'password': 'faker.password'
         }
-
```
NickSchimek, line 254: style: there should be a space here

tests/integration/test_users.py
```diff
@@ -264,4 +279,4 @@ def test_user_calls_create(self, mock_create):
         mock_create.assert_called_with(
             schema=UserRegisterSchema,
             payload=self.valid_payload
-        )
+        )
```
NickSchimek, line 282: style: it's generally preferred to have a new line at the end of the file

tests/integration/test_users.py
```diff
@@ -242,6 +242,22 @@ def test_get_user(client, auth_headers, new_user):
...
+        self.endpoint = '/api/user/invite'
+
+    def test_invite_user(self, valid_header):
+        property_manager_json = {
+            "email": "manager@domain.com",
+            "firstName": "Leslie",
+            "lastName": "Knope",
+            "phone": "505-503-4455",
+            "role": "STAFF"
+        }
```
NickSchimek, line 257: I thought there was a factory that returned property manager attributes at one time?

tests/integration/test_mailers.py
```diff
@@ -106,3 +106,8 @@ def test_reset_password_msg(self, send_mail_msg, new_user):
             Email.send_reset_password_msg(new_user)
 
         send_mail_msg.assert_called()
+
+    @patch.object(Mail, 'send')
+    def test_send_user_invite_msg(self, send_mail_msg, new_user):
+        Email.send_user_invite_msg(new_user)
```
NickSchimek: Can we get this line `with patch.object(Message, '__init__', return_value=None):` added above line 112 here. Without it this test will be really slow. 

```suggestion
        with patch.object(Message, '__init__', return_value=None):
            Email.send_user_invite_msg(new_user)
```

I think we would want to have some other tests here or at least a test in the integration test that checks that the right user is actually emailed, but I'm not to worried about at this time. I'm not sure if pytest offers some methods to easily do that either. We're going to need to switch to another emailer library so we will do it at that time.

resources/user_invite.py
```diff
@@ -0,0 +1,21 @@
...
+        user = UserModel.create(
+            schema=UserSchema,
+            payload={**data, "archived": False, "password": temp_password}
+        )
+        Email.send_user_invite_msg(user)
+        return {"message": "User Invited"}, 200
```
NickSchimek, line 21: I think we should actually return a status code of `201` here since a resource is being created. What do you think?

NickSchimek (changes requested):
This looks good. I'm half tempted to merge it in as is, however I think there are a few small things that could be updated, and then we'll be good to go.

NickSchimek:
Just realized that the `using_the_api.md` should also be updated to include this new endpoint info. The frontend devs use this so that they know how to integrate properly.
https://github.com/codeforpdx/dwellinglybackend/blob/development/doc/using_the_api.md#endpoints

resources/user_invite.py
NickSchimek: Created issue #475 to address this.

NickSchimek:
> Just realized that the `using_the_api.md` should also be updated to include this new endpoint info. The frontend devs use this so that they know how to integrate properly.
> https://github.com/codeforpdx/dwellinglybackend/blob/development/doc/using_the_api.md#endpoints

Created issue #476 to address this.

NickSchimek (approved):
Thanks for all your work on this @jls-github  - I created some issues to take care of the remaining items.

PR #142, opened 2020-11-10
Refactor Hash Password Architecture

### LINKED ISSUE #420: Refactor Hash Password Architecture
**Purpose**

Simplify the current architecture we have for storing/hashing passwords. 

**Proposed Solution**

Currently hashing methods are stored in /utils/auth.py and are imported to /resources/user.py to be used for the register, login, and patch endpoints. It also has to be imported to multiple files that deal with testing user endpoints.

To simplify it - the methods can be migrated directly into the User model, so hashing and checking can occur in one place.

**Acceptance Criteria**

N/A

N/A

**Additional context**

None

cecillesalazar:

Closes codeforpdx/dwellingly-app/issues/420

Nope, just a refactor - simplifying the current structure of hashing/storing passwords

Nope

Nope

Nope

### DIFF
--- a/models/user.py
+++ b/models/user.py
@@ -38,10 +38,18 @@ class UserModel(BaseModel):
         self.lastName = lastName
         self.email = email
         self.phone = phone
-        self.password = UserModel.hash_pw(password)
+        self._password = UserModel.hash_pw(password)
         self.role = role
         self.archived = False
         self.lastActive = datetime.utcnow()
+
+    @property
+    def password(self):
+        return self._password
+
+    @password.setter
+    def password(self, new_password):
+        self._password = UserModel.hash_pw(new_password)
 
     def update_last_active(self):
         self.lastActive = datetime.utcnow()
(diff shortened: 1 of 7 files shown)


NickSchimek (changes requested):
This is great. This handles automatic password hashing when the object is instantiated. However, we also need it for when an object is updated. I pushed up a missing test to expose it. One possible solution is to create getter and setters using pythons built-in decorators.

Something like this? Or do you have another idea?
```
def __init__(self, ...
    ...
    self._password = UserModel.hash_pw(password)
    ...


@property
def password(self):
   return self._password

@password.setter
def password(self, value):
   self._password = UserModel.hash_pw(value)
```

cecillesalazar:
@NickSchimek - nice catch. Test should be passing now - branch is rebased too~

models/user.py
```diff
@@ -68,7 +76,7 @@ def hash_pw(plaintext_password):
         return bcrypt.hashpw(bytes(plaintext_password, 'utf-8'), bcrypt.gensalt())
 
     def check_pw(self, plaintext_password):
-        return bcrypt.checkpw(bytes(plaintext_password, 'utf-8'), self.password)
+        return bcrypt.checkpw(bytes(plaintext_password, 'utf-8'), self._password)
```
NickSchimek, line 79: small detail: I think we should keep this as `self.password` which does end up returning `self._password`, however it is best practice to use getter methods for accessing private fields.

NickSchimek (changes requested):
one tiny suggestion.

NickSchimek (approved):
This is great! Thanks for updating this. I will go ahead and make that minor change in a PR to speed up the test suite. I will request your review.

PR #144, opened 2020-11-12
Create work factor env var

NickSchimek:
Created config var to set the work factor for bcrypt dependent on which environment is running.

I'm not really a fan of this, but I think it is probably the best solution right now to get the tests back up to running at a decent speed.

There were two tests failing after this change due to the app context was not present. This was happening because the database was not being used when a user was being created in the test.
 - I removed one of the tests because it is obsolete
 - Updated the second test.


cecillesalazar (dismissed):
Didn't even know you can adjust with that work factor! I'm sure there's a better way to do it - but this a good first pass. Running the entire suite for me went from 4.3 min to 15 sec 👍

PR #146, opened 2020-11-12
hash passwords on save

### LINKED ISSUE #420: Refactor Hash Password Architecture

NickSchimek:

Closes codeforpdx/dwellingly-app/issues/420

### DIFF
--- a/models/base_model.py
+++ b/models/base_model.py
@@ -32,8 +32,7 @@ class BaseModel(db.Model):
             abort(400, err.messages)
 
         obj = cls(**attrs)
-        db.session.add(obj)
-        db.session.commit()
+        obj.save_to_db()
 
         return obj
 
@@ -48,8 +47,7 @@ class BaseModel(db.Model):
         for k, v in attrs.items():
             setattr(obj, k, v)
 
-        db.session.add(obj)
-        db.session.commit()
+        obj.save_to_db()
 
         return obj
 
--- a/resources/user.py
+++ b/resources/user.py
@@ -102,7 +102,6 @@ class User(Resource):
 
         user.save_to_db()
 
-
         if user_id == get_jwt_identity():
             new_tokens = {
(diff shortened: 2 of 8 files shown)


models/base_model.py
```diff
@@ -32,8 +32,7 @@ def create(cls, schema, payload):
             abort(400, err.messages)
 
         obj = cls(**attrs)
-        db.session.add(obj)
```
cecillesalazar, line 35: Nice work drying this up here~

cecillesalazar (approved):
Looks good. I tested manually too - just checked through print statements. Any test we make will probably not be very explicit, so I'm OK with keeping things as is if you are 👍

models/base_model.py
NickSchimek: This change was actually necessary which highlights the fact that I should have a test for this.

resources/user.py
```diff
@@ -155,7 +153,7 @@ def post(self):
         if user and user.archived:
             return {"message": "Invalid user"}, 403
 
-        if user and check_pw(data['password'], user.password):
+        if user.check_pw(data['password']):
```
NickSchimek, line 156: It seems like this should be `if user and user.check_pw(...)` - Created ticket codeforpdx/dwellingly-app/issues/427 to deal with this.

models/base_model.py
NickSchimek: I will address in this ticket https://github.com/codeforpdx/dwellingly-app/issues/426

PR #147, opened 2020-11-16
Update hash architecture

### LINKED ISSUE #429: Updating user phone/email causes error
**Describe the bug**
When updating a users phone or email on the frontend the backend errors out with:
```
AttributeError: 'UserModel' object has no attribute '_password'
```

NickSchimek:
Using a fixture in the patch test when not updating a password
exposes a bug where `_password` does not exist, because it is only
available when objects are instantiated through the UserModel's init
method, but not when they're retrieved from the database.

Submitted this PR in two commits. First commit updates the test so that you can see it fail. second commits fixes the bug.

Fixes codeforpdx/dwellingly-app/issues/429


cecillesalazar (approved):
Good catch - looks good to me 👍

PR #149, opened 2020-11-16
Speed up mailer tests

NickSchimek:
The tests were taking about 11 seconds to run on my machine. It seems
the main culprit was the flask-mail library when instantiating the
Message object. The flask-mail is deprecated, and there is a ticket to
replace that. The mailer tests now take less than a half second to run.


htharker42 (commented):
Looks good, interesting accommodation

htharker42 (approved):

PR #150, opened 2020-11-18
427, check for null on find_by_email

### LINKED ISSUE #427: Remove null check on UserLogin
**Describe the bug**
The UserModel.find_by_email method can either return a user resource or a null object.

The UserLogin resource class has many branches that check for null before executing code. The code can probably be reorganized so this is not needed. For example: returning an error right away if the user does not exist, amongst other possible solutions. 

There is possibly a potential bug in this class due to the null check not being done. If this is the case - then please write a test to expose it, and refactor to remove the null check.

**Acceptance criteria**
The UserLogin resource class should be reorganized so that the multiple null checks can be avoided.

keppel2:

Closes codeforpdx/dwellingly-app/issues/427

no
no
no

### DIFF
--- a/resources/user.py
+++ b/resources/user.py
@@ -151,7 +151,7 @@ class UserLogin(Resource):
         if user and user.archived:
             return {"message": "Invalid user"}, 403
 
-        if user.check_pw(data['password']):
+        if user and user.check_pw(data['password']):
             access_token = create_access_token(identity=user.id, fresh=True)
             refresh_token = create_refresh_token(user.id)
             user.update_last_active()
--- a/tests/integration/test_users.py
+++ b/tests/integration/test_users.py
@@ -33,6 +33,13 @@ def test_user_auth(client, test_database, admin_user):
     assert responseMissingToken.json == \
             {'message': 'Missing authorization header'}
 
+def test_bad_user_auth(client, test_database):
+    login_response = client.post("/api/login", json={
+      "email": "bad@user.com",
+      "password": "pass"
+    })
+    assert login_response.status_code == 401
+
 def test_last_active(client, test_database, admin_user):
     user = UserModel.find_by_email(admin_user.email)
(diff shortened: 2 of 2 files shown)


resources/user.py
```diff
@@ -167,7 +167,7 @@ def post(self):
         if user and user.archived:
             return {"message": "Invalid user"}, 403
 
-        if user.check_pw(data['password']):
+        if user and user.check_pw(data['password']):
```
NickSchimek, line 154: I'm not sure if this null check is needed here. Could we get a test added for these null checks?

NickSchimek (changes requested):
Thanks @keppel2 for contributing. I was hoping to remove the null checks completely, however this is fine too. Although we do need a test created for when a request is posted with an invalid email. From what I can tell this test does not exist, but should exist.

resources/user.py
keppel2: I checked and this gives a more accurate response from the backend. From a 500 to a 401. Just try to log in with a non-existent user and it fixes the 500 (member access on a None).

keppel2:
I'll look into adding a test. The code change looks good.

resources/user.py
NickSchimek: Great! Can we get a test added?

tests/integration/test_users.py
```diff
@@ -33,6 +33,13 @@ def test_user_auth(client, test_database, admin_user):
     assert responseMissingToken.json == \
             {'message': 'Missing authorization header'}
 
+def test_bad_user_auth(client, test_database):
```
NickSchimek, line 36: Perfect thanks! Ideally this should use the `empty_test_db`, however this can be handled in a future refactoring. Thanks for getting this test added.

NickSchimek (approved):
Thanks @keppel2 !

PR #151, opened 2020-11-19
removed tenant from RoleEnum ran db migrations as per Contrib doc all…

### LINKED ISSUE #432: Remove the tenant role
**Purpose**
Remove the UserRole Tenant - Tenant is not a role and should be removed.

**Proposed Solution**
Remove tenant from the RoleEnum in the models/user. 

**Acceptance Criteria**

- [ ] all tests passing after removal of tenant role.

whiletrace:
… tests passing

Closes codeforpdx/dwellingly-app/issues/432

deleted the tenant attribute from the class RoleEnum 
ran migration as per the Contrib Docs all tests pass as is 
reseeding may be necessary since db table structure has changed  


migrations/versions/f7cb1b951d51_message.py
```diff
@@ -0,0 +1,28 @@
+"""message
```
NickSchimek: Can we remove this file? We're currently not using alembic and shouldn't be running migrations. I think there is a note in the README, but if it is not clear I'd be happy to get a PR for it.

NickSchimek (changes requested):
Let's remove the migration file. We're not using alembic yet. We will once we get closer to deployment. For now we just drop and rebuild the db with `pipenv run flask db recreate`

migrations/versions/f7cb1b951d51_message.py
whiletrace: ok Ill post a quick fix, The contrib docs are a bit confusing on this matter, maybe a little format or punctuation there could help. but Ill get rid of the migration file.

NickSchimek (approved):
Thanks @whiletrace

PR #152, opened 2020-11-19
Implement cors

### LINKED ISSUE #352: Enable CORS protection for specific origin
**Purpose**
Should only allow the frontend domain access to the backends resources. Currently setup to allow a request for a resource from any origin. https://github.com/codeforpdx/dwellinglybackend/blob/development/app.py#L104

This may need to be implemented closer to our first production deployment. Read more about [CORS](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS)

**Proposed Solution**

https://flask-cors.readthedocs.io/en/latest/#resource-specific-cors

**Acceptance Criteria**

- [ ] Should only allow CORS from a specific origin

yunnie:

Closes codeforpdx/dwellingly-app/issues/352

CORS now uses an environmental variable to set permitted origins. Currently set to "*" for development. This needs to be changed for production. 


yunnie:
I had to modify the Travis environment to include the CORS_ORIGINS environmental variable. I didn't include a default setting to ensure that it fails if the environmental variable is not correctly configured in production.

.travis.yml
```diff
@@ -2,7 +2,7 @@ language: python
 python:
   - "3.8"
 env:
-  - SECRET_KEY=somethinghereforciprtoexecuteintestbuilds
+  - SECRET_KEY=somethinghereforciprtoexecuteintestbuilds CORS_ORIGINS="*"
```
NickSchimek: Interesting, I didn't know yml allowed same line assignments. Could we break this out into two lines?
```suggestion
  - SECRET_KEY=somethinghereforciprtoexecuteintestbuilds
  - CORS_ORIGINS="*"
```

config.py
```diff
@@ -31,6 +32,9 @@ class Default(object):
     # Configure JWT error message key
     JWT_ERROR_MESSAGE_KEY = 'message'
 
+    # Configure CORS origins
+    CORS_ORIGINS = re.split(r'[,;\s]', os.environ.get('CORS_ORIGINS'))
```
NickSchimek: When I run the app without the env var set I get a cryptic error message returned. 
```
TypeError: expected string or bytes-like object
```
This is because the get method returns None and passes it into the regex. 
If we change this:
`os.environ.get('CORS_ORIGINS'))`
to this: 
`os.environ['CORS_ORIGINS'])`
Then the user will be informed that they're missing a key CORS_ORIGINS
```
KeyError: 'CORS_ORIGINS'
```
Still not that informative, but a bit better I think

NickSchimek: The regex seems a bit complex. I think having a test here would be valuable and would accomplish a couple things:
1. It would verify this works as intended
2. It would act as documentation to show what values could be inputted and the expected outputted values

I think Number 2 is important, because there is nothing to show whoever deploys the app what value to put in for CORS_ORIGINS. They might simply do '*' as the `env.example` shows.

NickSchimek (changes requested):
Thanks for the PR @yunnie and all the work you put into it.
I'm not excited about mixing both environment variables and config variables for this task. I think we should either choose one and stick to it, or is there a reason to have both a CORS_ORIGINS config var and env var?

Pro's to choosing an env var for this task:

1. The value is kept secret and hidden - although not necessary in this case
2. It can be easily changed - Probably not necessary. As origins usually don't change

Con's of using an env var:

1. Requires dev to set the env var - Not a big issue, however even if we communicate on slack to update the env file it will most likely cause issues for several developers over the next few weeks.

Pro's of choosing config var:

1. It is documented and version controlled
2. It cannot be changed easily
3. No need for devs to update env file

Cons of choosing config var:

1. It is documented and version controlled
2. It cannot be changed easily

I know you had some reservations about using strictly config var for this, however I think the same reservations apply for env vars.
For a config var we could set the dev and test environment explicitly and then in the production environment do any one of the following
```
# CORS_ORIGINS="site@origin.com"
```
or
```
CORS_ORIGINS="UPDATE THIS WITH FRONTEND ORIGIN"
```

Do the pros and cons sound about right? Do you have anything to add?
I think we can't go wrong with either way, however please address my other concerns I noted.

app.py
```diff
@@ -58,7 +58,7 @@ def create_app(env):
     create_routes(app)
 
     #allow cross-origin (CORS)
-    CORS(app)
+    CORS(app, origins=app.config['CORS_ORIGINS'])
```
NickSchimek, line 61: If we change this to:
```suggestion
    CORS(app, origins=[app.config['CORS_ORIGINS'])]
```
Do we still need to have the regex on the config var?

.travis.yml
yunnie: I originally split the assignment into two lines, but Travis created two separate environments. Documentation states to include all of the environmental variables on a single line.

app.py
yunnie: The regex is to allow for multiple separators in the case where multiple origins are set in the environmental variable. So "origin1,origin2" and "origin1;origin2" or "origin1 origin2" are all permitted. But if we move the the configuration into Development, Testing, and Production classes then the regex is not necessary.

yunnie:
Perhaps the best way to implement this is to use configuration. My main concern was making sure that people remember to set the CORS origins in production. After reading your comments, I think the best way to do this is to use a configuration setting in the Development, Testing and Production classes. We can set dummy values which are guaranteed to fail in the Testing and Production classes. This would ensure that we remember to update the origins when moving into Testing and Production. Let me know what you think.

BTW, I really appreciate the feedback. It helps force me to think about issues that I didn't consider. It's a great learning experience.

NickSchimek:
Sounds good @yunnie I'm glad the comments helped, and I'm glad your thinking about how to make sure it is set appropriately in the production environment, I hadn't really given that a thought. Your approach sounds good, although I'm not sure why we would want to make the testing environment fail? I guess it shouldn't matter what the test environment is set to. Whatever the answer is here I think your approach sounds good.

.travis.yml
NickSchimek: Interesting. I did not know that.

config.py
yunnie: CORS origins are now configurations and not in the environmental variable. So this allows us to get rid of the reg ex.

NickSchimek (approved):
Thanks @yunnie  - Looks good.

PR #153, opened 2020-11-19
Make property name unique

### LINKED ISSUE #395: Fix routes properties/name to use an unique identifier
**Describe the bug**
The properties/name routes should use id instead of name OR enforce an unique name for PropertyModel
Because, the name is not unique in the PropertyModel and the find_by_name method only return the first row if multiple property has the same name. There is a risk of get/update/delete the wrong property.
https://github.com/codeforpdx/dwellinglybackend/blob/fbff27efaa80ce29d1cf260b14c0fa560b37b5b8/models/property.py#L58

see 
```
# | GET    | `v1/property/:name`  | Gets a single property     |
# | PUT    | `v1/property/:name`  | Updates a single property  |
# | DELETE | `v1/property/:name`  | Deletes a single property  |
```

**Expected behavior**
We should get an unique identifier to ensure an unique property is being get/update/delete.

gussmith23:
Closes codeforpdx/dwellingly-app/issues/395. Makes the "name" field of a Property unique. Changes tests and seed data to fit this new requirement.

Any new dependencies to install? ❌
Any special requirements to test? ❌

### DIFF
--- a/models/property.py
+++ b/models/property.py
@@ -12,7 +12,7 @@ class PropertyModel(BaseModel):
     __tablename__ = "properties"
 
     id = db.Column(db.Integer, primary_key=True)
-    name = db.Column(db.String(100))
+    name = db.Column(db.String(100), unique=True)
     address = db.Column(db.String(250))
     unit = db.Column(db.String(20), default="")
     city = db.Column(db.String(50))
--- a/models/property.py
+++ b/models/property.py
@@ -60,6 +60,7 @@ class PropertyModel(BaseModel):
 
     @classmethod
     def find_by_name(cls, name):
+        # TODO Can we use a different call here, now that name is unique?
         return cls.query.filter_by(name=name).first()
 
     @classmethod
--- a/models/property.py
+++ b/models/property.py
@@ -60,7 +60,6 @@ class PropertyModel(BaseModel):
 
     @classmethod
     def find_by_name(cls, name):
-        # TODO Can we use a different call here, now that name is unique?
         return cls.query.filter_by(name=name).first()
 
     @classmethod
--- a/models/property.py
+++ b/models/property.py
@@ -60,7 +60,7 @@ class PropertyModel(BaseModel):
 
     @classmethod
     def find_by_name(cls, name):
-        return cls.query.filter_by(name=name).first()
+        return cls.query.get(name=name)
 
     @classmethod
     def find_by_manager(cls, manager_id):
--- a/models/property.py
+++ b/models/property.py
@@ -60,7 +60,7 @@ class PropertyModel(BaseModel):
 
     @classmethod
     def find_by_name(cls, name):
-        return cls.query.get(name=name)
+        return cls.query.filter_by(name=name).first()
 
     @classmethod
     def find_by_manager(cls, manager_id):
--- a/tests/integration/test_leases.py
+++ b/tests/integration/test_leases.py
@@ -155,17 +155,20 @@ class TestLeaseAuthorizations:
         response = self.client.get('/api/lease', headers=admin_header)
         assert response.status_code == 200
 
-    def test_pm_is_authorized_to_create(self, pm_header, create_tenant, create_property):
-        response = self.client.post('/api/lease', json=self.valid_payload(create_tenant().id, create_property().id), headers=pm_header)
+    def test_pm_is_authorized_to_create(self, pm_header, create_tenant):
+        tenant = create_tenant()
+        response = self.client.post('/api/lease', json=self.valid_payload(tenant.id, tenant.propertyID), headers=pm_header)
 
         assert response.status_code == 201
 
-    def test_staff_are_authorized_to_create(self, staff_header, create_tenant, create_property):
-        response = self.client.post('/api/lease', json=self.valid_payload(create_tenant().id, create_property().id), headers=staff_header)
+    def test_staff_are_authorized_to_create(self, staff_header, create_tenant):
+        tenant = create_tenant()
+        response = self.client.post('/api/lease', json=self.valid_payload(tenant.id, tenant.propertyID), headers=staff_header)
         assert response.status_code == 201
 
     def test_admin_is_authorized_to_create(self, admin_header, create_tenant, create_property):
-        response = self.client.post('/api/lease', json=self.valid_payload(create_tenant().id, create_property().id), headers=admin_header)
+        tenant = create_tenant()
+        response = self.client.post('/api/lease', json=self.valid_payload(tenant.id, tenant.propertyID), headers=admin_header)
         assert response.status_code == 201
 
     def test_pm_is_authorized_to_delete_lease(self, pm_header, create_lease):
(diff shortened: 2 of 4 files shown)


tests/integration/test_leases.py
```diff
@@ -155,17 +155,20 @@ def test_admin_is_authorized_to_get_all(self, admin_header, create_lease):
         response = self.client.get('/api/lease', headers=admin_header)
         assert response.status_code == 200
 
-    def test_pm_is_authorized_to_create(self, pm_header, create_tenant, create_property):
-        response = self.client.post('/api/lease', json=self.valid_payload(create_tenant().id, create_property().id), headers=pm_header)
+    def test_pm_is_authorized_to_create(self, pm_header, create_tenant):
+        tenant = create_tenant()
+        response = self.client.post('/api/lease', json=self.valid_payload(tenant.id, tenant.propertyID), headers=pm_header)
```
NickSchimek, line 160: `tenant.propertyID`

Technically, tenant's don't always have a property id. If this is a brand new lease then the tenant would not have a relationship with the property. Is there a reason why this is being changed?

NickSchimek: Perhaps it would've been more accurate for the existing test to do `create_tenant(propertyID=None)`
or
We can create a new fixture called `create_unhoused_tenant()`

I'm actually not a fan of having `propertyID` on the tenant model, but that's an issue for another day.

models/property.py
```diff
@@ -60,6 +60,7 @@ def json(self):
 
     @classmethod
     def find_by_name(cls, name):
+        # TODO Can we use a different call here, now that name is unique?
```
NickSchimek: YES! I would prefer we do. Good catch.

gussmith23:
Haha, that was YOUR catch, Nick :) you messaged me about it a while ago,
but the call you gave me isn't exactly right. I have to search through the
SQL alchemy documentation (I think?) to find the right one .

NickSchimek:
> Haha, that was YOUR catch, Nick :) you messaged me about it a while ago, but the call you gave me isn't exactly right. I have to search through the SQL alchemy documentation (I think?) to find the right one .

I read sql-alchemy docs and couldn't find anything. It seems this is the recommended way to do it. So no changes necessary. https://flask-sqlalchemy.palletsprojects.com/en/2.x/queries/#querying-records - apparently the select statement isn't fired until `first()` is called; otherwise it returns a query object. I wonder if there is a way to show what sql statements are being executed in the dev environment?

tests/integration/test_leases.py
gussmith23: The problem here stems from the fact that firing off both of these functions leads to two properties with the same name being added. This happens because `create_tenant()` creates a property, and `create_property()` creates a property with the same name!

Like you said, maybe creating a tenant shouldn't create a property? I've changed things a bit so that we can do `create_tenant(propertyID=None)`, but the request is now failing with code 400. I'm unsure what `valid_payload()` should do when it gets a `None` value for property id -- is that a problem? I'm totally unsure of what that request ends up looking like.

Also, I'm still understanding the database model here, but shouldn't all leases involve a tenant id and a property id? How could we add a lease without a property id? Just from the name, I assumed that leases were what bound tenants to properties.

NickSchimek: ah I understand the reason for these changes now. Because in the original tests they were creating two properties. I should've read your commit comments! Sorry I'm not used to looking, because I don't usually see that. Thanks for adding those. Next time I will try to be better.

> The problem here stems from the fact that firing off both of these functions leads to two properties with the same name being added

This is easily fixed. I'm actually going to be creating issues to implement faker on all of the factories. This will allow factories to be created with dynamic data. You can see an implementation of this in the User factory in #154 specifically on the email field.

> Like you said, maybe creating a tenant shouldn't create a property? I've changed things a bit so that we can do create_tenant(propertyID=None), but the request is now failing with code 400. I'm unsure what valid_payload() should do when it gets a None value for property id -- is that a problem? I'm totally unsure of what that request ends up looking like.

Yeah, The factory is not setup to accept None. That could be tweaked. It could look something like this:
```
def create_tenant():
    def _create_tenant(property=None):
        tenant = TenantModel(
                firstName="firstName",
                lastName="lastName",
                phone="phone",
                propertyID=property.id if property else None,
                staffIDs=[],
                unitNum=3
            )
        tenant.save_to_db()
        return tenant
    yield _create_tenant
```
And then (unfortunately) the lease factory and then some of the tests would need to be updated. I would opt for creating an unhoused tenant factory and use the faker library in both the new and existing factory. The 400 is being returned because there is probably a validation error, hence the valid_payload being sent is not valid. 

The request being sent is the valid payload. so it would like something like this:
```
           {
                'dateTimeStart': '11/25/2020 17:14:01',
                'dateTimeEnd': '11/25/2021 17:14:20',
                'tenantID': 1,
                'propertyID': 1
            }
```

> Also, I'm still understanding the database model here, but shouldn't all leases involve a tenant id and a property id? How could we add a lease without a property id? Just from the name, I assumed that leases were what bound tenants to properties.

Yes you are absolutely correct here. That is why I'm NOT a fan of having a propertyID on the tenant model, because that creates a situation where we need to always update two columns in two separate tables. One propertyID on the lease, and one propertyID on the tenant, and they should always be the same. Leaving it this way could eventually cause our database to have erroneous data. So yes. You are absolutely correct and I agree with you here. The lease should bound the tenant to the property. If you would like to create an issue for this that would be outstanding.

NickSchimek: I'm also fine with removing the propertyID on the tenant model in this PR too. If that is something you would like to do.

NickSchimek: Let me know if you'd prefer to this merge this and get issues created for this other stuff or add it to this PR. I'm happy either way.

gussmith23: Hm. Strangely, stuff started working after I just rebased. Tenant model is still exactly the same. Let's get this merged and do the other stuff in other PRs. To be clear, the other change we're talking about is removing propertyID on tenant model? Is there also a change to the lease model?

gussmith23:
Just rebased, checks are passing now.

tests/integration/test_leases.py
NickSchimek: FYI: Created issue codeforpdx/dwellingly-app/issues/455 to address the double relationship issue. Thanks for bringing it up again.

NickSchimek: > To be clear, the other change we're talking about is removing propertyID on tenant model? Is there also a change to the lease model?

 I think there was something else. I will need to jog my memory and I will get back to you. Thanks for getting this rebased. I will get it merged soon. Thank you

NickSchimek: @htharker42 recently updated the lease model. So I think we're all set there.

NickSchimek: Issue codeforpdx/dwellingly-app/issues/459 created to get Faker in use on all the factories. I was going to create an issue to update these tests afterwards, but I think these tests will actually end up being deleted in the future, because the lease endpoint is not being utilized by the FE UI.

NickSchimek (approved):
Thanks for knocking this out @gussmith23  - Excellent work 💫

PR #156, opened 2020-11-25
Add more settings to .vscode/settings.json

gussmith23:
No issue for this one.
I discussed this in standup: I switched to vscode, and my editor automatically added some settings. They seem pretty basic, would love to hear peoples' opinions.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,3 +1,11 @@
 {
-    "python.linting.enabled": false
+    "python.linting.enabled": false,
+    "python.formatting.provider": "black",
+    "editor.formatOnSave": false,
+    "python.testing.pytestArgs": [
+        "tests"
+    ],
+    "python.testing.unittestEnabled": false,
+    "python.testing.nosetestsEnabled": false,
+    "python.testing.pytestEnabled": true,
 }


NickSchimek (approved):
I don't think we should be committing this file, I believe I removed it once, but it got added back. However, since I don't use VS Code it doesn't affect me 😉

aedwardg:
I wonder if there is a better way to include these settings for VS Code users.  I've always put the `.vscode` folder in the `.gitignore` because VS Code will append things like the absolute path of your Python version into `settings.json` without telling you and you don't want to commit/push that to the repo.

Maybe we could add in something similar to the `.env.example` where newcomers can just copy those settings into their own `settings.json`, but have the `.vscode` folder in `.gitignore` to prevent accidental changes from being pushed.

gussmith23:
This would be fine with me! Ignoring the .vscode dir would be just as good
for my needs.

NickSchimek:
Yes, this should be in the `.gitignore` - I had added it there once before it, but it keeps making its way back in for some reason.  Maybe I didn't gitignore correctly, and since I don't use vscode I didn't notice it. I'd be happy to accept a pull request.

PR #158, opened 2020-11-30
Issue447 unit numfix

### LINKED ISSUE #447: Remove the field name from the lease Model
**Purpose**
The name is not being populated as the frontend does not utilize that field. It can be removed across the app.

htharker42:
-resolves location of unitNum. Moves it from tenant model to lease model, also removes name column from lease, updates tests, seed file.
Closes codeforpdx/dwellingly-app/issues/447

Is a re-seeding of the database necessary? yes
Any special requirements to test? re-seed db.

- [X] Code builds successfully


tests/integration/test_tenants.py
```diff
@@ -27,12 +27,12 @@ def test_tenants_POST(client, auth_headers):
...
         "phone": "111-111-1111",
         "propertyID": 1,
-        "staffIDs": [1, 2],
-        "unitNum": "237"
+        "staffIDs": [1, 2]
     }
 
     response = client.post(endpoint, json=newTenant,
                            headers=auth_headers["admin"])
+    print(response)
```
NickSchimek: Let's remove this print statement.

NickSchimek (changes requested):
Looks good. I've got one question, and found one print statement that should be removed.

tests/integration/test_tenants.py
htharker42: son of a... thought I took that out last night. just pushed update

models/lease.py
```diff
@@ -8,9 +8,9 @@ class LeaseModel(BaseModel):
     __tablename__ = "lease"
 
     id = db.Column(db.Integer, primary_key=True)
-    name = db.Column(db.String(100))
     propertyID = db.Column(db.Integer, db.ForeignKey('properties.id'), nullable=False)
     tenantID = db.Column(db.Integer, db.ForeignKey('tenants.id'), nullable=False)
     occupants = db.Column(db.Integer)
     dateTimeStart = db.Column(db.DateTime, nullable=False)
     dateTimeEnd = db.Column(db.DateTime, nullable=False)
+    unitNum = db.Column(db.String(10))
```
NickSchimek, line 16: Ooops I thought I made a comment here earlier. Should this be required? `nullable=False`

htharker42: I thought about this when I implemented it. I think making it nullable is fine because not all properties will have units. It could be an ADU type situation.

NickSchimek: OK sounds good. What is an ADU type situation?

NickSchimek: nm. I just googled it. Accessory Dwelling Unit.

NickSchimek (approved):
Thanks looks good!

PR #159, opened 2020-12-01
Remove .vscode directory, move example settings

aedwardg:

- Removes (hopefully) `.vscode` directory from repo and applies `.gitignore` to that folder moving forward.
- Moves useful VS Code settings into an example file for members to copy into their own `settings.json` if they want.

Discussed in #156

Any special requirements to test? YES 👇 

- After merge, double check to make sure development branch no longer has `.vscode` directory -- I'm not sure how it will work on merge, but I believe it will be gone.  If directory still exists, it will need to be manually deleted from development branch and then committed for the `.gitignore` to kick in.

In theory, this should prevent the `.vscode` folder from popping up again, unless someone keeps their own from a previously created branch when resolving merge conflicts or something. 🤞  


NickSchimek (approved):
Thanks for the update!

PR #161, opened 2020-12-04
Creating a tenant will now also create a lease, if appropriate

pogorobot:

Closes #codeforpdx/dwellingly-app/issues/452

This should allow integration of the calendar modal  #456
One new integration test added to confirm successful formatting

### DIFF
--- a/resources/tenants.py
+++ b/resources/tenants.py
@@ -7,9 +7,11 @@ from db import db
 from models.tenant import TenantModel
 from models.user import UserModel
 from models.lease import LeaseModel
+<<<<<<< HEAD
 from schemas.lease import LeaseSchema
 from datetime import datetime
 from utils.time import Time
+from datetime import datetime
 
 # | method | route                | action                    |
 # | :----- | :------------------- | :------------------------ |
@@ -50,7 +52,6 @@ class Tenants(Resource):
         tenantEntry = TenantModel(**data) 
         TenantModel.save_to_db(tenantEntry)
 
-<<<<<<< HEAD
         returnData = tenantEntry.json()
 
         leaseData = request.json
@@ -69,7 +70,6 @@ class Tenants(Resource):
             )
             returnData.update({'occupants': leaseData['occupants'], 'propertyID': leaseData['propertyID'], 'unitNum': leaseData['unitNum']})
             
-=======
         #Attempt to create a lease from given data 
         leaseData = Tenants.leaseParser.parse_args()
 
@@ -83,7 +83,6 @@ class Tenants(Resource):
             LeaseModel.save_to_db(leaseEntry)
             returnData.update({'occupants': leaseData.occupants, 'propertyID': leaseData.propertyID, 'unitNum': leaseData.unitNum})
 
->>>>>>> 60759c0... Creating a tenant will now also create a lease, if appropriate
         return returnData, 201
 
 
--- a/resources/tenants.py
+++ b/resources/tenants.py
@@ -9,6 +9,7 @@ from models.user import UserModel
 from models.lease import LeaseModel
 from schemas.lease import LeaseSchema
 from datetime import datetime
+from utils.time import Time
 
 # | method | route                | action                    |
 # | :----- | :------------------- | :------------------------ |
@@ -42,7 +43,6 @@ class Tenants(Resource):
 
 
     @admin_required
-    @jwt_required
     def post(self):
         data = Tenants.parser.parse_args()
         if TenantModel.find_by_first_and_last(data["firstName"], data["lastName"]):
@@ -56,12 +56,17 @@ class Tenants(Resource):
         leaseData = request.json
         leaseData.update({'tenantID': tenantEntry.id})
 
-        if ("occupants" in leaseData and "dateTimeEnd" in leaseData and "dateTimeStart" in leaseData and "propertyID" in leaseData):
+        #if this tenant has a lease
+        if ("dateTimeEnd" in leaseData and "dateTimeStart" in leaseData and "propertyID" in leaseData):
+
+            #convert dateTimeStart and dateTimeEnd from iso8601 format
+            leaseData["dateTimeStart"] = Time.format_date(datetime.fromisoformat(leaseData["dateTimeStart"]))
+            leaseData["dateTimeEnd"] = Time.format_date(datetime.fromisoformat(leaseData["dateTimeEnd"]))
+
             LeaseModel.create(
                 schema=LeaseSchema,
                 payload=leaseData
             )
-            print(leaseData['occupants'])
             returnData.update({'occupants': leaseData['occupants'], 'propertyID': leaseData['propertyID'], 'unitNum': leaseData['unitNum']})
             
         return returnData, 201
--- a/resources/tenants.py
+++ b/resources/tenants.py
@@ -7,10 +7,10 @@ from db import db
 from models.tenant import TenantModel
 from models.user import UserModel
 from models.lease import LeaseModel
-<<<<<<< HEAD
 from schemas.lease import LeaseSchema
 from datetime import datetime
 from utils.time import Time
+from schemas.lease import LeaseSchema
 from datetime import datetime
 
 # | method | route                | action                    |
@@ -29,7 +29,6 @@ class Tenants(Resource):
     parser.add_argument('propertyID',required=False,help="This field can be provided at a later time")
     parser.add_argument('staffIDs',action='append',required=False,help="This field can be provided at a later time")
 
-
     @admin_required
     def get(self, tenant_id=None):
         # GET /tenants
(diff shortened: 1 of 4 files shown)


resources/tenants.py
```diff
@@ -22,6 +24,13 @@ class Tenants(Resource):
     parser.add_argument('propertyID',required=False,help="This field can be provided at a later time")
     parser.add_argument('staffIDs',action='append',required=False,help="This field can be provided at a later time")
 
+    leaseParser = reqparse.RequestParser(bundle_errors=True)
```
NickSchimek: We are migrating away from using the RequestParser as it scheduled to be deprecated. We have implemented Marshmallow to handle input deserialization and validation. You can see how to create a lease in the `resources/lease.py` file. - Let me know if there are any questions.

resources/tenants.py
```diff
@@ -45,7 +54,20 @@ def post(self):
...
+        leaseData = Tenants.leaseParser.parse_args()
+
+        returnData = tenantEntry.json()
+
+        if (leaseData.occupants and leaseData.dateTimeEnd and leaseData.dateTimeStart and leaseData.propertyID):
+            leaseData.tenantID = tenantEntry.id
+            leaseData.dateTimeStart = datetime.strptime(leaseData.dateTimeStart, "%Y-%m-%dT%H:%M:%S.%fZ")
+            leaseData.dateTimeEnd = datetime.strptime(leaseData.dateTimeEnd, "%Y-%m-%dT%H:%M:%S.%fZ")
+            leaseEntry = LeaseModel(**leaseData)
+            LeaseModel.save_to_db(leaseEntry)
```
NickSchimek: This line does another transaction to the DB. It would be nice to have both the lease and tenant creation in a single transaction. Also we should use the `LeaseModel.create` method here so that validation errors will send back a 400 response accordingly.  The create method will invoke Marshmallow to do the validation checks and input deserialization. It's already setup and implemented for the lease. Just need to plug it in, and update the expected time formats for dateTimeStart and dateTimeEnd fields.

tests/integration/test_tenants.py
```diff
@@ -30,12 +31,30 @@ def test_tenants_POST(client, auth_headers):
...
     }
 
+    newTenantWithLease = {
+        "firstName": "Finn",
+        "lastName": "The Human",
+        "phone": "123-555-4321",
+        "propertyID": 2,
+        "occupants": "3",
+        "dateTimeEnd": "2021-04-29T07:00:00.000Z",
+        "dateTimeStart": "2020-11-26T08:00:00.000Z",
```
NickSchimek: Not critical, However we should probably use dynamic dates here. If we ever add validation checks that don't allow leases to be created with dateTimeEnd and start dates that are in the past then this test will start to fail. I'm sure Python has a method that returns this time format as it is a standard format, but I'm not 100% sure.

resources/tenants.py
NickSchimek: Someone is already in the process of removing the RequestParsers in this resource. So it would be nice to not have to add work to that.

NickSchimek (changes requested):
Thanks for the PR. Please check my comments and let me know if there are any questions. Also thanks for adding tests 💫

resources/tenants.py
pogorobot: I guess I'm not sure how to create a valid lease (which requires a tenant ID) until after the tenant object has been created. I will switch this over to invoke Marshmallow, though.

NickSchimek: Sounds good. It is not that critical. I will get a separate issue created for that.

resources/tenants.py
```diff
@@ -36,16 +42,29 @@ def get(self, tenant_id=None):
 
 
     @admin_required
+    @jwt_required
```
NickSchimek: What's the reason for adding this?

resources/tenants.py
```diff
@@ -36,16 +42,29 @@ def get(self, tenant_id=None):
...
-        
         TenantModel.save_to_db(tenantEntry)
 
-        return tenantEntry.json(), 201
+        returnData = tenantEntry.json()
+
+        leaseData = request.json
+        leaseData.update({'tenantID': tenantEntry.id})
+
+        if ("occupants" in leaseData and "dateTimeEnd" in leaseData and "dateTimeStart" in leaseData and "propertyID" in leaseData):
```
NickSchimek: I think we'll need to ask the FrontEnd to send us nested Lease object in the JSON request when a lease needs to be created too so that we can avoid these checks here as it is confusing to what purpose they serve.

resources/tenants.py
```diff
@@ -36,16 +42,29 @@ def get(self, tenant_id=None):
...
+            LeaseModel.create(
+                schema=LeaseSchema,
+                payload=leaseData
+            )
+            print(leaseData['occupants'])
```
NickSchimek: This print statement should be removed.

resources/tenants.py
NickSchimek: occupants is not a required field, Let's remove that check, and leave the others for now.

tests/integration/test_tenants.py
```diff
@@ -30,12 +31,30 @@ def test_tenants_POST(client, auth_headers):
...
     }
 
+    newTenantWithLease = {
+        "firstName": "Finn",
+        "lastName": "The Human",
+        "phone": "123-555-4321",
+        "propertyID": 2,
+        "occupants": 3,
+        "dateTimeEnd": Time.one_year_from_now(),
+        "dateTimeStart": Time.yesterday(),
```
NickSchimek: The frontend sends us an iso8601 time format. Can we test with that time format. I believe this test would fail with that time format, because the schema expects the same format that the Time class returns. So we'll need to update the `dateTimeEnd` and `dateTimeStart` fields in the schema to work with the iso8601 format. 

It would be fine to hard-code dates here if Python doesn't have anything we can use to get an iso8601 time format.

NickSchimek (changes requested):
Excellent, this is looking good. A few comments to address.

tests/integration/test_tenants.py
pogorobot: Test updated to use iso8601. Weirdly, I had to use utils.Time to re-convert it back into a string with our bespoke format, for Marshmallow to accept it as a valid datetime. (The datetime literal itself was not considered a valid datetime).

I'm not at all sure I'm not missing some part of the configuration, but at least I can verify that iso8601-formatted dates will save correctly now.

resources/tenants.py
pogorobot: I erroneously assumed it was necessary because it showed up in the function I was referencing.

tests/integration/test_tenants.py
NickSchimek: We could update the schema for the lease so that it will accept the iso format. That can be done here: https://github.com/codeforpdx/dwellinglybackend/blob/development/schemas/lease.py#L14

tests/integration/test_tenants.py
```diff
@@ -30,12 +32,30 @@ def test_tenants_POST(client, auth_headers):
...
     }
 
+    newTenantWithLease = {
+        "firstName": "Finn",
+        "lastName": "The Human",
+        "phone": "123-555-4321",
+        "propertyID": 2,
+        "occupants": 3,
+        "dateTimeEnd": Time.one_year_from_now_iso(),
+        "dateTimeStart": Time.yesterday_iso(),
```
NickSchimek, line 42: This test is failing, and I think it is because we are passing datetime objects rather than a stringified iso date. Can you confirm?

NickSchimek (changes requested):
There is one test failure that needs to be addressed.

tests/integration/test_tenants.py
pogorobot: I've spent today puzzling about this -- that test does not fail locally. I even checked out a clean branch with just the current development branch and these changes merged in, and all the tests run just fine. Time.one_year_from_now_iso() returns a stringified iso datetime, and it all proceeds as you would expect. Without being able to reproduce the failure, I'm at a loss for how to even investigate.

NickSchimek: Well, that is interesting. I'm going to create a test or two and push that up so that I can get a better sense of where it is failing.

NickSchimek: OK it's passing. This was the problem:
```
    @staticmethod
    def one_year_from_now_iso():
        return datetime.isoformat(datetime.today() + relativedelta(year=1))
```
The `year` parameter should be pluralized; otherwise it seems to return the year 1 instead of the relative year. Still unclear why it didn't fail locally, but on the CI the json response for the failing request was a validation error stating that it was an invalid datetime format.

Feel free to check things over and update if need be. I will give this another once over tomorrow/today, and will most likely get this merged in as I think it is good to go.

NickSchimek (approved):
Excellent work @pogorobot Thank you!

PR #162, opened 2020-12-04
Refactor emergency contacts

### LINKED ISSUE #399: Refactor emergency contact resource
**Purpose**
Refactor the Emergency Contact resource to use the new db methods `find`, `create`, `update`, `delete`

**Acceptance Criteria**
- [x] Create schema to use marshmallow for input validation
- [x] Add the BaseInterfaceTest to the Emergency Contact test and create unit tests for all public methods. see [this commit](https://github.com/codeforpdx/dwellinglybackend/blob/31610de13449c002d2394bfa7b2e48a8c12e5b1e/tests/unit/test_lease.py) for an example.
- [x] Add the BaseInterfaceTest to the Contact Number test and create unit tests for all public methods. see [this commit](https://github.com/codeforpdx/dwellinglybackend/blob/31610de13449c002d2394bfa7b2e48a8c12e5b1e/tests/unit/test_lease.py) for an example.
- [x] Use the new db methods to refactor the Emergency Contact resource
- [ ] Refactoring may cause some of the tests to become redundant, these tests can be removed. Will most likely need to only test the happy path for each action. (one test per action).

- See the lease files for an example. https://github.com/codeforpdx/dwellinglybackend/blob/31610de13449c002d2394bfa7b2e48a8c12e5b1e/resources/lease.py

**Scope**

Continue to use the .json method for serializing the data.  Implementing marshmallow to use as the serializer will be completed in another ticket

gheelan:
Closes codeforpdx/dwellingly-app/issues/399

n/a
n/a
n/a


models/emergency_contact.py
```diff
@@ -9,28 +8,18 @@ class EmergencyContactModel(BaseModel):
...
-                emergency_contact_id = self.id,
-                number = number['number'],
-                numtype = number['numtype'] if 'numtype' in number.keys() else '',
-                extension = number['extension'] if 'extension' in number.keys() else '',
-            )
-            db.session.add(item)
-            self.contact_numbers.append(item)
+    contact_numbers = relationship(
+        'ContactNumberModel',
+        backref='contact_numbers',
```
NickSchimek: oops I think this is one of my commits, however this line should be:
```suggestion
        backref='emergency_contact',
```
That will allow us to ask a ContactNumber for its emergency_contact info

models/base_model.py
```diff
@@ -27,7 +27,6 @@ def delete(cls, id):
     @classmethod
     def create(cls, schema, payload):
         attrs = cls.validate(schema, payload)
-
         obj = cls(**attrs)
```
NickSchimek: This is a personal style thing: Perhaps we can add this space back in?
```suggestion

        obj = cls(**attrs)
```

models/emergency_contact.py
```diff
@@ -9,28 +8,18 @@ class EmergencyContactModel(BaseModel):
...
+        lazy=True,
+        cascade="all, delete-orphan"
+    )
 
     def json(self):
         return {
-            'id': self.id,
```
NickSchimek: This removal will be a change in behavior. I don't think we want to remove this. Is there a reason for removing it?
I believe the frontend will need to have this `id` returned even if it is not being displayed to the user. This should allow them to send us the `id` of an emergency contact when they want to perform an action on it.

resources/emergency_contacts.py
```diff
@@ -1,89 +1,67 @@
...
             contactEntry.name = data.name
-        if('description' in data.keys()):
+        if ('description' in data.keys()):
             contactEntry.description = data.description if data.description else ""
-        if(data.contact_numbers):
-            numbersData, numbersError = parseContactNumbersFromJson(request.get_json(force=True))
+
+        # TODO: We should/need to create a ContactNumber resource to update this
+        if (data.contact_numbers):
+            numbersData, numbersError = parse_contact_numbers(request.get_json(force=True))
```
NickSchimek: Since we are not going to refactor the PUT action right now we should keep everything that it depends on. If we put the `parseContactNumbersFromJson` method back in and change this back to use that method then all the tests will pass and we can get this merged.
```suggestion
            numbersData, numbersError = parseContactNumbersFromJson(request.get_json(force=True))
```

resources/emergency_contacts.py
```diff
@@ -1,89 +1,67 @@
...
-    for number in json_data["contact_numbers"]:
-        if not 'number' in number.keys(): 
-            error = "One of the contact_numbers for the emergency contact is missing a number"
-            break
-        newNumber = { "number": number['number'], "id": "unavailable" }
-        if 'id' in number.keys(): newNumber['id'] = number['id']
-        if 'numtype' in number.keys(): newNumber['numtype'] = number['numtype']
-        if 'extension' in number.keys(): newNumber['extension'] = number['extension']
-        parsedData.append(newNumber)
-    return parsedData, error
```
NickSchimek: Let's not delete this method just yet as the PUT action still needs to depend on it.

resources/emergency_contacts.py
```diff
@@ -1,89 +1,67 @@
 from flask_restful import Resource, reqparse
 from flask import request
+
 from models.emergency_contact import EmergencyContactModel
 from models.contact_number import ContactNumberModel
 from resources.admin_required import admin_required
+from schemas.emergency_contact import EmergencyContactSchema
+
+
+# Not sure if this is necessary anymore
```
NickSchimek, line 10: I'm not sure either. It would be nice to remove it, However it is fine for now. We can take a look at it again in the PR to refactor the PUT action.

schemas/contact_number.py
```diff
@@ -0,0 +1,18 @@
+from ma import ma
+from models.contact_number import ContactNumberModel
+from utils.time import time_format
+from marshmallow import fields, post_load, EXCLUDE
+
+
+class ContactNumberSchema(ma.SQLAlchemyAutoSchema):
+    class Meta:
+        model = ContactNumberModel
+        unknown = EXCLUDE
```
NickSchimek, line 10: Interesting, I wasn't aware this configuration could be set here. I thought it had to be set at initialization. I was curious so I deleted it and all the tests still pass. The BaseModel methods already initialize this accordingly. I think it is safe to remove unless there was a reason for adding it here?

NickSchimek: If it is needed, could we add a test that would fail if this line is removed?

schemas/emergency_contact.py
```diff
@@ -0,0 +1,32 @@
...
+    class Meta:
+        model = EmergencyContactModel
+
+    created_at = fields.DateTime(time_format)
+    updated_at = fields.DateTime(time_format)
+
+    contact_numbers = fields.List(fields.Nested("ContactNumberSchema"))
+
+    @validates("contact_numbers")
+    def validate_contact_nums(self, value):
```
NickSchimek, line 18: Excellent. At first glance I thought there was a bug in this logic, but after checking it out. It works fine! - The only thing missing are some validation tests in the schema test file. 
Since I already wrote a test to check my curiousity. Here is what I used:
```
    def test_validate_contact_number(self, empty_test_db):
        invalid_payload = {
            'name': 'emergency contact name',
            'contact_numbers': [{}]
        }
        validation_errors = EmergencyContactSchema().validate(invalid_payload)

        assert 'contact_numbers' in validation_errors
```

tests/integration/test_emergency_contacts.py
```diff
@@ -21,6 +21,7 @@
 
 def test_emergency_contacts_GET_all(client, test_database):
   response = client.get(endpoint)
+  print(response)
```
NickSchimek: Let's remove this print statement

schemas/emergency_contact.py
NickSchimek: Ah I see there are some integration tests for this, however those could be removed in favor of the unit tests. I'd like to see unit tests created for these new validation methods, however we could do it in a separate ticket too.

tests/integration/test_emergency_contacts.py
```diff
@@ -85,9 +100,14 @@ def test_emergency_contacts_PUT(client, auth_headers):
...
     ]
   }
+
   response = client.put(f'{endpoint}/{id}', json=updatedInfo, headers=auth_headers["admin"])
   assert is_valid(response, 200) # OK
   assert response.json['name'] == 'Greg'
+  # assert response.json['contact_numbers'][0]['number'] == '503-291-9111'
+  # This test will fail, Contact Numbers need an id in order to be updated.
+  # The request is not submitting an id.
+  # TODO: Create issue to address non restful behavior.
```
NickSchimek, line 109: Thanks for adding this. We'll need to discuss this with the team.

NickSchimek (changes requested):
This looks good really good. I think we a few updates we can get this merged. Please review my comments and let me know if there are any questions.

models/emergency_contact.py
gheelan: Whoops, I think this just got deleted on accident somehow and I didn't notice! Good catch!

schemas/contact_number.py
gheelan: It's not necessary, just an artifact from when I was trying different things that I forgot about, I will remove it! FWIW, I found in the marshmallow docs that it can be set in the Meta class, when a schema is initialized (i.e. ContactNumberSchema(unknown=INCLUDE/EXCLUDE), or passed as in argument to load()

tests/integration/test_emergency_contacts.py
gheelan: Another careless mistake, sorry!

schemas/emergency_contact.py
gheelan: So this method should be tested within the EmergencyContact schema tests?

schemas/contact_number.py
NickSchimek: Nice! That is good to know.

schemas/emergency_contact.py
NickSchimek: Yes please

NickSchimek:
It looks like there is only one failure now after your latest commit. Just need to add the `id` field into the json unit test.
`FAILED tests/unit/test_emergency_contact.py::TestEmergencyContactModel::test_emergency_contact_json`

Let me know when your ready for me to take a look at this again.

tests/schemas/test_emergency_contact_schema.py
```diff
@@ -12,3 +15,49 @@ def test_valid_payload(self, empty_test_db):
...
+        invalid_payload = {
+            'name': 'emergency contact name',
+            'contact_numbers': [{}]
+        }
+        validation_errors = EmergencyContactSchema().validate(invalid_payload)
+
+        assert 'contact_numbers' in validation_errors
+
+    def test_validate_multiple_contact_numbers(self, empty_test_db):
+        invalid_payload = {
```
NickSchimek: Perhaps this should be called `valid_payload`

tests/schemas/test_emergency_contact_schema.py
```diff
@@ -0,0 +1,63 @@
...
+            'name': 'emergency contact name',
+            'contact_numbers': [
+                {"number": "503-291-9111", "numtype": "Call"},
+                {"number": "503-555-3321", "numtype": "Text"}
+            ]
+        }
+        no_validation_errors = {}
+
+        assert no_validation_errors == EmergencyContactSchema().validate(invalid_payload)
```
NickSchimek:
```suggestion
        valid_payload = {
            'name': 'emergency contact name',
            'contact_numbers': [
                {"number": "503-291-9111", "numtype": "Call"},
                {"number": "503-555-3321", "numtype": "Text"}
            ]
        }
        no_validation_errors = {}

        assert no_validation_errors == EmergencyContactSchema().validate(valid_payload)
```

NickSchimek (approved):
Thank you! Great work 💫

tests/schemas/test_emergency_contact_schema.py
gheelan: I agree, I really need to be more careful haha, sorry!

NickSchimek: No worries, it happens to me all the time.

PR #163, opened 2020-12-05
Add instructions for pulling dev branch

NickSchimek:
Adding documentation for the necessary commands that need to be ran
after pulling down a fresh copy.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -93,3 +93,21 @@ How to contribute to this project.
 ```
 
 (Step #3 creates a new branch titled <name of branch> and navigates you to that branch)
+
+#### Updating development branch
+
+To update your development branch with the latest changes run:
+
+1. First checkout the development branch if not already checked out.
+ - Then run: `git checkout development`
+
+2. Pull the latest changes from github down to your local copy.
+ - Assuming origin is set to the GitHub Code for PDX dev branch run:
+ - `git pull origin development`
+
+3. After pulling fresh copy it is a good habit to install any new deps and rebuild the database. Run the following two commands:
+ - `pipenv run install -d`
+ - `pipenv run flask db recreate`
+
+4. Finally - run the tests to ensure everything is passing.
+ - `pipenv run pytest`
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -96,7 +96,7 @@ How to contribute to this project.
 
 #### Updating development branch
 
-To update your development branch with the latest changes run:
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -93,3 +93,21 @@ How to contribute to this project.
...
+#### Updating development branch
+
+To update your development branch with the latest changes run:
+
+1. First checkout the development branch if not already checked out.
+ - Then run: `git checkout development`
+
+2. Pull the latest changes from github down to your local copy.
+ - Assuming origin is set to the GitHub Code for PDX dev branch run:
+ - `git pull origin development`
```
keppel2, line 106: I just tried this and it didn't work. It should be `git pull`.

keppel2: Scratch that. `git pull` was just puling in other branches!

jls-github (commented):
I'm sure this will be helpful for people! 

We may want to change the language a tad. Line 99 reads better without the word "run".

One thing that may be helpful to newer contributors as well is some info on merging a development branch into a feature branch

NickSchimek:
> I'm sure this will be helpful for people!

Thanks!

> We may want to change the language a tad. Line 99 reads better without the word "run".

Excellent, I will remove that.

> One thing that may be helpful to newer contributors as well is some info on merging a development branch into a feature branch

I disagree with this, for two reasons:

1. Merging the main brain into a feature branch is bad practice, and should not be done for various reasons. Hence, I do not want to add documentation that encourages its use. A rebase is more appropriate in this situation.
2. Merging should usually only be done in the other direction. A feature branch is merged into the development (main) branch.

CONTRIBUTING.md
```diff
@@ -93,3 +93,21 @@ How to contribute to this project.
 ```
 
 (Step #3 creates a new branch titled <name of branch> and navigates you to that branch)
+
+#### Updating development branch
+
+To update your development branch with the latest changes run:
```
NickSchimek: @jls-github suggestion.

```suggestion
To update your development branch with the latest changes:
```

jls-github:
> > One thing that may be helpful to newer contributors as well is some info on merging a development branch into a feature branch
> 
> I disagree with this, for two reasons:
> 
> 1. Merging the main brain into a feature branch is bad practice, and should not be done for various reasons. Hence, I do not want to add documentation that encourages its use. A rebase is more appropriate in this situation.
> 2. Merging should usually only be done in the other direction. A feature branch is merged into the development (main) branch.

After doing some research, I see that you are definitely right about this!

PR #164, opened 2020-12-05
Create unit tests for ticket model

### LINKED ISSUE #413: Create unit tests for ticket model
**Purpose**
public methods should have tests

**Acceptance Criteria**
- [ ] Add the BaseInterfaceTest to `tests/unit/test_ticket.py` and create unit tests for all public methods. see [this commit](https://github.com/codeforpdx/dwellinglybackend/blob/31610de13449c002d2394bfa7b2e48a8c12e5b1e/tests/unit/test_lease.py) for an example.

 - The init method does not need to be tested, and the `test_notes` method can be removed. Another issue is addressing tests for the notes model

cecillesalazar:

Closes codeforpdx/dwellingly-app/issues/413

Requires FE adjustments - see https://github.com/codeforpdx/dwellingly-app/issues/465

I made a change to the Ticket model clear a marshmallow error I bumped into. It's noted in the commit messages. Not entirely sure why my fix resolved it to be honest - but the fix was to change the arg name in the model from `tenant` to `tenantID`.

Not the worst unnecessary change per se, since there is some ambiguity in the terms for the returning json. We refer to `tenant` to be the tenant name (instead of being explicit by using tenantName) and do use `tenantID` to refer to the ID. 

Just noting it in case anyone has a better idea - since it did require some more reference changes than I'd like

Had to make changes to the seedData so probably.

Nope

Nope


schemas/ticket.py
```diff
@@ -0,0 +1,33 @@
...
+
+    @validates("assignedUser")
+    def validate_assigned_user(self, value):
+        if not UserModel.query.get(value):
+            raise ValidationError(f"{value} is not a valid user ID")
+
+    @validates("sender")
+    def validate_sender(self, value):
+        if not UserModel.query.get(value):
+            raise ValidationError(f"{value} is not a valid user ID")
```
NickSchimek: Very nice! Could we get some unit tests created for the custom validation methods?

resources/tickets.py
```diff
@@ -88,7 +88,14 @@ def get(self):
     @jwt_required
     def post(self):
         data = Tickets.parser.parse_args()
-        ticket = TicketModel(**data)
+        ticket = TicketModel(
+            issue=data.issue,
+            tenantID=data.tenant,
```
NickSchimek: Lets get an issue created for the FE to update the parameter names they're sending to us.

tests/factory_fixtures/ticket.py
```diff
@@ -0,0 +1,25 @@
...
+    yield _ticket_attributes
+
+@pytest.fixture
+def create_ticket(ticket_attributes, create_tenant, create_admin_user, create_join_staff):
+    def _create_ticket(issue="Leaky pipe"):
+        tenant = create_tenant()
+        ticket = TicketModel(**ticket_attributes(issue, tenant, create_admin_user(), create_join_staff()))
+        ticket.save_to_db()
+        return ticket
+    yield _create_ticket
```
NickSchimek, line 25: Very nice! Its good practice to create tests for the factories. If something changes in the future that breaks this ticket factory, then every test that depends on this will probably start to fail. Having an additional test failure that will clearly be a different flavor from all the rest that says "I couldn't create a ticket factory" can be helpful.

schemas/ticket.py
```diff
@@ -0,0 +1,33 @@
...
+from utils.time import time_format
+from marshmallow import fields, validates, ValidationError
+
+
+class TicketSchema(ma.SQLAlchemyAutoSchema):
+    class Meta:
+        model = TicketModel
+        include_fk = True
+
+    tenant = fields.Nested("TenantSchema")
```
NickSchimek, line 14: It looks like the TicketModel field names should all be updated. And I don't think the Tenant has its relationship set accordingly to need this line right now, but it is OK to leave it in. I believe this is why you were seeing errors. `tenant` was already defined as a field column to hold an integer, however now it is telling Marshmallow it is a Nested field. So it was contradictory.

schemas/ticket.py
```diff
@@ -0,0 +1,33 @@
...
+    user = fields.Nested("UserSchema")
```
NickSchimek, line 15: Same here. I don't think the relationships are setup on User, however lets leave this in as we will want it.

NickSchimek (changes requested):
This looks really good @cecillesalazar 

I looked at the ticket model and all of the FK columns should have an `id` suffix, because it is an id column and we would want to reserve the field name without an `id` suffix so that we could use that to refer to the actual resource.

For example.
`ticket = TicketModel.query.get(1)` # <- returns a ticket object
And we can ask it for the `tenantID`
`ticket.tenantID` # <- return an integer value (id)
And if the relationships are set up we can ask it for the tenant
`ticket.tenant` # <- returns a TenantModel object (specifically the tenant with the id `tenantID`)

I think you were seeing errors because in the schema the `tenant` field was being declared twice. Once by the AutoSchema to expect an integer field, and then again where the Nested field is being declared to expect an object.

To take care of any future issues can we update all the FK's in this PR on the Ticket Model to have the `id` suffix, and then let's get either a PR submitted or an issue created to update the FE. I think we will also need to update the README `using-the-api` file since we're changing what we want to receive.

schemas/ticket.py
cecillesalazar: Good call! Added some in the unit tests folder~

resources/tickets.py
cecillesalazar: I'll create an issue and link it to this PR~

tests/factory_fixtures/ticket.py
cecillesalazar: Good call - added a basic one in the integrations folder~

schemas/ticket.py
cecillesalazar: Doh! Yep, yep - I see in the ticket model that we established the relationship for notes `notes = db.relationship(...)` - but not for tenant~

tests/integration/test_tickets.py
```diff
@@ -141,3 +142,10 @@ def test_tickets_DELETE(client, auth_headers):
     # NOT FOUND - Trying to delete a non-existing ticket or an already deleted ticket
     assert is_valid(response, 404)
     assert response.json == {'message': 'Ticket not found'}
+
+@pytest.mark.usefixtures("empty_test_db")
+class TestFixtures:
+    def test_create_ticket(self, create_ticket):
+        ticket = create_ticket()
+        assert ticket
+        assert ticket.issue == "Leaky pipe"
```
NickSchimek, line 151: This is probably better placed in the `tests/unit/test_ticket.py` file since we're not making a request through an endpoint and this more of a unit test in the first place. I do like how you broke it out into a separate class so that it is easy to see that this is clearly a different flavor of tests from the rest in the file and the class name tells us exactly what it is doing, testing fixtures. Let's get this merged and we can always move it later the next time if need be.

tests/unit/test_ticket.py
```diff
@@ -1,45 +1,26 @@
...
+        assert invalid_tenant_validation_error == TicketSchema().validate(invalid_tenant)
+
+    def test_validate_assigned_user_and_sender(self):
+        invalid_assigned_user_and_sender = {
+            'assignedUserID': 777,
+            'senderID': 888
+        }
+
+        invalid_assigned_user_validation_error = {'assignedUserID': ['777 is not a valid user ID'], 'senderID': ['888 is not a valid user ID']}
+        assert invalid_assigned_user_validation_error == TicketSchema().validate(invalid_assigned_user_and_sender)
```
NickSchimek, line 26: Very nice. I think these tests are better suited in the `tests/schemas/test_ticket_schema.py` since it is testing what the schema does (input validation), and not testing what the model does (ticket creation/querying). However, let's get this merged and we can just make a mental note to move these when appropriate.

NickSchimek (approved):
Thanks @cecillesalazar Great work!

PR #165, opened 2020-12-06
Issue438 update db tenant

### LINKED ISSUE #438: Update db for Tenant Model
**Purpose**
The following fields should be required:
`firstName`
`lastName`
`phone`

**Proposed Solution**
add `nullable=False` to the three fields

aedwardg:

Closes codeforpdx/dwellingly-app/issues/438

- Ensures `firstName`, `lastName` and `phone` fields are required for a new Tenant.
- Validates existence and length of those three fields
- Tests validations with pytest

### DIFF
--- a/schemas/tenant.py
+++ b/schemas/tenant.py
@@ -8,8 +8,8 @@ class TenantSchema(ma.SQLAlchemyAutoSchema):
     class Meta:
         model = TenantModel
 
-    created_at = fields.DateTime(time_format, required=True)
-    updated_at = fields.DateTime(time_format, required=True)
+    created_at = fields.DateTime(time_format)
+    updated_at = fields.DateTime(time_format)
 
     firstName = fields.Str(
         required=True,
--- a/schemas/tenant.py
+++ b/schemas/tenant.py
@@ -1,13 +1,30 @@
 from ma import ma
 from models.tenant import TenantModel
-from marshmallow import fields
+from marshmallow import fields, validate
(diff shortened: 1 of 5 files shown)


schemas/tenant.py
```diff
@@ -1,13 +1,30 @@
...
 
-  addedOn = fields.DateTime(time_format)
-  created_at = fields.DateTime(time_format)
-  updated_at = fields.DateTime(time_format)
+    created_at = fields.DateTime(time_format)
+    updated_at = fields.DateTime(time_format)
+
+    firstName = fields.Str(
+        required=True,
+        validate=validate.Length(min=1, max=100),
```
NickSchimek, line 16: I'm not familiar with this method. It seems assigning it to validate has some automatic message handling happening? Is that correct?

NickSchimek (approved):
Thanks for the PR. This looks really good. 💫

schemas/tenant.py
aedwardg: Yes, I just used one of the built-in validators for this:
https://marshmallow.readthedocs.io/en/stable/marshmallow.validate.html#api-validators

Looking at the source code, each one has built-in message(s) if you don't provide the optional `error` argument:
https://marshmallow.readthedocs.io/en/stable/_modules/marshmallow/validate.html#Length

NickSchimek: good to know! Thank you.

PR #166, opened 2020-12-09
route to allow archival of multiple properties at once.

### LINKED ISSUE #466: New route to archive a list of properties
**Purpose**
The Manage Properties page on the front end allows a bulk edit functionality to archive properties (#440). This feature request is to provide a backend route to support multi-property archival

**Acceptance Criteria**
- [x] Add resource `api/properties/archive`  
- [x] Accepts a POST request with the following body: `{ ids: [<string:ids>] }`
- [x] admin required
- [x] 201 confirmation upon success
- [x] Endpoint added to USING_THE_API.md

DaveTrost:
Closes codeforpdx/dwellingly-app/issues/466

### DIFF
--- a/resources/property.py
+++ b/resources/property.py
@@ -61,9 +61,12 @@ class ArchiveProperties(Resource):
     parser.add_argument('ids', action='append')
 
     @admin_required
-    def post(self):
+    def patch(self):
         data = ArchiveProperties.parser.parse_args()
         propertyList = []
+
+        if not ('ids' in data and type(data['ids']) is list):
+            return{'message': 'Property IDs missing in request'}, 400
 
         for id in data['ids']:
             property = PropertyModel.find_by_id(id)
@@ -72,10 +75,9 @@ class ArchiveProperties(Resource):
             property.archived = True
             propertyList.append(property)
 
-        for p in propertyList:
-            p.save_to_db()
-
-        return {'properties': [p.json() for p in PropertyModel.query.all()]}, 201
+        db.session.bulk_save_objects(propertyList)
+        db.session.commit()
+        return {'properties': [p.json() for p in PropertyModel.query.all()]}, 200
 
 # single property/name
 class Property(Resource):
--- a/resources/property.py
+++ b/resources/property.py
@@ -75,7 +75,7 @@ class ArchiveProperties(Resource):
         for p in propertyList:
             p.save_to_db()
 
-        return {'archived': [p.id for p in propertyList]}, 201
+        return {'properties': [p.json() for p in PropertyModel.query.all()]}, 201
 
 # single property/name
 class Property(Resource):
--- a/resources/property.py
+++ b/resources/property.py
@@ -3,7 +3,6 @@ from flask_restful import Resource, reqparse
 from resources.admin_required import admin_required
 from db import db
 from models.property import PropertyModel
-from models.user import UserModel
 
 # | method | route                | action                     |
 # | :----- | :------------------- | :------------------------- |
@@ -56,6 +55,27 @@ class ArchiveProperty(Resource):
         property.save_to_db()
 
         return property.json(), 201
+
+class ArchiveProperties(Resource):
+    parser = reqparse.RequestParser()
+    parser.add_argument('ids', action='append')
+
+    @admin_required
+    def post(self):
+        data = ArchiveProperties.parser.parse_args()
+        propertyList = []
+
+        for id in data['ids']:
+            property = PropertyModel.find_by_id(id)
+            if(not property):
+                return{'message': 'Properties cannot be archived'}, 400
+            property.archived = True
+            propertyList.append(property)
+
+        for p in propertyList:
+            p.save_to_db()
+
+        return {'archived': [p.id for p in propertyList]}, 201
 
 # single property/name
 class Property(Resource):
--- a/tests/factory_fixtures/property.py
+++ b/tests/factory_fixtures/property.py
@@ -2,18 +2,28 @@ import pytest
 from models.property import PropertyModel
 
 @pytest.fixture
-def create_property(create_property_manager):
-    def _create_property(pm=create_property_manager()):
-        property = PropertyModel(
-                name='the heights',
-                address='111 SW Harrison',
-                city="Portland",
-                unit="101",
-                state="OR",
-                zipcode="97207",
-                propertyManagerIDs=[pm.id],
-                archived=False
-            )
+def property_attributes(faker):
+    def _property_attributes(archived=False, pm=None):
+        return {
+            'name': faker.unique.name(),
+            'address': faker.address(),
+            'city': faker.city(),
+            'unit': faker.building_number(),
+            'state': faker.state() ,
+            'zipcode': faker.postcode(),
+            'propertyManagerIDs': [pm.id] if pm and pm.id else [],
+            'archived': archived,
+        }
+    yield _property_attributes
+
+@pytest.fixture
+def create_property(property_attributes, create_property_manager):
+    def _create_property():
+        property = PropertyModel(**property_attributes(
+            pm=create_property_manager(),
+            archived=False,
+        ))
         property.save_to_db()
         return property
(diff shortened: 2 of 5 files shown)


DaveTrost:
Wait! I found an issue. The parse_args function is only extracting the first id in the sent array. I'm working to figure out why

*EDIT*: fixed it and rebased to squash commits

DaveTrost:
@NickSchimek Can you take a look at this PR? 

Note: It failed Travis with a testing failure, and it's not for a file where I made changes. It appears that there is some time sensitivity in the test_tickets check suite. The assertion error was `AssertionError: assert '12/10/2020 06:54:22' == '12/10/2020 06:54:21'`

NickSchimek:
This is an intermittent failure described here https://github.com/codeforpdx/dwellingly-app/issues/405 - It would be pretty easy to fix this error, but then again perhaps someone else can learn from trying to fix it. Eventually it will get closed. Until then I just rebuild the test when I see this error.

resources/property.py
```diff
@@ -57,6 +56,27 @@ def post(self, id):
 
         return property.json(), 201
 
+class ArchiveProperties(Resource):
+    parser = reqparse.RequestParser()
+    parser.add_argument('ids', action='append')
+
+    @admin_required
+    def post(self):
```
NickSchimek: Perhaps a PUT or PATCH is more appropriate as we are updating existing properties.

resources/property.py
```diff
@@ -57,6 +56,27 @@ def post(self, id):
...
+        data = ArchiveProperties.parser.parse_args()
+        propertyList = []
+
+        for id in data['ids']:
+            property = PropertyModel.find_by_id(id)
+            if(not property):
+                return{'message': 'Properties cannot be archived'}, 400
```
NickSchimek, line 74:
```suggestion
            property = PropertyModel.find(id)
```
Using the find method will automatically return a 400 if the property cannot be found.

resources/property.py
```diff
@@ -57,6 +56,27 @@ def post(self, id):
...
+            property.archived = True
+            propertyList.append(property)
```
NickSchimek, line 76: I don't think we need to append the properties to a separate list as we're updating it.

resources/property.py
```diff
@@ -57,6 +56,27 @@ def post(self, id):
...
+
+        for p in propertyList:
+            p.save_to_db()
```
NickSchimek: We should be able to do something like this?
```suggestion
       db.session.bulk_save_objects(propertyList)
```

resources/property.py
```diff
@@ -57,6 +56,27 @@ def post(self, id):
...
+
+        return {'properties': [p.json() for p in PropertyModel.query.all()]}, 201
```
NickSchimek: We are not creating anything here so I think a 200 response is more appropriate.

tests/integration/test_properties.py
```diff
@@ -67,6 +67,43 @@ def test_archive_property_by_id(client, auth_headers, new_property, test_databas
     assert responseBadPropertyID.get_json() == {'message': 'Property cannot be archived'}
     assert responseBadPropertyID.status_code == 400
 
+def test_archive_property_by_id(client, auth_headers, new_property, test_database):
```
NickSchimek: I'd prefer to use the valid_header and empty_test_db factory for all new tests. This will help keep the tests running as fast as possible and prevent issues with tests failing to due changes in the seed file. It doesn't look like the new_property factory is being used here. However, there is a property factory that can be used with the empty test db (it probably needs to be updated to use faker). `tests/factory_fixtures/property.py`
```suggestion
def test_archive_property_by_id(client, valid_header, create_property, empty_test_db):
```

NickSchimek (changes requested):
Thanks for the PR @DaveTrost - Overall I think this looks good. However, I did comment on a few things I would like to see addressed. 

Preferably, for all new functionality I would like to see marshmallow used instead of the reqparser. The reqparser is being deprecated by the flask team, and we are refactoring everything to use marshmallow instead. Since this is new functionality it would be nice to use marshmallow here; otherwise someone else will need to refactor this soon. 

If you would like a walk through on how the backend is setup now, feel free to ping me. I'm planning on getting some documentation added as well.

resources/property.py
NickSchimek: Apparently I did this exact same architecture recently. It is probably more readable this way though, and we probably wont be dealing with huge lists anyway.

resources/property.py
DaveTrost: I agree, that would be more restful. In the spirit of PATCH being a set of instructions for change, I'm making it a PATCH. In this case it is an idempotent application of PATCH because no matter how many times you hit this archive route, it will end up with the listed properties having their archived field set to True.

resources/property.py
DaveTrost: Thank you for the tip!

resources/property.py
DaveTrost: Sure thing. I changed it

DaveTrost:
> Preferably, for all new functionality I would like to see marshmallow used instead of the reqparser. The reqparser is being deprecated by the flask team, and we are refactoring everything to use marshmallow instead. Since this is new functionality it would be nice to use marshmallow here; otherwise someone else will need to refactor this soon.

I see that a refactor of this route is already planned (codeforpdx/dwellingly-app/issues/403). And, this PR is blocking a front-end PR (codeforpdx/dwellingly-app/pull/467) so I prefer to move forward with reqparser intact. When there are more examples to work from for implementing Marshmallow, I would love to pick up issue 403 and do the refactor.

NickSchimek:
> When there are more examples to work from for implementing Marshmallow, I would love to pick up issue 403 and do the refactor.

Sounds good

tests/integration/test_properties.py
DaveTrost: I'm not sure I've done the most efficient job of this, but the tests run now and they use the new fixtures `admin_header`, `empty_test_db` and `create_property`
Also, I added faker to the create_property factory
@NickSchimek - you are welcome to check it out!

tests/factory_fixtures/property.py
```diff
@@ -2,18 +2,28 @@
...
+            'state': faker.state() ,
+            'zipcode': faker.postcode(),
+            'propertyManagerIDs': [pm.id] if pm and pm.id else [],
+            'archived': archived,
+        }
+    yield _property_attributes
+
+@pytest.fixture
+def create_property(property_attributes, create_property_manager):
+    def _create_property():
```
NickSchimek, line 21: This looks excellent. 
Probably keep in mind that we may want to update this to allow the developer to specify specific property managers when creating a property, but that road can be crossed if the need arises.

tests/integration/test_properties.py
```diff
@@ -67,41 +68,49 @@ def test_archive_property_by_id(client, auth_headers, new_property, test_databas
...
-    assert responseBadPropertyID.get_json() == {'message': 'Properties cannot be archived'}
-    assert responseBadPropertyID.status_code == 400
+@pytest.mark.usefixtures('client_class', 'empty_test_db')
+class TestPropertyArchivalMethods:
+    def test_archive_properties_non_admin(self, staff_header):
+        """The server responds with a 401 error if a non-admin tries to archive"""
+        responseNoAdmin = self.client.patch(f'/api/properties/archive', json={}, headers=staff_header)
+        assert responseNoAdmin == 401
+
+    def test_archive_properties(self, admin_header, create_property):
```
NickSchimek, line 78: FYI: The `admin_header` takes extra time to run, because it hits the auth endpoint each time it is used. So technically when the `auth_header` is used we are testing more than one endpoint. The `valid_header` was created to bypass that for speed and to ensure only the endpoint under test is being hit. 

The idea/goal is that the `valid_header` will be used for testing behavior. And then specific `admin_header`, `staff_header`, `pm_header` will be used for auth tests.

NickSchimek (approved):
Thanks @DaveTrost !

PR #167, opened 2020-12-10
Created Notes schema and added tests for schema

### LINKED ISSUE #457: Create AutoSchema for the Notes Model
**Purpose**
Create Schema for the Notes Model

yunnie:

Closes codeforpdx/dwellingly-app/issues/457

### Additional comments
Managing serialization and deserialization was a pain because the Notes model uses the name ```user``` for user_id, but serializes ```user``` as ```"{firstname} {lastname}"```, but deserialization required ```user``` to be an integer value for the foreign key. This required creating a relationship in the ```NotesModel```, and over writing the ```user``` field in the ```NotesSchema``` to include the appropriate serialization and deserialization. 

An alternative is to change the ```user``` field in the ```NotesModel``` to ```userid```, but this would break so many things it's just too painful to consider.


schemas/notes.py
```diff
@@ -0,0 +1,29 @@
...
+from models.user import UserModel
+from utils.time import time_format
+from marshmallow import fields, validates, ValidationError
+
+class NotesSchema(ma.SQLAlchemyAutoSchema):
+    class Meta:
+        model = NotesModel
+        include_fk = True
+        include_relationships = True
+        exclude = ["userinfo"]
```
NickSchimek: I believe this only affects serialization. I would prefer to leave this out since we can specify exclude parameters at initiation of the schema. Of course that is just my preference. We could discuss this more too.

models/notes.py
```diff
@@ -14,6 +14,9 @@ class NotesModel(BaseModel):
     ticketid = db.Column(db.Integer, db.ForeignKey('tickets.id'),
         nullable=False)
 
+    userinfo = db.relationship('UserModel',
+                               backref=db.backref('Notes', lazy=True))
```
NickSchimek: on line 12 let's rename the `user` field to `user_id` and then let's name this relationship `user`. And `Notes` should be all lowercase.

tests/factory_fixtures/notes.py
```diff
@@ -0,0 +1,31 @@
...
+    def _note_attributes(text, user, ticket):
+        return {
+            'text': text,
+            'user': user.id,
+            'ticketid': ticket.id
+        }
+    yield _note_attributes
+
+@pytest.fixture
+def create_note(note_attributes, create_join_staff):
```
NickSchimek: These two factories here don't seem to be in use. Is there a reason you decided to not use them, and instead create Fakes?

tests/factory_fixtures/notes.py
```diff
@@ -0,0 +1,31 @@
...
+    class SimpleTicket:
```
NickSchimek: Technically, if we're going to use Fakes. Then these should have tests too, however there is not really anything to test in these fakes, and that in itself raises some flags for me. When a Note factory is used then it is going to be referring to a User and Ticket that are not true representations of what they're doubling. And that could cause issues for a developer when they need the `note.user` or the `note.ticket` when testing. I realize that it is possible to pass in a real user and ticket with this structure, however it is common practice for a factory to return real objects by default.

models/notes.py
NickSchimek: And line 12 could also use a `nullable=False`

tests/schemas/test_note_schema.py
```diff
@@ -0,0 +1,39 @@
+import pytest
+from schemas.notes import *
+from models.notes import NotesModel
+
+
+@pytest.mark.usefixtures("empty_test_db")
+class TestNotesSchemaSerialization:
```
NickSchimek, line 7: Thanks for adding tests! These tests look good. I'm not sure if the deserialization test is all that useful since it is testing default marshmallow behavior, however I don't see any harm in leaving it in for now.

tests/schemas/test_note_schema.py
```diff
@@ -0,0 +1,39 @@
...
+        note = note_schema.load(payload)
+        assert payload['user'] == note['user']
+        assert payload['text'] == note['text']
+        assert payload['ticketid'] == note['ticketid']
+
+    def test_notes_user_validation(self):
+        note_schema = NotesSchema(exclude=["userinfo"])
+        payload = { 'user': 500,
+                    'text': "This should fail",
+                    'ticketid': 1
```
NickSchimek: Can we add a validation test for the ticketid? If the ticketid does not exist then it should also return a validation error.

tests/schemas/test_note_schema.py
```diff
@@ -0,0 +1,39 @@
...
+                   'ticketid': 1}
+        note = note_schema.load(payload)
+        assert payload['user'] == note['user']
+        assert payload['text'] == note['text']
+        assert payload['ticketid'] == note['ticketid']
+
+    def test_notes_user_validation(self):
+        note_schema = NotesSchema(exclude=["userinfo"])
+        payload = { 'user': 500,
+                    'text': "This should fail",
```
NickSchimek: This text comment is a bit confusing. When I first read it I thought there I would see an assertion on the text field, however it ended up checking for a valid user. Also the text field is not required for a valid_payload. I think it would be reasonable to add `nullable=False` to the text field in the model.

schemas/notes.py
```diff
@@ -0,0 +1,29 @@
...
+    created_at = fields.DateTime(time_format)
+    updated_at = fields.DateTime(time_format)
+
+    userinfo = fields.Nested("UserSchema")
+    user = fields.Method("showFullName", deserialize = "getUserId")
+    def showFullName(self, obj):
+        return "%s %s" % (obj.userinfo.firstName, obj.userinfo.lastName)
+
+    def getUserId(self, value):
+        return int(value)
```
NickSchimek: It seems you know more about marshmallow than I do at this point. I'm confused by the lack of decorators on the two methods above. Where are the `obj` and `value` parameters coming from?

NickSchimek (changes requested):
I think I have one or two suggested changes, and one or two questions. Overall this looks good. Thank you for being thorough.

models/notes.py
NickSchimek: Just realized why this was bugging me. This relationship doesn't belong here. We still need to change line 12 to `user_id`. However, this relationship should be defined on the UserModel.  See: https://flask-sqlalchemy.palletsprojects.com/en/2.x/models/#one-to-many-relationships for more info.

schemas/notes.py
yunnie: You are correct. the exclude only affects serialization, and can be specified when creating a schema object. Leaving it out does make it more flexible.

models/notes.py
yunnie: Renamed ```user``` to ```userid``` in the ```NoteModel``` to be consistent with ```ticketid```. 

Moving the relationship to the UserModel makes sense.

tests/factory_fixtures/notes.py
yunnie: I initially did this to avoid creating a ticket fixture, since someone else was working on that.

tests/factory_fixtures/notes.py
yunnie: I tried rebasing to so that I could use the ticket fixture. But I get a fixture 'faker' not found error. This seems to be due to the 'user_attributes' function in the user factory fixtures. Am I missing something?

tests/schemas/test_note_schema.py
yunnie: I initially included the deserialization test because I was over writing the default behavior for 'user'. But I've since change 'user' to 'userid', so this test isn't necessary any more.

tests/factory_fixtures/notes.py
yunnie: Rerunning ```pipenv install -d``` fixed the issue.

tests/schemas/test_note_schema.py
yunnie: Changed the text to indicate the payload is simply to validate the user.

tests/schemas/test_note_schema.py
yunnie: Added validation for the ticketid

schemas/notes.py
yunnie: It's not obvious from the documentation, but ```obj``` refers to the class object passed. In most Python code, this would be ```self```. And ```value``` refers to the value in a for a given json key. So if an json object contains ```{id: 1, key: arbitrary}```, and you were over-writing the deserialization for ```id```, the ```value``` would be ```1```.

models/user.py
```diff
@@ -40,6 +42,9 @@ class UserModel(BaseModel):
     archived = db.Column(db.Boolean)
     lastActive = db.Column(db.DateTime, default=datetime.utcnow)
 
+    note = db.relationship(models.notes.NotesModel,
```
NickSchimek: A user can have many notes. Perhaps this should be pluralized. `notes = ...`

models/notes.py
```diff
@@ -9,19 +9,20 @@ class NotesModel(BaseModel):
...
     def json(self):
-        user = UserModel.find_by_id(self.user)
+        userobj = models.user.UserModel.find_by_id(self.userid)
 
         return {
             'id':self.id,
             'ticketid': self.ticketid,
             'text': self.text,
-            'user': user.full_name(),
+            'user': userobj.full_name(),
```
NickSchimek: Now that the relationship is setup., We should be able to do.
```suggestion
            'user': self.user.full_name(),
```
And then line 19 can be removed.

yunnie: I didn't realize the relationship made that possible.

schemas/notes.py
```diff
@@ -0,0 +1,29 @@
...
+from models.tickets import TicketModel
+from utils.time import time_format
+from marshmallow import fields, Schema, validates, ValidationError
+
+class NotesSchema(Schema):
+    id = fields.Integer()
+    text = fields.Str()
+    userid = fields.Integer()
+    ticketid = fields.Integer()
+
```
NickSchimek: Now that I see this. I really don't like it. haha

NickSchimek: I think we need to add the `require=True` for the `userid` and `ticketid` - I believe the auto-schema would've done that automagically.

NickSchimek (changes requested):
Now that we're not using the auto-schema, I think we need to set required=True on the fields that have nullable=False set at the model.

NickSchimek (approved):
Thanks for all your hard work on this @yunnie  - I think we could probably use two more validation tests to check those required fields, but let's get this merged. It would be nice to use this PR as a talking point so that we can compare and contrast the pro's and con's of the auto schema in our next meeting.

PR #157, opened 2020-11-26
Refactor RegisterUser endpoint

NickSchimek:
Refactored UserRegister to use marshmallow for input validation. While
refactoring found a bug that would allow a bad actor to register as any
role (admin). Patched this to raise an error if the 'role' field is
submitted to the UserRegister endpoint.

Fixes codeforpdx/dwellingly-app/issues/445

PR #168, opened 2020-12-10
Add required fields for registering user

NickSchimek:
The firstName, lastName, and phone should be required when a user
registers. Also updated and refactored the tests for the user schema.

This addresses some missing required fields I noticed after a self review of PR #157


tests/integration/test_users.py
```diff
@@ -268,23 +269,22 @@ def test_invite_user(self, valid_header):
...
+        payload = self.valid_payload(user_attributes())
+        response = self.client.post(self.endpoint, json=payload)
 
         mock_create.assert_called_with(
             schema=UserRegisterSchema,
-            payload=self.valid_payload
-        )
\ No newline at end of file
+            payload=payload
+        )
```
aedwardg, line 290: There should be a blank line at the end of the file.  There wasn't one before, but you might as well add one in before merging this PR.

tests/schemas/test_user_schema.py
```diff
@@ -10,17 +18,28 @@ def test_unique_email(self, empty_test_db, create_join_staff):
...
-    pass
+class TestUserSchemaValidations(UserSchemaValidations):
+    def setup(self):
+        self.schema = UserSchema
 
 
-class TestUserRegisterSchemaValidations(UserSchemaValidations):
+class TestInheritedUserRegisterSchemaValidations(UserSchemaValidations):
+    def setup(self):
+        self.schema = UserRegisterSchema
```
aedwardg, line 34: What are these classes doing? Am I missing something, or do these still need to be fleshed out?

If these don't do anything yet, it might be worth throwing in a couple of comments explaining what should go there for the time being, and then opening up a corresponding issue.

Other than that, the changes look good.

aedwardg (commented):
This looks good.  I suggested a couple of changes in my comments, but there would be no problem merging this as-is.

tests/integration/test_users.py
NickSchimek: Totally agree

tests/schemas/test_user_schema.py
NickSchimek: Thanks for asking @aedwardg ! These are absolutely doing something. They're confirming the contract between the `UserRegisterSchema` and the `UserSchema`. 

In the `schemas/user.py` file there is a `UserSchema` and a `UserRegisterSchema`. The `UserSchema` has a validation for the `email` field that the `UserRegisterSchema` inherits and must validate too.

As a general testing principle, inherited behavior must be tested to ensure the inherited class does nothing to override behavior that should not be overridden. To do this the `UserSchemaValidations` class was created to assert the email validations (and any other validations that get added in the future). This class does not start with the word `Test` so it is not ran by pytest.

The two classes above that appear to do nothing actually executes the tests defined in `UserSchemaValidations` because they include all the methods defined in that class through inheritance. 

Originally I made a mistake when I set this up since the test method in the `UserSchemaValidations` was always calling the `UserSchema`. So now it is set up correctly and injects the schema's we need to use for the email validation test as seen on line 21.

NickSchimek: Let me know if this explanation helps or not.

tests/integration/test_users.py
NickSchimek: hmmm this has a newline at the end of the file. I checked the other files just in case and they all do too.

tests/schemas/test_user_schema.py
aedwardg: @NickSchimek: 

So, just to make sure I understand this, `UserSchemaValidations` is not run by pytest because it doesn't start with the word "Test," but since these two classes _do_ start with "Test" and they inherit from `UserSchemaValidations`, pytest runs them and also invokes the inherited `test_unique_email()` method for each using the `self.schema` you provide in each class's `setup()` method?

If that's it, I think I understand, but it definitely feels like I'm in the movie Inception 😄

NickSchimek: Yep, you got it! That is one of OOPs strongest features. As long as the walls don't start falling down - I think you'll be good.

tests/integration/test_users.py
aedwardg: It's not showing up on my machine or on GitHub.
https://github.com/codeforpdx/dwellinglybackend/blob/ns-update_user_schema_tests/tests/integration/test_users.py

Maybe it has it in your local code?

NickSchimek: I don't see GitHub flagging for a missing newline character. Feel free to push up a change.

aedwardg: @NickSchimek just realized I was looking at this file on the _development branch_ locally 🤦‍♂️ 
Your branch has the newline. Sorry about that! 😬

NickSchimek: 😆  no worries. It happens to the best of us.

aedwardg (approved):
Looks good to me.

PR #169, opened 2020-12-11
Remove hard-coded role verification

### LINKED ISSUE #433: Remove hard-coded role dependency when verifying a role exists
**Issue**
There is a validation for checking if the payload contains a valid role integer. Update this check so that we can send a message to the RoleEnum to check for us. For example something like:

```
if RoleEnum.has_role(role_query):
...
```

The validation check is located here: https://github.com/codeforpdx/dwellinglybackend/pull/110/files#diff-517c606412b518afde6c3c486a06070eb1edce1ce32ddadd3cfa46e04d5c2ebeR215

This is needed because a role will not always be in range of 0 to 5. since a couple roles that are currently implemented will be removed and the role integer values may not always be consecutive.

aedwardg:
models/user.py:
- Added class methods to RoleEnum to retrieve member values and check if role exists
- Removed duplicate import and blank line

resources/user.py:
- Removed hard-coded integer range from validation

tests/integration/test_users.py:
- Changed unknown-role value from 5 to the max role value plus 1

Closes codeforpdx/dwellingly-app/issues/433

### DIFF
--- a/models/user.py
+++ b/models/user.py
@@ -6,7 +6,6 @@ import jwt
 from db import db
 from enum import Enum
 from models.base_model import BaseModel
-from flask import current_app
 from jwt import ExpiredSignatureError
 from utils.time import Time
 
@@ -16,6 +15,15 @@ class RoleEnum(Enum):
     PROPERTY_MANAGER = 2
     STAFF = 3
     ADMIN = 4
+
+    @classmethod
+    def get_values(cls):
+        return [member.value for name, member in cls.__members__.items()]
+
+    @classmethod
+    def has_role(cls, role):
+        role_values = cls.get_values()
+        return role in role_values
 
 
 class UserModel(BaseModel):
@@ -113,4 +121,3 @@ class UserModel(BaseModel):
             self.hash_digest = UserModel._hash_pw(self.password)
         db.session.add(self)
         db.session.commit()
-
--- a/resources/user.py
+++ b/resources/user.py
@@ -203,7 +203,7 @@ class Users(Resource):
 
         role_query = int(request.args["r"])
 
-        if role_query not in range(0, 5):
+        if not RoleEnum.has_role(role_query):
             return {"message": "Invalid role"}, 400
 
         role = RoleEnum(role_query)
--- a/tests/integration/test_users.py
+++ b/tests/integration/test_users.py
@@ -239,7 +239,8 @@ def test_get_user(client, auth_headers, new_user):
 
     """Queries with a non-existing role returns a 400 response"""
 
-    unknown_user_response = client.get('api/user?r=5', headers=auth_headers["admin"])
+    unknown_role = max(RoleEnum.get_values()) + 1
+    unknown_user_response = client.get(f'api/user?r={unknown_role}', headers=auth_headers["admin"])
     assert is_valid(unknown_user_response, 400)
 
     """Non-admin requests return a 401 status code"""


models/user.py
```diff
@@ -17,6 +16,15 @@ class RoleEnum(Enum):
...
     ADMIN = 4
 
+    @classmethod
+    def get_values(cls):
+        return [member.value for name, member in cls.__members__.items()]
+
+    @classmethod
+    def has_role(cls, role):
+        role_values = cls.get_values()
+        return role in role_values
```
NickSchimek, line 26: I'm half tempted to ask for a unit test for this in `tests/unit/test_user.py` - If you have the time to write one up that would be great. otherwise no worries.

models/user.py
```diff
@@ -17,6 +16,15 @@ class RoleEnum(Enum):
     STAFF = 3
     ADMIN = 4
 
+    @classmethod
+    def get_values(cls):
```
NickSchimek, line 20: Should this be designated a private method by using a leading underscore? `_get_values`

NickSchimek (commented):
This is great. A couple comments, but I don't see any reason why this couldn't be merged as is.

models/user.py
aedwardg: Sure, I could do that.

models/user.py
aedwardg: I suppose we could do that, but I just left it as a regular class method because Enum classes seem to work a little differently than normal classes.  

For instance, if you make a normal method with the `self` parameter, it's referring to the Enum member rather than the class instance.  Also, the docs say that names that _start and end_ with a single underscore (e.g., `_ignore_`) are reserved by enum and can't be used.  I don't _think_ changing it to `_get_values` would cause an issue, but I erred on the side of caution and wrote it as a normal class method.  In a normal class I probably would have had a private variable called `_values` and then a getter method that returned that variable.

As a side note, I call `RoleEnum.get_values()` directly in `tests/integration/test_users.py` -- would that be proper for a private method (assuming python had _truly private_ methods)?

NickSchimek: > As a side note, I call RoleEnum.get_values() directly in tests/integration/test_users.py -- would that be proper for a private method (assuming python had truly private methods)?

I noticed that. The easy answer is No, The more complicated answer is 'it depends'.

Let's not make any changes here. It sounds like leaving it as is is the best route forward.

models/user.py
aedwardg: Two quick questions.

1. Do you want me to test for `False` here the same way I did in the integration test (using something like `max(RoleEnum.get_values()) + 1`) or should I hard-code in a value that will likely never get used (e.g., `99`)?

2. When testing for values we expect to return `True`, I figure there are a couple of ways to go about it.  I could loop over each member in `RoleEnum` and assert `RoleEnum.has_role()` on each member's value.  Or I could explicitly create a user with the `empty_test_db` for each of the admin, staff, pm and unauthorized user fixtures and assert `.has_role()` on their role value.  Which would you prefer/think is better?

Or, if you have other suggestions on a better implementation let me know!

NickSchimek: these are good questions, questions that I have answers too, that I could easily spend an hour talking about. 
The fast answer is:
1. neither
2. We only have 3 roles to test so it probably isn't a bad idea, however exhaustive testing usually isn't scalable and can often be unattainable. But in this case that doesn't apply. I'd be happy with one or all three tests. FYI: unauthorized_user should not have a role. 

OK. Let's talk about question 1 a bit. The goal of testing is to test the behavior and not the implementation. Your solutions for question 1 tests implementation. First it uses a semi-private method that returns the Enums values, and assumes that these values will always be integers, and again the hard-coded input makes another assumption, however I would lean toward the hard-coded input for the test; otherwise your test is dependent on a semi-private method that could become obsolete and removed in the future. Perhaps someone decides that the has_role method doesn't need to depend on the get_values method and removes it and updates the has_role method to:

```
    @staticmethod
    def has_role(role):
        return RoleEnum(role)
```

Ideally all the tests would continue to pass when the implementation is changed, and not the behavior. Although implementing it this way technically causes an implementation change since a value error is now raised when a role is invalid. Although in this form the has_role method doesn't really do much for us now, and can probably be considered to remove it as well. With this method gone we don't need the unit tests anymore. Which raises the question: Where should we put tests when we receive input from the client and need to validate it accordingly?

aedwardg: > FYI: unauthorized_user should not have a role.

The role I was referring to was `PENDING`, which is currently assigned by default, absent a specified role.  The pytest fixture that facilitates testing this right now is called `create_unauthorized_user()`...perhaps this name is misleading?

> OK. Let's talk about question 1 a bit.

Thanks for setting me straight on this one.  I'm basically brand new to writing tests like this, but I'm really enjoying it so far, so I look forward to learning as much as I can. 🤓 

So to test the _behavior_ of the `has_role()` method, I suppose I should be able to pass in any argument of any type into `has_role()` and it should return `False` if that role doesn't exist.

However, moving on to your hypothetical where `has_role()` gets removed:

> With this method gone we don't need the unit tests anymore. Which raises the question: Where should we put tests when we receive input from the client and need to validate it accordingly?

I don't really know the answer to this, but I would guess the test should be directly inside the resource.  If that's the case, though, it seems to me you would mostly be trading away a layer of abstraction in exchange for code that is more stable, but potentially harder to read. Would you agree?

NickSchimek: The pending role shouldn't be a role. It is slated to go away. Sorry for the confusion on that. In my thought process, I treat it as it is already gone, but I forget that it still exists. 

> I'm basically brand new to writing tests like this, but I'm really enjoying it so far, so I look forward to learning as much as I can.

Very nice. It is a never ending learning process for all of us.

> I don't really know the answer to this, but I would guess the test should be directly inside the resource. If that's the case, though, it seems to me you would mostly be trading away a layer of abstraction in exchange for code that is more stable, but potentially harder to read. Would you agree?

Good assessment, and it would be harder to read if the code is in the resource. However, that is not where it belongs. Where it belongs depends on the framework/architecture being used. In this app there is a place for it and we're slowly moving everything there. My goal this weekend is to update the documentation so that new and existing contributors know how the application is laid out and where stuff belongs. 

There is no reason not to merge this right now. The code is inline with the old architecture which is prevalent in the app right now. If you would like to refactor this in a pairing session and talk more about these topics then I'd be happy to.

NickSchimek (approved):
Thanks for knocking this out so quickly. Your asking very good questions. Keep it up, and let's talk more about this.

PR #175, opened 2020-12-19
#441

Protopaco:
Allows /manage/properties/:name to update properties.unit.  Just adds single conditional and assignment.  I tried to follow the style of the rest of the file.

In reference to issue #441

### DIFF
--- a/resources/property.py
+++ b/resources/property.py
@@ -107,6 +107,9 @@ class Property(Resource):
         if(data.state):
             rentalProperty.state = data.state
 
+        if (data.unit):
+            rentalProperty.unit = data.unit
+
         if data.propertyManagerIDs:
             rentalProperty.managers = PropertyModel.set_property_managers(data.propertyManagerIDs)
 
--- a/view_coverage.py
+++ b/view_coverage.py
@@ -14,7 +14,7 @@ def stop_coverage(sig, frame):
 signal.signal(signal.SIGINT, stop_coverage)
 
 # the approach below is from: https://2ality.com/2014/06/simple-http-server.html
-PORT = 5000
+PORT = 8001
 Handler = http.server.SimpleHTTPRequestHandler
 with socketserver.TCPServer(("", PORT), Handler) as httpd:
     print(f"Serving coverage details at http://localhost:{PORT}/htmlcov/")
--- a/view_coverage.py
+++ b/view_coverage.py
@@ -14,7 +14,7 @@ def stop_coverage(sig, frame):
 signal.signal(signal.SIGINT, stop_coverage)
 
 # the approach below is from: https://2ality.com/2014/06/simple-http-server.html
-PORT = 8001
+PORT = 5000
 Handler = http.server.SimpleHTTPRequestHandler
 with socketserver.TCPServer(("", PORT), Handler) as httpd:
     print(f"Serving coverage details at http://localhost:{PORT}/htmlcov/")


NickSchimek:
I've probably asked this question before. Do properties have units? If so why?

Also @Protopaco this PR is linked to close a frontend issue. We probably don't want to do that

aedwardg:
> I've probably asked this question before. Do properties have units? If so why?

Even though "unit" is in a few places in the property resource and property model, I think unit (i.e., the unit number) should exist in leases instead.  Looking at the [Figma](https://drive.google.com/file/d/1YqboQogczYm1HkyRqEtVSzeQ61T9hWU2/view), in particular on pages 12, 15, 49 and 50, it appears that properties don't have a field called "unit", but instead have "Units" which appears to represent the total number of units (or maybe available units??) in that property.

That makes a lot more sense, and I don't think the backend reflects that right now.  I think on the backend, the property files should be changed to have "units", "num_units" or "numUnits" instead of "unit" singular.  

Whatever we call it, @Protopaco rightly noticed that the properties resource didn't currently update that field with a PUT request, so 
we could probably merge this after he changes it so that it doesn't close frontend issue 411. Then we could open up a separate issue to change the name of property.unit across the board if there is consensus to do so.

What do you think @NickSchimek?

Also, I noticed the Manage Properties page (Figma pg. 14) doesn't have a column for the number of units.  I wonder if that is intentional or not? Maybe one of the frontend people can chime in on that  🙂

NickSchimek:
That makes sense @aedwardg . Do you have any insight @htharker42 @Chris-Boe @DaveTrost ?

Protopaco:
I apologize for the confusion.  @aedwardg answered it exactly.  Currently the properties have a value of "unit" which reflects the number of units in the property.  The PUT updates all of the values except for the .unit value.  Which is why I made this pull request.

That being said, I think there could be better ways to reflect the number of units in the property.  Ideally I would think an array of unit numbers that can be assigned to individual tenants, as well as array.length could reflect the total number of units.  That is more of a big vision question however. 

For this application I'm happy to work with whatever data structures make the most sense.  Thanks!

NickSchimek:
Thanks! That answers my question. Let's get this merged. However, we should get that attribute named correctly.  I would suggest `num_units`.

@Protopaco 
> I think there could be better ways to reflect the number of units in the property. Ideally I would think an array of unit numbers that can be assigned to individual tenants

Generally, it is not ideal to use an array data structure as a column type on a table for a relational DB. However, if the client wants to store each unit number on the DB for every property, then there are ways to capture this, such as having a Unit Model with an associated Unit table in the DB.

@pogorobot mentioned the use of a Unit Model so that certain questions could be asked to the DB. Storing the data that way makes perfect sense if we need to ask those questions. There are a lot of different ways to store the data depending on the type of questions you want to ask. As of right now, the requirements don't require a Unit Table, and I'm not about to guess what the client wants because I know I will be wrong. So let's get this column name changed, and I think we will be good to go. If someone wants to create an issue for that and reference this ticket then that would be ~~just~~ super awesome.

resources/property.py
```diff
@@ -107,6 +107,9 @@ def put(self, name):
         if(data.state):
             rentalProperty.state = data.state
 
+        if (data.unit):
+            rentalProperty.unit = data.unit
```
NickSchimek, line 111: hopefully we'll get this refactored soon so that issues like this don't arise.

NickSchimek (approved):
Looks good. Thank you.

PR #176, opened 2020-12-19
Update fixtures to use Faker

### LINKED ISSUE #459: Update fixtures to use Faker
**Description**
Update fixtures to use Faker - Faker gives us dynamic values which has a lot of benefits, furthermore it allows to create unique values for our fixtures. For example: If we create multiple users for a test then each user can automatically have a unique email.

pogorobot:

Closes codeforpdx/dwellingly-app/issues/459

I did remove one small test that depended on the specific hard-coded value it had previously. Nothing else should be affected.

### DIFF
--- a/tests/factory_fixtures/contact_number.py
+++ b/tests/factory_fixtures/contact_number.py
@@ -1,15 +1,13 @@
 import pytest
-from faker import Faker
 from models.contact_number import ContactNumberModel
 
 @pytest.fixture
-def contact_number_attributes():
+def contact_number_attributes(faker):
     def _contact_number_attributes():
-        fake = Faker()
         return {
-            "number": fake.phone_number(),
-            "numtype": "mobile",
-            "extension": "1"
+            "number": faker.phone_number(),
+            "numtype": faker.random_element(("home", "work", "mobile")),
+            "extension": faker.bothify(text="?###"),
         }
     yield _contact_number_attributes()
 
--- a/tests/factory_fixtures/contact_number.py
+++ b/tests/factory_fixtures/contact_number.py
@@ -1,11 +1,13 @@
 import pytest
+from faker import Faker
 from models.contact_number import ContactNumberModel
 
 @pytest.fixture
 def contact_number_attributes():
     def _contact_number_attributes():
+        fake = Faker()
         return {
-            "number": "555-555-5555",
+            "number": fake.phone_number(),
             "numtype": "mobile",
             "extension": "1"
         }
--- a/tests/factory_fixtures/lease.py
+++ b/tests/factory_fixtures/lease.py
@@ -1,25 +1,23 @@
 import pytest
-from faker import Faker
 from models.lease import LeaseModel
 from datetime import datetime
 
 @pytest.fixture
-def lease_attributes():
+def lease_attributes(faker):
     def _lease_attributes(unitNum, tenant, property):
-        fake = Faker()
         return {
             "unitNum": unitNum,
             "tenantID": tenant.id,
             "propertyID": property.id,
-            "dateTimeStart": fake.date_time_this_decade(),
-            "dateTimeEnd": fake.date_time_this_decade(False, True), #the future
-            "occupants": 3
+            "dateTimeStart": faker.date_time_this_decade(),
+            "dateTimeEnd": faker.date_time_this_decade(before_now=False, after_now=True),
+            "occupants": faker.random_number(digits=2)
         }
     yield _lease_attributes
 
 @pytest.fixture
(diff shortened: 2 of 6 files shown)


aedwardg:
This looks good to me.  

I'm not sure if @NickSchimek will want these to be in the same format as the fixtures for [property](https://github.com/codeforpdx/dwellinglybackend/blob/development/tests/factory_fixtures/property.py) and [user](https://github.com/codeforpdx/dwellinglybackend/blob/development/tests/factory_fixtures/user.py).  Those used Faker's [built-in pytest fixture](https://faker.readthedocs.io/en/master/pytest-fixtures.html) instead of explicitly importing `faker` and creating an instance of the `Faker` class in each file, but the end result looks to be the same.

tests/factory_fixtures/contact_number.py
```diff
@@ -1,11 +1,13 @@
...
 
 @pytest.fixture
 def contact_number_attributes():
     def _contact_number_attributes():
+        fake = Faker()
         return {
-            "number": "555-555-5555",
+            "number": fake.phone_number(),
             "numtype": "mobile",
             "extension": "1"
```
NickSchimek: We should generate some dynamic values using faker for numtype and extension too.

tests/factory_fixtures/lease.py
```diff
@@ -1,16 +1,18 @@
...
+        fake = Faker()
         return {
             "unitNum": unitNum,
             "tenantID": tenant.id,
             "propertyID": property.id,
-            "dateTimeStart": datetime.utcnow(),
-            "dateTimeEnd": datetime.utcnow(),
+            "dateTimeStart": fake.date_time_this_decade(),
+            "dateTimeEnd": fake.date_time_this_decade(False, True), #the future
             "occupants": 3
```
NickSchimek: Very nice I like this. The comment is great, however passing false and true arguments without knowing what they're for is confusing. Perhaps it could be changed to do this? `fake.date_time_this_decade(before_now=False, after_now=True)` - Then the comment is not needed.

Will also want to use faker for the occupants field, and the unitNum field on line 22.

tests/factory_fixtures/tenant.py
```diff
@@ -1,13 +1,15 @@
...
 @pytest.fixture
 def create_tenant(create_property):
     def _create_tenant(property=create_property()):
+        fake = Faker()
         tenant = TenantModel(
-                firstName="firstName",
-                lastName="lastName",
-                phone="1234567890",
+                firstName=fake.unique.first_name(),
+                lastName=fake.unique.last_name(),
```
NickSchimek: There are not unique constraints for the firstName and lastName on the tenant model. People are allowed to have the same name. so I don't think using unique here is necessary.

Also, the way Faker is being used here will not guarantee unique data. Pytest has built-in support for Faker and it is available as factory. We should be using the faker factory so that when we do call unique we can be sure we will get unique values. Plus, the faker factory should run faster than having to instantiate a Faker object each time we call a factory.

tests/factory_fixtures/ticket.py
```diff
@@ -17,7 +18,8 @@ def _ticket_attributes(issue, tenant, assignedUser, sender):
 
 @pytest.fixture
 def create_ticket(ticket_attributes, create_tenant, create_admin_user, create_join_staff):
-    def _create_ticket(issue="Leaky pipe"):
+    fake = Faker()
+    def _create_ticket(issue=fake.sentence()):
```
NickSchimek: Line 15 needs to be updated too.

NickSchimek (changes requested):
Thanks @pogorobot ! Looking good. There is a few places that were missed that are highlighted below. Additionally the Emergency Contact file needs to be updated too.

Please see my comment regarding the use of instantiating a faker object versus the use of the built-in factory object.

tests/factory_fixtures/contact_number.py
```diff
@@ -1,15 +1,13 @@
...
+def contact_number_attributes(faker):
     def _contact_number_attributes():
-        fake = Faker()
         return {
-            "number": fake.phone_number(),
-            "numtype": "mobile",
-            "extension": "1"
+            "number": faker.phone_number(),
+            "numtype": faker.random_element(("home", "work", "mobile")),
+            "extension": faker.bothify(text="?###"),
```
NickSchimek, line 10: Very cool!

NickSchimek (approved):
Thanks @pogorobot ! Nice work 💫  Looks like the emergency_contact factory still needs to be updated, however we can get that in a different issue. Let's get this merged. Thank you!

PR #177, opened 2020-12-20
Speeding up a slow test

NickSchimek:
There is no issue for this one. Just noticed a slow test that could be sped up. I also set archived to default to False on the UserModel.

### DIFF
--- a/tests/integration/test_user_invite.py
+++ b/tests/integration/test_user_invite.py
@@ -1,6 +1,5 @@
 import pytest
 from unittest.mock import patch
-from models.user import RoleEnum
 from resources.email import Email
 
 
--- a/tests/integration/test_user_invite.py
+++ b/tests/integration/test_user_invite.py
@@ -1,0 +1,18 @@
+import pytest
+from unittest.mock import patch
+from models.user import RoleEnum
+from resources.email import Email
+
+
+@pytest.mark.usefixtures('client_class', 'empty_test_db')
+class TestUserInvite:
+    def setup(self):
+        self.endpoint = '/api/user/invite'
+
+    @patch.object(Email, 'send_user_invite_msg')
+    def test_invite_user(self, send_user_invite_msg, valid_header, user_attributes):
+        response = self.client.post(self.endpoint, headers=valid_header, json=user_attributes(role='STAFF'))
+
+        send_user_invite_msg.assert_called()
+        assert response.status_code == 200
+        assert response.json == {'message': 'User Invited'}
--- a/models/user.py
+++ b/models/user.py
@@ -39,7 +39,7 @@ class UserModel(BaseModel):
     phone = db.Column(db.String(20), nullable=False)
     hash_digest = db.Column(db.LargeBinary(60))
     password = db.Column(db.String(), default=None, onupdate=None)
-    archived = db.Column(db.Boolean)
+    archived = db.Column(db.Boolean, default=False, nullable=False)
     lastActive = db.Column(db.DateTime, default=datetime.utcnow)
 
     notes = db.relationship(models.notes.NotesModel,
--- a/resources/user_invite.py
+++ b/resources/user_invite.py
@@ -1,7 +1,7 @@
 from flask_restful import Resource
 from flask import request
 from resources.admin_required import admin_required
-from models.user import UserModel, RoleEnum
+from models.user import UserModel
 from schemas import UserSchema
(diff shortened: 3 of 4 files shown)


tests/integration/test_user_invite.py
```diff
@@ -0,0 +1,18 @@
+import pytest
+from unittest.mock import patch
+from models.user import RoleEnum
```
aedwardg: `RoleEnum` isn't used here, so we can delete this import.

aedwardg (commented):
I'm okay with this for now, because we know the `send_user_invite_msg()` static method works as it's implemented right now. 

That said, we still want to transition away from flask-mail, so we'll have to update that method in the near future.  When we do, will we still keep this test mocked? And if so, how will we test that the invite email sends with the new mail extension? Manual testing?

Other than those questions, this looks good.  The only change I'd suggest is getting rid of the unused import I mentioned in my other comment.

NickSchimek:
> I'm okay with this for now, because we know the `send_user_invite_msg()` static method works as it's implemented right now.
> 
> That said, we still want to transition away from flask-mail, so we'll have to update that method in the near future. When we do, will we still keep this test mocked? And if so, how will we test that the invite email sends with the new mail extension? Manual testing?
> 
> Other than those questions, this looks good. The only change I'd suggest is getting rid of the unused import I mentioned in my other comment.

It sounds like you're a bit concerned with mocking this method. If so, let me try to alleviate some of that concern. Previously we could've deleted this statement `Email.send_user_invite_msg(user)` in the `resources/user.py` file without any test failing. However, now this test will fail if the statement is deleted or changed. This is a better test than the one we had before. I mentioned this change was done to speed up a test, but it also asserts behavior that we expect in our resource that we were not checking before. And as you said the `send_user_invite_msg` is already tested and confirmed to work elsewhere so we can trust that it will do its job.

This test is better than what we had, and it could be slightly improved by asserting the correct arguments are passed in, and stating that it is only called once.
`assert_method_called_once_with(user)`

Part 2 of your question is easy to answer. It depends. 

Ultimately it depends on what kind of testing you want to do and the tools at your disposal. If you want to do a complete integration/system test and the tools that you are working with allow you to do that then we would probably not mock the email send message here. If the mail library we're using is environment aware, (it knows when it is being executed in dev/test/prod mode) then it will act accordingly depending on the environment it is being executed in. Emails should not be sent in dev and test environments; instead we can assert that the message is queued for delivery, hence we can assert that the message is queued/sent successfully without actually sending a message. Otherwise, you may have to mock, and the mocking tools that the unit test and pytest libraries provide do check that the methods that are mocked do exist on the receiver. So if the mocked objects public api ever changes then the test will fail.

I usually don't do any manual testing in the development environment. Once this app is pushed out to a production server then it can be manually tested there, to your hearts content, before releasing to prod. However, yes you could manually test to ensure you have implemented the new library correctly, although you'd probably only do it once and then never again. And deciding to manually test in the first place would probably depend on the developer's confidence level working with the library.

tests/integration/test_user_invite.py
NickSchimek: Thanks, I will remove this.

tests/integration/test_user_invite.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+@pytest.mark.usefixtures('client_class', 'empty_test_db')
+class TestUserInvite:
+    def setup(self):
+        self.endpoint = '/api/user/invite'
+
+    @patch.object(Email, 'send_user_invite_msg')
+    def test_invite_user(self, send_user_invite_msg, valid_header, user_attributes):
+        response = self.client.post(self.endpoint, headers=valid_header, json=user_attributes(role='STAFF'))
```
NickSchimek, line 13: I created issue codeforpdx/dwellingly-app/issues/491 to address the issue of passing a string as a value rather than the agreed upon ints. This is the first resource to use the User schema and the schema is requiring string values instead of ints.

aedwardg:
> It sounds like you're a bit concerned with mocking this method. If so, let me try to alleviate some of that concern. 

Thanks for writing up such a great explanation!  I was mostly just concerned with mocking up a method that we may have to un-mock in a week or two.  However, now that I see the mock's `assert_called` method makes the tests more robust in addition to speeding them up, I agree that this is worthwhile to add in regardless.

aedwardg (approved):
Looks good to me!

PR #178, opened 2020-12-21
Update validation to use simplified iso format

NickSchimek:
No issue that I know of. This issue was reported by @DaveTrost on slack.

The lease was expecting time formats for the lease start and end date in the standard specified by Pythons iso8601 format although the javascript frontend is sending a simplified extended iso 8601 format. We had agreed upon the simplified extended  iso format so the backend is being changed to support that.


aedwardg (approved):
Looks good, passes all tests, resolves the error 👍 

Thanks for getting this done so quickly! 💪

PR #179, opened 2020-12-22
Create authorization decorators

NickSchimek:
## Description

Created authorization decorators for each role (admin, staff, pm). This is done because the app in its current state does not behave differently if the user is a join staff member or a property manager. And we require this behavior to distinguish between the two; otherwise why would we have the roles. 

The `@jwt_required` decorator should be considered insecure and has been removed across the app and replaced with a somewhat equivalent `pm_level` decorator. The `pm_level` decorator is slightly stronger than the `jwt_required` decorator because it checks the role of the user where as the `jwt_required` decorator did not. 

One test had to be updated because it was using a user with the role `pending` to make patch requests on the user endpoint. It's been changed to use a property manager as the requestor, and pending roles no longer have access with the `pm_level` decorator.

I did not create unit tests for these auth decorators. If someone would like to that would be great because I'm not quite sure how to right now, and I'm not sure if I will have the time to figure out how. For now we can rely on the integration tests to check them. 

The new auth decorators are:

`staff_level_required`
`pm_level_required `

along with the already existing `admin_required` required make up the three different authorizations we can have. 

There are also three convenient helper methods to easily determine if the `current_user` is of a certain role:
`admin()`, `staff()`, `pm()` - They return `True` if they are and `False` if they are not.


tests/integration/test_users.py
```diff
@@ -174,7 +174,7 @@ def test_patch_user(client, auth_headers, new_user, create_admin_user):
 
     newEmail = "unauthorizedpatch@test.com"
 
-    unauthorizedResponse = client.patch(f"/api/user/{userToPatch.id}", json={"email": newEmail}, headers=auth_headers["pm"])
+    unauthorizedResponse = client.patch(f"/api/user/{userToPatch.id}", json={"email": newEmail}, headers=staff_header)
```
aedwardg: Should this be `pm_header` instead of `staff_header` in lines 129 & 177?

aedwardg (commented):
This looks really great @NickSchimek, thanks for working on it!  This PR solves some of our previous auth issues and cleans things up quite a bit.  I just have the one question below regarding your changes in `test_users.py`.

On a broader scope, however, should we be entirely replacing `auth_headers` with `admin_header`, `staff_header` and `pm_header` in all our tests?  You mentioned this back in September in PR #113.  If so, we should probably add an issue for this to prevent more tests being made with `auth_headers`.

This would be in addition to switching from `jwt_required` to the new decorators mentioned in c8377ad

tests/integration/test_users.py
NickSchimek: Good catch. It shouldn't matter since it is ultimately testing wether a non-admin is not allowed, but to be consistent with the previous test it should be. I changed this because the jwt's being used are unrealistic of actual jwts and causing an issue. Let's fix this.

NickSchimek:
The way this has been created. We probably wont need the `valid_header` anymore since it will just be a copy of the `admin_header` - Admins should have access to the whole site so... I think perhaps it should be kept around just because the name implies that this test is not checking authorizations, but that's just my personal preference. Will push up one more commit to update the valid_header jwt.

NickSchimek:
> On a broader scope, however, should we be entirely replacing auth_headers with admin_header, staff_header and pm_header in all our tests?

Yes, all the tests need to be refactored still. The auth_headers take a bit longer to run because they hit the login endpoint each time they're executed. When the auth_header is used in a test then it is also testing the login endpoint, which does not need to be done, because the login endpoint should already be under test.

aedwardg (commented):
Looks good to me.

I like the refactor using `_user_claims()` and I have no problem with keeping `valid_header` around for tests where we don't want to give the impression that we are testing auth.

I'm ready to approve unless you want to keep this open for a while to see if anyone else wants to chime in.

NickSchimek:
Feel free to approve. I’m not going to merge this for a few more days in
order to give others a chance to review.

aedwardg (approved):
LGTM 👍

PR #181, opened 2020-12-23
Remove set_property_managers method

### LINKED ISSUE #489: Properties/:id won't remove all property managers
**Describe the bug**
When the frontend sends a .put request to /properties/:id with data.propertyManagerIDs as an empty array, indicating that all property managers have been removed, the backend does not remove the remaining property managers.  The frontend receives a response that includes the final property manager.

On line 135 of resources/property.py, the following code deals with this instance:

```
        if data.propertyManagerIDs:
            rentalProperty.managers = PropertyModel.set_property_managers(data.propertyManagerIDs)

```
I believe that if we change the line to:

```
        if data.propertyManagerIDs != []:
            rentalProperty.managers = PropertyModel.set_property_managers(data.propertyManagerIDs)

```
it will recognize the empty array and work the way it is intended.  

If this is an expected response, i.e. the user shouldn't be able to remove all of a property's property manager, than we can add notices on the front end side.

**To Reproduce**
1. Go to '/manage/properties/2'
2. Click on 'edit button'
3. Click on 'remove property manager icon (top right of property manager box)'

**Expected behavior**
User should be able to remove all property managers.

 - OS: Catalina 10.15.7
 - Browser: Chrome
 - Version: 87
 

**Additional context**
Please let me know if this is the correct way to be doing this.  I want to work within the established workflows.  

Thank you!
Paul

NickSchimek:

This uses the Schema to validate and deserialize request data for the POST and PUT resource on the Property so that the set_property_managers method and the init method could be removed.

Partially resolves issue codeforpdx/dwellingly-app/issues/403
Fixes codeforpdx/dwellingly-app/issues/489

Is a re-seeding of the database necessary? Rebuilding of the DB is necessary


data/seedData.py
```diff
@@ -11,6 +11,8 @@
 from models.contact_number import ContactNumberModel
 from models.lease import LeaseModel
 from utils.time import time_format
+from schemas import *
```
aedwardg: The only schema used here is `PropertySchema`, so it's probably best to only import that, rather than using a star-import.

models/property.py
```diff
@@ -1,10 +1,9 @@
 from db import db
 from utils.time import Time
-from marshmallow import ValidationError
 
 from models.tenant import TenantModel
 from models.base_model import BaseModel
-from models.user import UserModel, RoleEnum
+from models.user import UserModel
 from models.property_assignment import PropertyAssignment
```
aedwardg, line 7: `PropertyAssignment` isn't used. Let's get rid of this import.

resources/property.py
```diff
@@ -1,8 +1,12 @@
 import json
 from flask_restful import Resource, reqparse
+from flask import request
 from utils.authorizations import admin_required
 from db import db
 from models.property import PropertyModel
+from schemas.property_assignment import PropertyAssignSchema
+from models.user import UserModel, RoleEnum
```
aedwardg: Unused imports:
- `json`
- `PropertyAssignSchema`
- `UserModel`, `RoleEnum`

schemas/property.py
```diff
@@ -13,6 +15,20 @@ class Meta:
...
+        PropertyModel.validate(PropertyAssignSchema, payload, partial=True, many=True)
+
+    @validates("name")
+    def validates_uniqueness_of_name(self, value):
+        if PropertyModel.find_by_name(value):
+            raise ValidationError('A property with this name already exists')
+
+    @post_load
+    def make_property_attributes(self, data, **kwargs):
+        data['managers'] = [ UserModel.find(manager) for manager in data['propertyManagerIDs'] ]
```
aedwardg: There's a lot of horizontal whitespace in the list comprehensions (lines 22 & 32).  We can either get rid of that now, or wait until pre-commit hook is in place and `Black` will do it for us.

Also, on line 22, `range(0, len(value))` doesn't technically need the 0, it could just be `range(len(value))`.  
I'm okay with leaving it in though if you think it improves readability.

tests/factory_fixtures/property.py
```diff
@@ -1,29 +1,25 @@
...
-def property_attributes(faker):
-    def _property_attributes(archived=False, pm=None):
+def property_attributes(faker, create_property_manager):
+    def _property_attributes(archived=False, manager_ids=[create_property_manager().id]):
         return {
             'name': faker.unique.name(),
             'address': faker.address(),
             'city': faker.city(),
             'unit': faker.building_number(),
             'state': faker.state() ,
```
aedwardg: Let's get rid of the whitespace before the comma here.  
It's from a previous PR, but we might as well get rid of it.

tests/schemas/test_property_schema.py
```diff
@@ -6,9 +6,10 @@
 from models.property import PropertyModel
```
aedwardg, line 4: Unused imports:
- `UserSchema`
- `UserModel`
- `PropertyModel`

tests/unit/test_property.py
```diff
@@ -1,26 +1,38 @@
...
-    assert actual == expected
+        assert prop
+        assert prop.managers == [pm_1, pm_2]
 
+    def test_manager_update(self, create_property, create_property_manager):
+        prop = create_property()
+        pm_1 = prop.managers[0]
+        pm_2 = create_property_manager()
+        pm_3 = create_property_manager()
+        payload = { 'propertyManagerIDs': [pm_2.id, pm_3.id] }
```
aedwardg: This variable is never used.  Do we need it?

aedwardg (commented):
Nice job.  This covers a significant amount of  the refactoring we needed to do on the Properties resource.  I don't see any problems, I just have a couple of formatting/clean-up suggestions in the comments.

schemas/property.py
NickSchimek: Nope. this is better `range(len(value))`

models/property.py
NickSchimek: PropertyAssignment is used on line 24 when defining the relationship. I'm not sure, but maybe we could use the class name on line 24 instead of the table name; otherwise you'll just have to ignore what your editor is telling you.

NickSchimek: I tried putting in the class name, but no go. I remember looking in the docs to see if we had to use a the table name string representation, when reviewing the PR for that, however I was unsuccessful in finding another way to define the many-to-many relationship.

tests/unit/test_property.py
NickSchimek: I removed `pm_1` var

models/property.py
aedwardg: Oh interesting, I didn't catch that.  That seems really strange...I feel like there _should_ be a better way to do it, but there may not be.  I'll make a note for myself to look into it, but for now I guess that's how it has to be 🤷‍♂️

aedwardg: Passing in `PropertyAssignment.__tablename__` instead of the string works.  It's kind of ugly though.  Maybe would could assign it to a variable or something to make it look nicer and go that route?

NickSchimek:
See, I make mistakes too. Pushed up a change to only deserialize propertManagerIDs if the data is present. However, I have a feeling the frontend will always send the propertyManagerIDs data regardless if an update is needed. If this is the case, then it highlights my main issue with doing things this way. The propertyManager table will always be updated regardless. However, this is fine for now, but something to probably take note of. I just don't like recreating rows in the db if there is no change. Perhaps this is common practice? I don't know.

models/property.py
aedwardg: @NickSchimek, I was thinking about this 👆 over the holiday.

What do you think about adding the following code to the `BaseModel`?

```python
@classmethod
def get_tablename(cls):
    return cls.__tablename__
```

I would prefer this over over importing the model and then only using the string representation of the table name.  I think it's less confusing to see `secondary=PropertyAssignment.get_tablename()` or something similar because it exposes what's taking place and Flake8 won't flag the unused import.

aedwardg:
> I have a feeling the frontend will always send the propertyManagerIDs data regardless if an update is needed. If this is the case, then it highlights my main issue with doing things this way. The propertyManager table will always be updated regardless. However, this is fine for now, but something to probably take note of. I just don't like recreating rows in the db if there is no change. 

I agree.  We'll have to keep an eye on what the frontend ends up sending and either adjust or have them adjust as needed.  Ideally, we'd want `propertyManagerIDs` to be excluded from the payload if it's not being updated.

aedwardg (commented):
The new changes look good.  I'm ready to approve as soon as I hear your feedback on the tablename issue I commented on.

NickSchimek:
> > I have a feeling the frontend will always send the propertyManagerIDs data regardless if an update is needed. If this is the case, then it highlights my main issue with doing things this way. The propertyManager table will always be updated regardless. However, this is fine for now, but something to probably take note of. I just don't like recreating rows in the db if there is no change.
> 
> I agree. We'll have to keep an eye on what the frontend ends up sending and either adjust or have them adjust as needed. Ideally, we'd want `propertyManagerIDs` to be excluded from the payload if it's not being updated.

Maybe the ORM doesn't recreate if not necessary. I honestly don't know. I think there was a way to show what SQL commands the ORM sends when playing around in the shell. I thought having the environment sent to development would do the trick, but apparently not. I think it is something worthwhile to look into.

models/property.py
NickSchimek: Very nice! This is a much better solution. I don't think I'd have any issue doing `PropertyAssignment.__tablename__` although we can also put it in the BaseModel as well, although I'd prefer to do `tablename` as the method name. `PropertyAssignment.tablename`.

aedwardg (approved):
LGTM

PR #183, opened 2020-12-24
Remove init method on tenant model

NickSchimek:
Partially resolves issue codeforpdx/dwellingly-app/issues/398

Sorry, did this in one commit. I was streaming part of this work and I have a hard time talking, thinking, and typing at the same time. It is pretty similar to PR #181 

Removed validation for tenants with the same first and last name. If this is needed we can add it to the schema, although there are many people with the same first and last name so I would advise against this unless it is specifically requested from the client.

Increased length validation on phone_number as the faker library can and will produce phone_numbers greater than 20 digits. For example `+1-604-876-4759x38242` - and some international phone numbers can have even more digits. It's probably best to be lenient with our validations unless specifically required, especially since multi-tenancy can be a thing in the future.

Is a re-seeding of the database necessary? Yes, i think db needs to be re-created

### DIFF
--- a/resources/tenants.py
+++ b/resources/tenants.py
@@ -7,6 +7,7 @@ from models.tenant import TenantModel
 from models.user import UserModel
 from models.lease import LeaseModel
 from schemas.lease import LeaseSchema
+from schemas.tenant import TenantSchema
 from datetime import datetime
 from utils.time import Time
 from schemas.lease import LeaseSchema
@@ -21,12 +22,6 @@ from datetime import datetime
 # | DELETE | `v1/tenants/:id`     | Deletes a single tenant   |
 
 class Tenants(Resource):
-    parser = reqparse.RequestParser()
-    parser.add_argument('firstName',type=str,required=True,help="This field cannot be blank")
-    parser.add_argument('lastName',type=str,required=True,help="This field cannot be blank")
-    parser.add_argument('phone',type=str,required=True,help="This field cannot be blank")
-    parser.add_argument('staffIDs',action='append',required=False,help="This field can be provided at a later time")
-
     @admin_required
     def get(self, tenant_id=None):
         # GET /tenants
@@ -42,12 +37,7 @@ class Tenants(Resource):
 
     @admin_required
     def post(self):
-        data = Tenants.parser.parse_args()
-        if TenantModel.find_by_first_and_last(data["firstName"], data["lastName"]):
-            return { 'message': 'A tenant with this first and last name already exists'}, 401
-
-        tenantEntry = TenantModel(**data)
-        TenantModel.save_to_db(tenantEntry)
+        tenantEntry = TenantModel.create(schema=TenantSchema, payload=request.json)
 
         returnData = tenantEntry.json()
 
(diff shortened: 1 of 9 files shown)


schemas/tenant.py
```diff
@@ -25,6 +29,17 @@ class Meta:
 
     phone = fields.Str(
         required=True,
-        validate=validate.Length(min=10, max=20),
+        validate=validate.Length(min=10, max=30),
```
aedwardg, line 32: I noticed last week that the front end is allowing a much shorter minimum phone length: 5 or 7 characters, depending on the page.  I believe they check for a maximum of 20 characters as well, but I haven't looked at every frontend page. 

We should put it on the roadmap to come to a consensus on this.  I'm fine with changing the max and min values, so long as it's uniform across the app.

tests/schemas/test_tenant_schema.py
```diff
@@ -54,7 +56,7 @@ def test_phone_min_10(self):
         assert 'phone' in validation_errors
 
     def test_phone_max_20(self):
-        long_phone_number = '8' * 21
+        long_phone_number = '8' * 41
         validation_errors = TenantSchema().validate({'phone': long_phone_number})
```
aedwardg: Should we maybe change the name of this test now that it's not testing for a maximum of 20 chars?
Also, what's the reasoning behind using 41 here as opposed to 31 (since the schema allows for a max of 30 chars now)?

resources/tenants.py
```diff
@@ -7,6 +7,7 @@
 from models.user import UserModel
 from models.lease import LeaseModel
 from schemas.lease import LeaseSchema
+from schemas.tenant import TenantSchema
 from datetime import datetime
 from utils.time import Time
 from schemas.lease import LeaseSchema
```
aedwardg: There are a bunch of unused imports in this file.  Not sure if you want to clean them up on this go-around:
- `json`
- `reqparse`
- `db`
- `UserModel`
- `Time`
- `datetime` (imported twice, never used)
- `LeaseSchema` (this one is used, but it's imported twice)

aedwardg (commented):
Looks good to me.  Just a couple of questions/comments.

schemas/tenant.py
NickSchimek: Agreed, although I would be careful to not implement validations that are restrictive without a good reason. If the client suggested it then that would be a good reason. If the client hasn't then we probably don't even need validations. This app will be used by a business so we probably don't have to worry about users intentionally putting in bad data.

tests/schemas/test_tenant_schema.py
NickSchimek: > Should we maybe change the name of this test now that it's not testing for a maximum of 20 chars?

Gosh, I don't think I read the method name. Preferably method names should not be so specific. I will udpate.

> Also, what's the reasoning behind using 41 here as opposed to 31

I don't remember. This tests gives me the confidence that the validation is in place so I don't think it matters if it is 41 or 31. Perhaps we should do an exhaustive test? However, that is impossible to do the infinite number of integers. Regardless if this test is set to 31 or 41 ... I could still make the test pass while allowing very long phone numbers to be entered into the db. So looking at the implementation of the validation and its test, I have the confidence that the validation is working and will not allow phone number lengths greater than 30 to be passed in.

aedwardg (dismissed):
LGTM.  We are slowly sniping off those `__init__` methods!

PR #185, opened 2020-12-30
Python style enforced with pre-commit hook

### LINKED ISSUE #348: Add python style check for backend development
**Purpose**
Add style check to enforce consistent style for backend code. This will help code readability. This should not affect actual functionality or testing in backend.

**Proposed Solution**
Hugh mentioned he tried flake8, and a few other style check that didn't play nice with flask. We should document the list to find the right solution.

**Acceptance Criteria**
- [ ] enforce style when running test. Fail the test for code not within the style guideline.

N/A

**Additional context**

aedwardg:

Closes codeforpdx/dwellingly-app/issues/348

This PR adds Flake8-bugbear, Black and Pre-commit dependencies in order to enforce clean and uniform python style throughout the backend.  

Once installed, the pre-commit hook will automatically perform the following actions when you try to make a commit:
- Add a newline to the end of the file if it doesn't have one
- Strip trailing whitespace on every line
- Format all Python code in Black style
- Lint with Flake8 and Flake8-bugbear for non-formatting errors

If the hook re-formatted any of your code or if you had to fix something to resolve a Flake8 error, you will need to add and commit those files again in git because they have been further modified.

A few files had some specific code changed in order to comply with the new, stricter linting, but most of the changes are formatting.

I tried to keep all important changes in their own commits, with the most recent commit reserved for the non-critical formatting changes.

Is a re-seeding of the database necessary? I don't think so.
Any new dependencies to install? YES
Any special requirements to test? YES

**New instructions for dependency install**
Because Black still hasn't released an official version 1.0.0, you have to allow for pre-release packages when installing with pipenv:
`pipenv install -d --pre`

Also, once you install the dependencies, you will need to install the pre-commit hook for it to work (only once, though we might need to do this again if the pre-commit config gets changed in the future):
`pipenv run pre-commit install`


aedwardg:
You might notice that in `conftest.py` and `app.py` there are some comments that say: `noqa:` followed by some error codes.
This is how to make Flake8 ignore a particular error on that line.  If we figure out a way to resolve those two lines, we can remove the comments.

Also, the Flake8 pre-commit is currently configured to ignore `__init__` files, since we are importing modules that are not being used directly in `__init__` and also using star imports to import all factory fixtures.  If we need to ignore additional files in the future (such as auto-generated Alembic files), we'll need to update the pre-commit-config.yaml with the appropriate regex patterns.

aedwardg:
Also, documentation should probably be made on how to set up Flake8 and Black for use on editors and the command line (e.g., format on save and auto-linting).  When used properly in your workflow, you won't have to change much at commit time because most of it will already pass the pre-commit checks.

This could maybe be a separate issue?

tests/integration/test_leases.py
```diff
@@ -33,188 +30,211 @@ def test_get_all_leases(self, valid_header, create_lease):
...
     def test_admin_is_authorized_to_get(self, admin_header, create_lease):
         lease = create_lease()
 
-        response = self.client.get(f'/api/lease/{lease.id}', headers=admin_header)
+        response = self.client.get(f"/api/lease/{lease.id}", headers=admin_header)
         assert response.status_code == 200
 
     def test_pm_is_authorized_to_get_all(self, pm_header, create_lease):
-        lease = create_lease()
+        create_lease()
```
NickSchimek: yeah, we don't need this line. These are testing auths, and it shouldn't matter if there is data or not.

tests/factory_fixtures/contact_number.py
```diff
@@ -9,15 +10,19 @@ def _contact_number_attributes():
...
+
     yield _contact_number_attributes()
 
+
 @pytest.fixture
 def create_contact_number(contact_number_attributes, create_emergency_contact):
-    def _create_contact_number(emergency_contact=create_emergency_contact()):
+    emergency_contact = create_emergency_contact()
+
+    def _create_contact_number(emergency_contact=emergency_contact):
```
NickSchimek: this keeps the current behavior intact. However, since these are fixtures, it would be better if each call to the fixture is completely dynamic. 
Perhaps we can do something like this instead: It is ugly, but I believe this is the only way to do, what I intended to do, in Python.
```
    def _create_contact_number(emergency_contact=None):
        if not emergency_contact:
            emergency_contact = create_emergency_contact()
```

CONTRIBUTING.md
```diff
@@ -76,22 +76,24 @@ NOTE: Database is SQLite3 via SQLAlchemy
 2. Install [pipenv](https://pipenv.pypa.io/en/latest/#install-pipenv-today)
     - Note: It is not necessary to install python before installing pipenv.
     - Please install pipenv according to their docs for your OS.
-3. Install dependencies `pipenv install -d`
+3. Install dependencies `pipenv install -d --pre`
```
NickSchimek: This is getting to be annoying. I wish pipenv would install everything by default

CONTRIBUTING.md
```diff
@@ -160,6 +162,12 @@ How to contribute to this project.
 
 (Step #3 creates a new branch titled <name of branch> and navigates you to that branch)
 
+4. Commit your changes (and resolve commit errors)
+    - Our pre-commit hook is designed to be strict in order to ensure all code being committed to the codebase is as clean and uniform as possible. Commit errors are a good thing, and completely expected!
+    - The pre-commit hook will check your code every time you try to commit. If your code needs re-formatting, that will happen automatically. If you have non-formatting errors in your code, these will be printed to the terminal with the error, filename and line number. You will need to resolve these yourself.
+    - After your code has been reformatted and/or you have resolved other errors, you will need to add and commit those files again (because they have been modified).
+    - Successful commits can then be pushed to your remote branch like normal.
+
```
NickSchimek, line 170: Looks good! Thank you for updating the docs!

resources/tenants.py
```diff
@@ -14,49 +14,57 @@
...
-        if ("dateTimeEnd" in leaseData and "dateTimeStart" in leaseData and "propertyID" in leaseData):
-            LeaseModel.create(
-                schema=LeaseSchema,
-                payload=leaseData
+        # if this tenant has a lease
+        if (
+            "dateTimeEnd" in leaseData
+            and "dateTimeStart" in leaseData
+            and "propertyID" in leaseData
+        ):
```
NickSchimek: We should probably create an issue to refactor this or do it now. I'd like to remove this completely, but that might not be possible without changing the frontend.

There are two issues here that make it hard to read.
1. The comment
2. the multi-line if statement

I would suggest changing it to something like this instead:
```
if _lease():
...
...


def _lease():
     return (
         "dateTimeEnd" in leaseData
         and "dateTimeStart" in leaseData
         and "propertyID" in leaseData
     )
```

runtime.txt
```diff
@@ -1 +1 @@
-python-3.8.2
\ No newline at end of file
+python-3.8.2
```
NickSchimek, line 1: I don't see any difference. What changed here? 🤔

NickSchimek (dismissed):
Thanks for doing all of this work! I have a few suggestions, but nothing that would stop this from being merged.

NickSchimek:
> Also, documentation should probably be made on how to set up Flake8 and Black for use on editors and the command line (e.g., format on save and auto-linting). When used properly in your workflow, you won't have to change much at commit time because most of it will already pass the pre-commit checks.
> 
> This could maybe be a separate issue?

Yes, separate issue is fine. This would be great information to have, but I imagine this info is editor dependent and we can't support docs for every editor. I think it would be best if we could redirect to some external documentation.

runtime.txt
aedwardg: It added a newline.  I'm not sure why GitHub makes it difficult to see whitespace changes sometimes.

CONTRIBUTING.md
aedwardg: I agree.  One thing we could do is add a new custom script in the `[scripts]` section of the pipfile.

For example, we could add this line:
`get_deps = "pipenv install -d --pre"`

Then when we want to install and update dependencies, we would run:
`pipenv run get_deps`

tests/integration/test_leases.py
aedwardg: Cool, I thought so, but wasn't sure.  I'll delete that line from the three tests.

tests/factory_fixtures/contact_number.py
aedwardg: Okay, I'll do that for the four fixtures that were changed like this one.

CONTRIBUTING.md
NickSchimek: That would be great! Perhaps naming it pipenv run dev-install

resources/tenants.py
aedwardg: That makes sense.  I've refactored it and got it working.  If it still bothers us down the line, we can always refactor again 😆

aedwardg:
@NickSchimek thanks for taking the time to review this massive PR!

I made the changes you suggested.  Let me know if you see anything else that needs cleaning up.

NickSchimek (approved):
This looks great! Thank you!

PR #190, opened 2021-01-07
Modified UserSchema to accept and return int values for RoleEnum

### LINKED ISSUE #491: User Schema requires a string for a Role instead of an int
**Describe the bug**
The marshmallow auto schema validations need to be overridden for the Role. Using the schema this way we would need to pass the string values; however it has been decided to use ints. Update the schema role validations to validate on int values.

yunnie:

Closes codeforpdx/dwellingly-app/issues/491

UserSchema now loads and returns integers.

### DIFF
--- a/schemas/user.py
+++ b/schemas/user.py
@@ -1,5 +1,5 @@
 from ma import ma
-from models.user import UserModel
+from models.user import UserModel, RoleEnum
 from utils.time import time_format
 from marshmallow import fields, validates, ValidationError
 
@@ -11,10 +11,21 @@ class UserSchema(ma.SQLAlchemyAutoSchema):
     created_at = fields.DateTime(time_format)
     updated_at = fields.DateTime(time_format)
 
+    role = fields.Method("get_role_value", deserialize="load_role_enum")
+
     @validates("email")
     def validate_uniqueness_of_email(self, value):
         if UserModel.find_by_email(value):
             raise ValidationError(f"A user with email '{value}' already exists")
+
+    def get_role_value(self, obj):
+        return obj.role.value
+
+    def load_role_enum(self, value):
+        if not RoleEnum.has_role(value):
+            raise ValidationError("Invalid role")
+        else:
+            return RoleEnum(value)
 
 
 class UserRegisterSchema(UserSchema):
--- a/tests/schemas/test_user_schema.py
+++ b/tests/schemas/test_user_schema.py
@@ -27,10 +27,16 @@ class UserSchemaValidations:
 
     def test_role_enum_validation(self):
         user_schema = UserSchema()
-        payload = {"role": 6}
+        payload = {"role": 999}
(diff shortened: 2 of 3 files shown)


schemas/user.py
```diff
@@ -11,11 +11,22 @@ class Meta:
     created_at = fields.DateTime(time_format)
     updated_at = fields.DateTime(time_format)
 
+    role = fields.Method("get_role_value", deserialize="load_role_enum")
```
aedwardg, line 14: This is nice.  I hadn't realized Marshmallow let you handle serialization and deserialization all at once like this.

tests/schemas/test_user_schema.py
```diff
@@ -24,6 +25,23 @@ def test_unique_email(self, empty_test_db, create_join_staff):
             f"A user with email '{user.email}' already exists"
         ]
 
+    def test_role_enum_validation(self):
+        user_schema = UserSchema()
+        payload = {"role": 6}
+        validation_error = user_schema.validate(payload)
+        assert "role" in validation_error
+        assert validation_error["role"] == ["Invalid role"]
```
aedwardg: Although this works now, a role with the value `6` could conceivably be added in the future, which would break this test.
I'd prefer to see either an extremely unlikely int, a string, or both here. Perhaps something like this:

```suggestion
        payload = {"role": 999}
        validation_error = user_schema.validate(payload)
        assert "role" in validation_error
        assert validation_error["role"] == ["Invalid role"]
        
        payload = {"role": "STAFF"}
        validation_error = user_schema.validate(payload)
        assert "role" in validation_error
        assert validation_error["role"] == ["Invalid role"]
        
```

I'm also not sure we need to be testing for the specific ValidationError message, but we can leave that in for now, since there's another test that also raises a ValidationError for `role`.

aedwardg (changes requested):
Looks great @yunnie!

Thanks for taking care of this.  I just have one change I'd like made in the `test_role_enum_validation` test before I approve.  If you can think of a better way to implement my suggested change, feel free to do that instead.

schemas/user.py
yunnie: I also considered extending the ```fields.Field``` class to create a field to generalize serializing enums. But that seemed like overkill for just the one. But if we find that we need to something similar for another one, let me know.

aedwardg (approved):
@yunnie

Apologies for having to do the extra commit.  I think I unknowingly added an extra line and some whitespace in my suggested change here on GitHub, so when you clicked "Accept changes" it added those in 😬

Looks like you saw it and fixed it though 😅 so this looks good to me and I see no reason why it shouldn't be merged.

yunnie:
No problem at all! It was easy to spot and fix. And I learned that I shouldn't be so lazy...

NickSchimek (approved):
Nice work! Thanks

PR #191, opened 2021-01-11
Require contact numbers

### LINKED ISSUE #497: Contact Numbers should not be empty for Emergency Contacts
**Purpose**
Ensure contact numbers are not empty when creating emergency contact. 

**Proposed Solution**
- add `required=True` to `contact_numbers` in the `EmergencyContactSchema`.
- update contact numbers validation to check that length of `contact_numbers` is at least 1 

**Acceptance Criteria**
- [ ] have tests!

aedwardg:

Closes codeforpdx/dwellingly-app/issues/497

Summary of changes:

- Updated `EmergencyContactSchema` to require `contact_numbers` field and validate whether `contact_numbers` had a length of less than 1.
- Updated and moved around tests as needed.

Note: I didn't add much to the `integration/test_emergency_contacts` file because the schema should be taking care of all that validation and we are already cover that in the schema tests.  I believe once this refactor is complete we will only need to test the happy path for the resource, as other tests will likely be redundant.

### DIFF
--- a/tests/integration/test_emergency_contacts.py
+++ b/tests/integration/test_emergency_contacts.py
@@ -96,12 +96,6 @@ def test_emergency_contacts_POST(client, auth_headers):
     assert is_valid(
         response, 400
     )  # BAD REQUEST - {'name': 'This Field Cannot Be Blank.'}
-    assert response.json == {
-        "message": {
-            "name": ["Missing data for required field."],
-            "contact_numbers": ["Missing data for required field."],
-        }
-    }
 
 
 def test_emergency_contacts_PUT(client, auth_headers):
--- a/tests/integration/test_emergency_contacts.py
+++ b/tests/integration/test_emergency_contacts.py
@@ -96,7 +96,12 @@ def test_emergency_contacts_POST(client, auth_headers):
     assert is_valid(
         response, 400
     )  # BAD REQUEST - {'name': 'This Field Cannot Be Blank.'}
-    assert response.json == {"message": {"name": ["Missing data for required field."]}}
+    assert response.json == {
+        "message": {
+            "name": ["Missing data for required field."],
+            "contact_numbers": ["Missing data for required field."],
+        }
+    }
 
 
 def test_emergency_contacts_PUT(client, auth_headers):
--- a/models/contact_number.py
+++ b/models/contact_number.py
@@ -7,7 +7,7 @@ class ContactNumberModel(BaseModel):
     __tablename__ = "contact_numbers"
 
     id = db.Column(db.Integer, primary_key=True)
-    number = db.Column(db.String(20), nullable=False)
+    number = db.Column(db.String(30), nullable=False)
     numtype = db.Column(db.String(30))
     extension = db.Column(db.String(10))
     emergency_contact_id = db.Column(
--- a/schemas/emergency_contact.py
+++ b/schemas/emergency_contact.py
@@ -11,9 +11,16 @@ class EmergencyContactSchema(ma.SQLAlchemyAutoSchema):
     created_at = fields.DateTime(time_format)
     updated_at = fields.DateTime(time_format)
 
-    contact_numbers = fields.List(fields.Nested("ContactNumberSchema"))
+    contact_numbers = fields.List(fields.Nested("ContactNumberSchema"), required=True)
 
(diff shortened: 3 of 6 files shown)


whiletrace (commented):
question on tests/schemas/test_contact_number_schema.py line 5 -9  is this a valid format for a phone number `valid_payload = {"number": "503-503-503"}` wouldn't we want to validate a payload such as `valid_payload = {"number": "503-503-5031"}`?  also any string is able to pass validation as long as it is format `valid_payload = "number": "abc-ble-fgh"}` suggest we implement pythons re library for string pattern recognition we could do something like `vailid_phone = re.compile((\d{3}[-\.\s]??\d{3}[-\.\s]??\d{4})` `match = re.fullmatch(valid_phone, "ourcontact#")` if not match validation_error. We could separate into another issue?

aedwardg:
> question on tests/schemas/test_contact_number_schema.py line 5 -9 is this a valid format for a phone number `valid_payload = {"number": "503-503-503"}` wouldn't we want to validate a payload such as `valid_payload = {"number": "503-503-5031"}`? 

Good catch @whiletrace .  This was already there in the [current file](https://github.com/codeforpdx/dwellinglybackend/blob/development/tests/schemas/test_contact_number_schema.py), but I can update it so that it looks more like a regular phone number.

> also any string is able to pass validation as long as it is format `valid_payload = "number": "abc-ble-fgh"}`

This is true, and it goes even deeper than that. 

Currently, `"abcdefghij"` would pass the validation for any of the phone number fields in the app (user, tenant, etc.), not just contact numbers.  The only validations we've done on phone numbers so far has been the length of the string, and even that varies both between frontend/backend and even among files within each part of the app. (See my comment on this problem [here](https://github.com/codeforpdx/dwellingly-app/pull/488#discussion_r551580102)) This is something we need to come to a consensus on as a team so that we can make the min an max lengths uniform across the app.  If I don't remember to bring this up on Wednesday, please speak up because I keep forgetting 😅 

> suggest we implement pythons re library for string pattern recognition we could do something like `vailid_phone = re.compile((\d{3}[-\.\s]??\d{3}[-\.\s]??\d{4})` `match = re.fullmatch(valid_phone, "ourcontact#")` if not match validation_error. We could separate into another issue?

This is a good idea.  I think either the frontend, the backend, or both should be validating phone strings to make sure it's not just a bunch of letters or symbols.  We'll have to be careful not to be overly exclusive though since we don't know what kind of non-standard numbers might be used in this app in the future. 

Let's bring this up at the team meeting also and see what others think about where this validation should be taking place.  If it's wanted in the backend, we can make a separate issue for that and mention that the validation should be for all phone number fields, not just contact numbers.

whiletrace (dismissed):
looks good! yeah, agreed we should bring it up next meeting, and if we get consensus we could separate into separate issue, I think that it makes sense to do it on our end since we are validating payload, If payload doesn't validate because of incorrect number pattern we can send a custom error message that can bubble up to the front end that they can handle how the choose.

tests/integration/test_emergency_contacts.py
```diff
@@ -96,7 +96,12 @@ def test_emergency_contacts_POST(client, auth_headers):
     assert is_valid(
         response, 400
     )  # BAD REQUEST - {'name': 'This Field Cannot Be Blank.'}
-    assert response.json == {"message": {"name": ["Missing data for required field."]}}
+    assert response.json == {
+        "message": {
+            "name": ["Missing data for required field."],
+            "contact_numbers": ["Missing data for required field."],
+        }
```
NickSchimek, line 103: I'd say we can delete this assert. It's retesting what the schema already tests. And testing the response message is testing the Marshmallow library. We can create an issue to refactor these tests. I think all of the tests except the one testing the update action needs to be refactored, because the update action hasn't been refactored yet in the resource.

NickSchimek: I'd say we can delete this assert. It's retesting what the schema already tests. And testing the response message is testing the Marshmallow library. 
We can create an issue to refactor these tests if one doesn't already exist. I think all of the tests except the one testing the update action needs to be refactored, because the update action hasn't been refactored yet in the resource.

NickSchimek (approved):
Looks good! Thanks

PR #195, opened 2021-01-13
#490 lease with get tenants

Protopaco:

Resolves the issue covered in #490
codeforpdx/dwellingly-app#490

These changes as a json() method to "leases" model, in order to return information from database as a json.
Also adds lease array to property object and the tenant object when they are returned. 

I attempted to mimic the style and functionality of existing code base.


Pipfile
```diff
@@ -33,6 +33,7 @@ bcrypt = "==3.1.7"
 flask-marshmallow = "*"
 marshmallow-sqlalchemy = "*"
 alembic = "*"
+black = "*"
```
Chris-Boe: It looks like we already have this listed under dev-packages. Do we need it here for some reason?

Chris-Boe (changes requested):

Pipfile
Protopaco: No, it must have been added automatically while I was adding Black to my system.  Sorry about that!

Chris-Boe (approved):

PR #197, opened 2021-01-14
Prevent users without a role from logging in

### LINKED ISSUE #501: Users without a role should NOT be able to log in
**Purpose**
Step 1 of 4 in transitioning away from the `PENDING` user role.

In preparation for `PENDING` role to be removed, we need to ensure that users not assigned a role cannot login. Once a user is assigned a role, they should be able to log in to the app.

Currently, any user can log in if they are not archived.

**Proposed Solution**
Update the `UserLogin` endpoint to check whether the user has a role, and return a 403 code if they do not (see how archived users are handled in this endpoint).

**Acceptance Criteria**
- [x] A user with no role should **not** be able to log in.
- [x] A test verifying this should be included in `tests/integration/test_users.py`

Chris-Boe:
Closes codeforpdx/dwellingly-app/issues/501

### DIFF
--- a/resources/user.py
+++ b/resources/user.py
@@ -196,7 +196,7 @@ class UserLogin(Resource):
 
         user = UserModel.find_by_email(data["email"])
 
-        if user and user.archived:
+        if user and (user.archived or user.role is None):
             return {"message": "Invalid user"}, 403
 
         if user and user.check_pw(data["password"]):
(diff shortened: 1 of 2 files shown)


aedwardg (approved):
Looks good to me.  I had assumed the test would include the `new_user` pytest fixture (altered to have a role of `None`), but this works as well since it's still checking that a user with no role can't log in.

PR #198, opened 2021-01-15
Set property fields to mandatory

### LINKED ISSUE #500: Make appropriate Property fields required in database
**Purpose**
Ensure Property fields Name, Address, City, State and Zip cannot be null in the database.

**Proposed Solution**
Add `nullable=False` to these fields.
Can add tests to Schema if you're up to it in `tests/schema/test_property_schema.py`

**Acceptance Criteria**
- [ ] All tests are still passing

Chris-Boe:
Closes codeforpdx/dwellingly-app/issues/500

### DIFF
--- a/tests/schemas/test_property_schema.py
+++ b/tests/schemas/test_property_schema.py
@@ -1,5 +1,4 @@
 import pytest
-from utils.time import Time
 from schemas import PropertySchema
 from models.property import PropertyModel
 from db import db
@@ -10,8 +9,6 @@ class TestPropertyManagerValidations:
     def test_valid_payload(self, create_property_manager):
 
         valid_payload = {
-            "created_at": Time.one_year_from_now(),
-            "updated_at": Time.today(),
             "name": "the heights",
             "address": "111 SW Harrison",
             "city": "Portland",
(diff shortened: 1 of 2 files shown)


tests/schemas/test_property_schema.py
```diff
@@ -24,6 +24,21 @@ def test_valid_payload(self, create_property_manager):
         no_validation_errors = {}
         assert no_validation_errors == PropertySchema().validate(valid_payload)
 
+    def test_missing_required_parameters(self, create_property_manager):
+        invalid_payload = {
+            "created_at": Time.one_year_from_now(),
+            "updated_at": Time.today(),
```
aedwardg: Let's get rid of `created_at` and `updated_at` here and also in the test above it (`test_valid_payload`).  These fields are created by the database and shouldn't be included in the payload.  Plus, using `Time.one_year_from_now()` isn't appropriate here (`created_at` shouldn't be one year after `updated_at` 🤔).  It must have slipped in with a previous PR.

Once those are removed, the `utils.time` import won't be necessary either, so let's get rid of that as well.  

Other than that, this test looks good 👍

aedwardg (changes requested):
Thanks for tackling this, @Chris-Boe! It looks good, just a couple quick changes before it's merge-ready.

tests/schemas/test_property_schema.py
Chris-Boe: Ah yeah I just copied and pasted, whoops! I'll update those

aedwardg (approved):
LGTM 👍

PR #199, opened 2021-01-15
Prevent users from archiving or deleting themselves

### LINKED ISSUE #496: Add validation to check that admin (current user) cannot delete themselves
**Purpose**
Per [figma](https://drive.google.com/file/d/1YqboQogczYm1HkyRqEtVSzeQ61T9hWU2/view) designs on page 18 the admin is not allowed to delete themselves.

**Proposed Solution**
Use marshmallow for validation

**Acceptance Criteria**
- [ ] Tests are created to check that an admin (the current_user) cannot delete themselves

Chris-Boe:
Closes codeforpdx/dwellingly-app/issues/496

### DIFF
--- a/resources/user.py
+++ b/resources/user.py
@@ -155,9 +155,13 @@ class User(Resource):
 
     @admin_required
     def delete(self, user_id):
+        if user_id == get_jwt_identity():
+            return {"message": "Cannot delete self"}, 400
+
         user = UserModel.find_by_id(user_id)
         if not user:
             return {"message": "Unable to delete User"}, 400
+
         user.delete_from_db()
         return {"message": "User deleted"}, 200
 
@@ -165,6 +169,9 @@ class User(Resource):
 class ArchiveUser(Resource):
     @admin_required
     def post(self, user_id):
+        if user_id == get_jwt_identity():
+            return {"message": "Cannot archive self"}, 400
+
         user = UserModel.find_by_id(user_id)
         if not user:
             return {"message": "User cannot be archived"}, 400
--- a/tests/integration/test_users.py
+++ b/tests/integration/test_users.py
@@ -129,7 +129,7 @@ def test_user_roles(client, auth_headers):
     assert response.status_code == 200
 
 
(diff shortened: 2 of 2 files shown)


aedwardg (commented):
This looks like it works as intended.  That said, I know @NickSchimek wanted to move away from using `auth_headers` in testing and instead use `admin_header`, `staff_header`, `pm_header`, or `valid_header` to avoid hitting the login endpoint each time we test something that needs an auth header.

Unfortunately, it's not a simple switch to make in this case, since `auth_headers['admin']` uses the same user as the `admin_user` fixture, which makes it easy to test the current admin trying to delete or archive themselves. `admin_header`, on the other hand, creates a new admin user with random attributes, so we'd have to implement a way to get the current user w/that header in order to test with it.

What do you think, @Chris-Boe? Should we just go with this for now, since we know it works and then refactor the tests later?

Chris-Boe:
> This looks like it works as intended. That said, I know @NickSchimek wanted to move away from using `auth_headers` in testing and instead use `admin_header`, `staff_header`, `pm_header`, or `valid_header` to avoid hitting the login endpoint each time we test something that needs an auth header.
> 
> Unfortunately, it's not a simple switch to make in this case, since `auth_headers['admin']` uses the same user as the `admin_user` fixture, which makes it easy to test the current admin trying to delete or archive themselves. `admin_header`, on the other hand, creates a new admin user with random attributes, so we'd have to implement a way to get the current user w/that header in order to test with it.
> 
> What do you think, @Chris-Boe? Should we just go with this for now, since we know it works and then refactor the tests later?

Since we're using auth_headers everywhere, I'd probably make a separate issue for it and refactor all the tests at once

NickSchimek:
FYI: stubbing `get_jwt_identity()` to return the `user_id` your testing against will allow you to test using `valid_header`. It should be straight forward, and I believe there is an example somewhere in the codebase that already does that.

NickSchimek (approved):

PR #200, opened 2021-01-16
Use unit_num for Property field

### LINKED ISSUE #487: Rename Property "unit" field in backend
### **Description**
Currently, there is a "unit" field in the property model and resources.  Looking at the [Figma](https://drive.google.com/file/d/1YqboQogczYm1HkyRqEtVSzeQ61T9hWU2/view), in particular on pages 12, 15, 49 and 50, properties shouldn't have a field called "Unit" (which sounds like the unit number in leases/tenants), but should instead have "Units" to represent the total number of units (or maybe available units??) in that property.

Because the frontend POSTs this form field data as "units", but the backend is looking for "unit", the Add New Property function does not fill in the number of units for that record in the database. (Note: editing a property -- PUT request -- has not been implemented on the front end yet -- see #441)

This was discussed in part in [dwellinglybackend PR 175](https://github.com/codeforpdx/dwellinglybackend/pull/175)

### **To Reproduce**
1. Start both backend and frontend
2. Sign in as an admin
3. Go to Manage Properties
4. Click ADD NEW
5. Fill in the Property Information, including the "Units" field and click SAVE
6. Open database or use CURL GET request on the properties API endpoint to see no number of units recorded for the new property

### **Expected behavior**
Hitting ADD NEW should properly update the db.

### **Proposed Solution**
Change all references to a property's `unit` to either `units` or `num_units` to avoid confusion.  `num_units` would be clearer. 

**Backend files that will need to be changed**:
- [ ] `data/seedData.py`
- [ ] `models/property.py`
- [ ] `resources/property.py`
- [ ] `tests/factory_fixtures/property.py`
- [ ] `tests/integration/properties.py`
- [ ] `tests/schemas/test_property_schema.py`
- [ ] `tests/serializers/property/test_property_serializer.py`
- [ ] The [API Usage](https://github.com/codeforpdx/dwellinglybackend/blob/development/doc/using_the_api.md) page

**Frontend integration:**

Currently, the frontend is referencing this field as `units` in the [addProperty view](https://github.com/codeforpdx/dwellingly-app/blob/development/src/views/addProperty/addProperty.js).
To solve this, we could:

- match the name between the frontend and backend (set both to either `units` or `num_units`); or
- keep `units` on frontend, `num_units` on backend and make sure the backend parses the request data for `units`
  - Note: `resources/property.py` is still using the deprecated `RequestParser`

Chris-Boe:
Closes codeforpdx/dwellingly-app/issues/487

Is a re-seeding of the database necessary? y

### DIFF
--- a/tests/factory_fixtures/property.py
+++ b/tests/factory_fixtures/property.py
@@ -12,7 +12,7 @@ def property_attributes(faker, create_property_manager):
             "name": faker.unique.name(),
             "address": faker.address(),
             "city": faker.city(),
-            "num_units": faker.building_number(),
+            "num_units": faker.random_int(min=1),
             "state": faker.state(),
             "zipcode": faker.postcode(),
             "archived": archived,
--- a/tests/factory_fixtures/property.py
+++ b/tests/factory_fixtures/property.py
@@ -12,7 +12,7 @@ def property_attributes(faker, create_property_manager):
             "name": faker.unique.name(),
             "address": faker.address(),
             "city": faker.city(),
-            "unit": faker.building_number(),
+            "num_units": faker.building_number(),
             "state": faker.state(),
             "zipcode": faker.postcode(),
             "archived": archived,
--- a/tests/schemas/test_property_schema.py
+++ b/tests/schemas/test_property_schema.py
@@ -6,18 +6,9 @@ from db import db
 
 @pytest.mark.usefixtures("empty_test_db")
 class TestPropertyManagerValidations:
-    def test_valid_payload(self, create_property_manager):
+    def test_valid_payload(self, create_property_manager, property_attributes):
 
-        valid_payload = {
-            "name": "the heights",
-            "address": "111 SW Harrison",
-            "city": "Portland",
-            "num_units": 101,
-            "state": "OR",
-            "zipcode": "97207",
-            "propertyManagerIDs": [create_property_manager().id],
-            "archived": False,
-        }
+        valid_payload = property_attributes()
         no_validation_errors = {}
         assert no_validation_errors == PropertySchema().validate(valid_payload)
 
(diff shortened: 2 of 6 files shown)


tests/factory_fixtures/property.py
```diff
@@ -12,7 +12,7 @@ def _property_attributes(archived=False, manager_ids=None):
             "name": faker.unique.name(),
             "address": faker.address(),
             "city": faker.city(),
-            "unit": faker.building_number(),
+            "num_units": faker.building_number(),
```
aedwardg: So `faker.building_number()` actually returns a string of a number, but since Marshmallow tries to convert it to an int during deserialization, this works.  

That said, I'd prefer to pass in an actual integer just to be explicit and replicate the actual data as closely as possible.
I believe `faker.building_number()` was probably only used due to the confusion about the "unit" name before. 

I'd suggest something like this:
```suggestion
            "num_units": faker.random_int(min=1),
```

tests/schemas/test_property_schema.py
```diff
@@ -15,7 +15,7 @@ def test_valid_payload(self, create_property_manager):
             "name": "the heights",
             "address": "111 SW Harrison",
             "city": "Portland",
-            "unit": "101",
+            "num_units": 101,
```
aedwardg: This works, though it would probably be cleaner to change this test to use the `property_attributes` fixture instead of a hand-typed dict as the `valid_payload`:

```python
def test_valid_payload(self, property_attributes):
    valid_payload = property_attributes()
    no_validation_errors = {}
    assert no_validation_errors == PropertySchema().validate(valid_payload)
```

If you don't mind, I think it's worth making that change (and deleting the `utils.time` import) since we're already altering the file in this PR.

aedwardg (changes requested):
Looks great @Chris-Boe! Just a couple of suggested changes I think we should make while we're updating these files.

aedwardg (approved):
LGTM 🙌 🚀

PR #201, opened 2021-01-20
#475 Return a 201 resource created status code in invite users endpoint

ovsyankimo:
### What issue is this solving?
Closes codeforpdx/dwellingly-app/issues/475

### Any helpful knowledge/context for the reviewer?
Is a re-seeding of the database necessary? no? (unsure)
Any new dependencies to install? no
Any special requirements to test? no

### Please make sure you've attempted to meet the following coding standards
- [ ] Code builds successfully
- [ ] Code has been tested and does not produce errors
- [ ] Code is readable and formatted
- [ ] There isn't any unnecessary commented-out code

PR #203, opened 2021-01-20
Document UserInvite

keppel2:
Close #476

### DIFF
--- a/doc/using_the_api.md
+++ b/doc/using_the_api.md
@@ -222,6 +222,19 @@ All endpoints are prefixed with `/api/`
     ]
 ```
 
+#### ENDPOINT: USER INVITE
+
+| method | route           | action                 |
+| :----- | :-------------- | :--------------------- |
+| POST   | `/user/invite`  | Invite a user          |
+
+##### Response (200)
+
+```javascript
+    {
+        "message": "User Invited"
+    }
+```
 
 #### ENDPOINT: WIDGETS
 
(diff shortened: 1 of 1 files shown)


doc/using_the_api.md
```diff
@@ -222,6 +222,19 @@ All endpoints are prefixed with `/api/`
     ]
 ```
 
+#### ENDPOINT: USER INVITE
+
+| method | route           | action                 |
+| :----- | :-------------- | :--------------------- |
+| POST   | `/user/invite`  | Invite a user          |
+
+##### Response (200)
```
aedwardg: As of backend PR #201, we'll be returning a 201 response from this endpoint.  Let's go ahead and reflect that change here. 
```suggestion
##### Response (201)
```

doc/using_the_api.md
```diff
@@ -222,6 +222,19 @@ All endpoints are prefixed with `/api/`
...
+
+```javascript
+    {
+        "message": "User Invited"
+    }
+```
```
aedwardg, line 247: In addition to the response data, I think it would be useful to include a section showing some example request data.
Let's go ahead and add that with the following fields:
- `firstName` (string)
- `lastName` (string)
- `phone` (string)
- `email` (string)`
- `role` (int)

aedwardg (changes requested):
Hey @keppel2, thanks for taking care of this!

The PR looks good, but I have a couple of changes that I think will help future-proof these docs a little more.  If you have the time to make these quick updates that would be awesome.

doc/using_the_api.md
keppel2: Okay. I'll look for some sample request data and add that.

aedwardg: Awesome, thanks!

aedwardg (approved):
LGTM 🚀

PR #205, opened 2021-01-21
Include tenants in property JSON

### LINKED ISSUE #505: New endpoint to return Lease + Tenants on Lease
**Purpose**
We want an endpoint that will give us all the necessary info for the list of tenants on pages 49/50 of the Dwellingly mockups https://drive.google.com/file/d/1YqboQogczYm1HkyRqEtVSzeQ61T9hWU2/view

**Proposed Solution**
* New method in the LeaseModel to query by property and get all tenant information on the lease
* New class/endpoint in Property resource to get all leases + tenant information associated with the property

**Acceptance Criteria**
- [ ]

image

**Additional context**

gussmith23:
Closes codeforpdx/dwellingly-app/issues/505

This change is basically verbatim what @NickSchimek suggested here: https://github.com/codeforpdx/dwellingly-app/issues/505#issuecomment-757426976

This change does not trigger any failed tests, which is surprising for me. Shouldn't there be a test for the property GET route somewhere?

### DIFF
--- a/models/property.py
+++ b/models/property.py
@@ -47,6 +47,12 @@ class PropertyModel(BaseModel):
             "updated_at": Time.format_date(self.updated_at),
         }
 
+    def tenants(self):
+        tenants = []
+        for lease in self.leases:
+            tenants.append(lease.tenant.json())
+        return tenants
+
     @classmethod
     def find_by_name(cls, name):
         return cls.query.filter_by(name=name).first()
--- a/resources/property.py
+++ b/resources/property.py
@@ -78,9 +78,9 @@ class Property(Resource):
 
         if rentalProperty:
             json = rentalProperty.json()
-            json['tenants'] = []
+            json["tenants"] = []
             for lease in rentalProperty.leases:
-                json['tenants'].append(lease.tenant.json())
+                json["tenants"].append(lease.tenant.json())
             return json
 
         return {"message": "Property not found"}, 404
--- a/resources/property.py
+++ b/resources/property.py
@@ -77,11 +77,11 @@ class Property(Resource):
         rentalProperty = PropertyModel.find_by_id(id)
 
         if rentalProperty:
-            json = rentalProperty.json()
-            json["tenants"] = []
+            property = rentalProperty.json()
+            property["tenants"] = []
             for lease in rentalProperty.leases:
-                json["tenants"].append(lease.tenant.json())
-            return json
+                property["tenants"].append(lease.tenant.json())
+            return property
 
         return {"message": "Property not found"}, 404
 
--- a/resources/property.py
+++ b/resources/property.py
@@ -74,16 +74,10 @@ class ArchiveProperties(Resource):
 class Property(Resource):
     @admin_required
     def get(self, id):
-        rentalProperty = PropertyModel.find_by_id(id)
-
-        if rentalProperty:
-            property = rentalProperty.json()
-            property["tenants"] = []
-            for lease in rentalProperty.leases:
-                property["tenants"].append(lease.tenant.json())
-            return property
-
-        return {"message": "Property not found"}, 404
+        property = PropertyModel.find(id)
+        property_json = property.json()
+        property_json["tenants"] = property.tenants()
+        return property_json
 
     @admin_required
     def delete(self, id):
--- a/resources/property.py
+++ b/resources/property.py
@@ -77,7 +77,12 @@ class Property(Resource):
         rentalProperty = PropertyModel.find_by_id(id)
 
         if rentalProperty:
-            return rentalProperty.json()
+            json = rentalProperty.json()
+            json['tenants'] = []
+            for lease in rentalProperty.leases:
+                json['tenants'].append(lease.tenant.json())
+            return json
+
         return {"message": "Property not found"}, 404
 
     @admin_required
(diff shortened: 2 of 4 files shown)


resources/property.py
```diff
@@ -77,7 +77,12 @@ def get(self, id):
         rentalProperty = PropertyModel.find_by_id(id)
 
         if rentalProperty:
-            return rentalProperty.json()
+            json = rentalProperty.json()
+            json["tenants"] = []
+            for lease in rentalProperty.leases:
+                json["tenants"].append(lease.tenant.json())
+            return json
+
```
aedwardg: This is a stylistic thing, but I think we should probably avoid using `json` as a variable name, since it's already used as a library, method and kwarg throughout the app. Perhaps change this to `property_data` or `property_json`?

aedwardg (changes requested):
Looks good @gussmith23! I just have one stylistic suggestion for the code you added.

Also, I think we should probably test this (even though the property integration tests are a mess and need to be refactored 😬 ).  

For now it would be sufficient to add a line to the current `test_get_property_by_id()` test in `tests/integration/test_properties.py`:
```python
assert property_info["tenants"] == [TenantModel.find_by_id(1).json()]
```
(and don't forget to import the TenantModel)

Other that that, this looks good to me.  We can always decide later if we want to refactor this to be a Marshmallow serialization.

resources/property.py
NickSchimek: naming things are hard. I think @aedwardg has a good point though. Using the name json doesn't describe what contents the variable is holding. I'm not too concerned about the data structure the variable is, but rather what it is, which is a property object. so I would suggest `property`.

I would suggest the following change to increase readability, and one or two tests. One unit test on the model, and then one integration test on the resource.

If we use the `find` method then we can remove the conditional too.
```
    def get(self, id):
        property = PropertyModel.find(id).json()
        property["tenants"] = property.tenants()
        return property
```

And in the `PropertyModel` add the tenants method.
```
    def tenants(self):
        tenants = []
        for lease in self.leases:
            tenants.append(lease.tenant.json())
        return tenants
```

gussmith23: Great, renaming to `property_json` given Nick's and your suggestion! Thank you, and good catch!

gussmith23: Variable renamed, and code restructured as you suggest!

I'm trying to write a unit test now. I figured you meant unit-testing the new `tenants()` method at the very least -- if that's the case, I ran into an issue, and I'm wondering if my solution is appropriate. The property fixture `create_property()` seems to create a property without leases. To get a property _with_ leases, I actually used the `create_lease()` fixture, and then used the `lease.propertyID` to `find()` the property and check that `tenants()` works the way we expect.

gussmith23: Integration test updated, and unit test added. Let me know if this is enough testing -- I'm still getting a feel for exactly what to test! It didn't seem like any of the other property integration tests should be updated, at the very least.

tests/unit/test_property.py
```diff
@@ -8,3 +8,9 @@ def setup(self):
         self.object = PropertyModel()
         self.custom_404_msg = "Property not found"
         self.schema = PropertySchema
+
+    def test_tenants_attached_to_property(self, create_lease):
```
aedwardg, line 12: When we are using the `create_{db_entry}` fixtures for tests, we should be using the test databases from now on.  
You can do this a couple of ways. You can add `empty_test_db` as the second parameter (after `self`), or you can add the decorator line 
```python
```
above the class definition. 

I think my preferred method is to use the decorator (since it will apply to all methods within the test class, in case people forget), but either way is valid.

Add this in, and I think we should be ready to merge this. 😄

aedwardg (commented):
@gussmith23 thanks for updating this and adding the tests!

I just have the one comment about using the `empty_test_db`. After that, I think it should be good to go.

tests/unit/test_property.py
NickSchimek: This test actually uses the empty test db through inheritance; otherwise the test would have failed without a db being specified.

models/property.py
```diff
@@ -47,6 +47,12 @@ def json(self):
             "updated_at": Time.format_date(self.updated_at),
         }
 
+    def tenants(self):
+        tenants = []
+        for lease in self.leases:
+            tenants.append(lease.tenant.json())
+        return tenants
```
NickSchimek, line 54: I would really like to know if sql alchemy offers a way to declare that a property has many tenants through the lease relationship. If not, it would probably be fun to create our own dsl to declare those types of relationships On the model.

NickSchimek (approved):
Looks good @gussmith23 ! Thank you!

NickSchimek:
> This change does not trigger any failed tests, which is surprising for me. Shouldn't there be a test for the property GET route somewhere?

Yes, good question. You are absolutely right. This change should've caused a failure since it's a change in behavior. Many of the tests need to be refactored and improved. This is one of them. You can see the test for the get endpoint here. 'Tests/integration/_propert.py' well I'm on my phone and I can't type. test_propert.py

tests/unit/test_property.py
aedwardg: Oh, you're right! Totally missed that!

aedwardg (approved):
Thanks @gussmith23 !

NickSchimek:
@aedwardg one of things we will probably need to do in the future is to only return active tenants for a property that have active leases. Might be good to create an issue for this and mark it as not ready, unless it is.

gussmith23:
Thanks all for the reviews, and for the answers to my questions!

aedwardg:
> 
> 
> @aedwardg one of things we will probably need to do in the future is to only return active tenants for a property that have active leases. Might be good to create an issue for this and mark it as not ready, unless it is.

@NickSchimek good call, I'll open up that issue ASAP

PR #206, opened 2021-01-22
Issue502 Create endpoint to return users that are pending authorization

### LINKED ISSUE #502: Add ability to query all users who have no role
**Purpose**
Step 2 of 4 in transitioning away from the `PENDING` user role.

In preparation for `PENDING` role to be removed, we need a way to access all users who have not been assigned a role.

Currently, the frontend grabs these "pending" users through the `UsersRole` endpoint, by providing a `userrole` with the Enum value for the `PENDING` role.

**Proposed Solution**
A new endpoint should probably be created to handle this, since there will no longer be a role to pass to the `UsersRole` endpoint.

**Acceptance Criteria**
- [ ] Implement a way to access a list of all users who have no role.
- [ ] Create appropriate integration tests for this functionality

ovsyankimo:
Closes codeforpdx/dwellingly-app/issues/502

Is a re-seeding of the database necessary? yes
Any special requirements to test? ~~@aedwardg mentioned that we should have had to update `UserSchema` to allow a user to have role=None, but we didn't and we're not seeing any test failures, so there may be a testing gap there.~~ I think this got resolved

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -38,20 +38,6 @@ def admin_user():
 
 
 @pytest.fixture
-def new_user_no_role():
-    newUserNoRole = UserModel(
-        email="missy@dwelling.org",
-        password=plaintext_password,
-        firstName="Missy",
-        lastName="Role",
-        phone="503-555-5557",
-        role=None,
-        archived=0,
-    )
-    return newUserNoRole
-
-
-@pytest.fixture
 def new_user():
     newUser = UserModel(
         email="someone@domain.com",
@@ -142,14 +128,13 @@ def pm_header(create_property_manager):
 
 
 @pytest.fixture
-def test_database(app, admin_user, new_user, new_user_no_role, property_manager_user):
+def test_database(app, admin_user, new_user, property_manager_user):
     db.create_all()
 
     seedData()
 
     db.session.add(admin_user)
     db.session.add(new_user)
-    db.session.add(new_user_no_role)
     db.session.add(property_manager_user)
     db.session.commit()
 
--- a/conftest.py
+++ b/conftest.py
@@ -142,13 +142,14 @@ def pm_header(create_property_manager):
 
 
 @pytest.fixture
-def test_database(app, admin_user, new_user, property_manager_user):
+def test_database(app, admin_user, new_user, new_user_no_role, property_manager_user):
     db.create_all()
 
     seedData()
 
     db.session.add(admin_user)
     db.session.add(new_user)
+    db.session.add(new_user_no_role)
     db.session.add(property_manager_user)
     db.session.commit()
 
--- a/conftest.py
+++ b/conftest.py
@@ -38,6 +38,20 @@ def admin_user():
 
 
 @pytest.fixture
+def new_user_no_role():
+    newUserNoRole = UserModel(
+        email="missy@dwelling.org",
+        password=plaintext_password,
+        firstName="Missy",
+        lastName="Role",
+        phone="503-555-5557",
+        role=None,
+        archived=0,
+    )
+    return newUserNoRole
+
+
+@pytest.fixture
 def new_user():
     newUser = UserModel(
         email="someone@domain.com",
--- a/data/seedData.py
+++ b/data/seedData.py
@@ -11,7 +11,7 @@ from models.emergency_contact import EmergencyContactModel
 from models.contact_number import ContactNumberModel
 from models.lease import LeaseModel
 from utils.time import time_format
-from schemas import PropertySchema, TenantSchema
+from schemas import PropertySchema, TenantSchema, UserRegisterSchema
 
 
 def seedData():
@@ -68,45 +68,48 @@ def seedData():
         archived=False,
     )
     user_gray_pouponn.save_to_db()
-
-    user_anthony_redding = UserModel(
-        email="pending1@dwellingly.org",
-        role=RoleEnum.PENDING,
-        firstName="Anthony",
-        lastName="Redding",
-        password="1234",
-        phone="555-555-5555",
-        archived=False,
+    user_anthony_redding = UserModel.create(
+        payload={
+            "email": "pending1@dwellingly.org",
+            "firstName": "Anthony",
+            "lastName": "Redding",
+            "password": "1234",
+            "phone": "555-555-5555",
+        },
+        schema=UserRegisterSchema,
     )
     user_anthony_redding.save_to_db()
-    user_ryan_dander = UserModel(
-        email="pending2@dwellingly.org",
-        role=RoleEnum.PENDING,
-        firstName="Ryan",
-        lastName="Dander",
-        password="1234",
-        phone="555-555-5555",
-        archived=False,
+    user_ryan_dander = UserModel.create(
+        payload={
+            "email": "pending2@dwellingly.org",
+            "firstName": "Ryan",
+            "lastName": "Dander",
+            "password": "1234",
+            "phone": "555-555-5555",
+        },
+        schema=UserRegisterSchema,
     )
     user_ryan_dander.save_to_db()
-    user_amber_lemming = UserModel(
-        email="pending3@dwellingly.org",
-        role=RoleEnum.PENDING,
-        firstName="Amber",
-        lastName="Lemming",
-        password="1234",
-        phone="555-555-5555",
-        archived=False,
+    user_amber_lemming = UserModel.create(
+        payload={
+            "email": "pending3@dwellingly.org",
+            "firstName": "Amber",
+            "lastName": "Lemming",
+            "password": "1234",
+            "phone": "555-555-5555",
+        },
+        schema=UserRegisterSchema,
     )
     user_amber_lemming.save_to_db()
-    user_jeremy_quazar = UserModel(
-        email="pending4@dwellingly.org",
-        role=RoleEnum.PENDING,
-        firstName="Jeremy",
-        lastName="Quazar",
-        password="1234",
-        phone="555-555-5555",
-        archived=False,
+    user_jeremy_quazar = UserModel.create(
+        payload={
+            "email": "pending4@dwellingly.org",
+            "firstName": "Jeremy",
+            "lastName": "Quazar",
+            "password": "1234",
+            "phone": "555-555-5555",
+        },
+        schema=UserRegisterSchema,
     )
     user_jeremy_quazar.save_to_db()
     user_janice_joinstaff = UserModel(
@@ -139,36 +142,6 @@ def seedData():
         archived=False,
     )
     user_xander_dander.save_to_db()
-
-    user_bobby_lowroll = UserModel(
-        email="broll@joinpdx.org",
-        firstName="Bobby",
-        lastName="Lowroll",
-        password="1234",
-        phone="555-555-5673",
-        archived=False,
-    )
-    user_bobby_lowroll.save_to_db()
-
-    user_cindy_norole = UserModel(
-        email="cindynorole@joinpdx.org",
-        firstName="Cindy",
-        lastName="Norole",
-        password="1234",
-        phone="555-555-5555",
-        archived=False,
-    )
-    user_cindy_norole.save_to_db()
-
-    user_joe_zero = UserModel(
-        email="joez@joinpdx.org",
-        firstName="Joe",
-        lastName="Zero",
-        password="1234",
-        phone="555-345-5555",
-        archived=False,
-    )
-    user_joe_zero.save_to_db()
 
     property_test1 = PropertyModel.create(
         payload={
--- a/data/seedData.py
+++ b/data/seedData.py
@@ -140,6 +140,36 @@ def seedData():
     )
     user_xander_dander.save_to_db()
 
+    user_bobby_lowroll = UserModel(
+        email="broll@joinpdx.org",
+        firstName="Bobby",
+        lastName="Lowroll",
+        password="1234",
+        phone="555-555-5673",
+        archived=False,
+    )
+    user_bobby_lowroll.save_to_db()
+
+    user_cindy_norole = UserModel(
+        email="cindynorole@joinpdx.org",
+        firstName="Cindy",
+        lastName="Norole",
+        password="1234",
+        phone="555-555-5555",
+        archived=False,
+    )
+    user_cindy_norole.save_to_db()
+
+    user_joe_zero = UserModel(
+        email="joez@joinpdx.org",
+        firstName="Joe",
+        lastName="Zero",
+        password="1234",
+        phone="555-345-5555",
+        archived=False,
+    )
+    user_joe_zero.save_to_db()
+
     property_test1 = PropertyModel.create(
         payload={
             "name": "test1",
--- a/data/seedData.py
+++ b/data/seedData.py
@@ -68,7 +68,7 @@ def seedData():
         archived=False,
     )
     user_gray_pouponn.save_to_db()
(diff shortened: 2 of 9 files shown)


tests/integration/test_users.py
```diff
@@ -140,6 +140,17 @@ def test_user_roles(client, auth_headers):
     assert response.status_code == 200
 
 
+def test_user_no_roles(client, auth_headers, test_database):
```
ovsyankimo: We followed the way another integration test was written and had the assertion check not only users seeded in the database but also a user added via the test fixtures. This felt strange, as it seemed like the test should have only checked the seeded database OR the test database, not both. If there's not a particular reason to check both I'd like to just check the ones in the seeded database.

conftest.py
```diff
@@ -37,6 +37,20 @@ def admin_user():
...
+    newUserNoRole = UserModel(
+        email="missy@dwelling.org",
+        password=plaintext_password,
+        firstName="Missy",
+        lastName="Role",
+        phone="503-555-5557",
+        role=None,
+        archived=0,
+    )
+    return newUserNoRole
```
aedwardg: Since the `new_user` fixture should create user with no role once [issue 504](https://github.com/codeforpdx/dwellingly-app/issues/504) is completed, we shouldn't actually need this.

resources/user.py
```diff
@@ -26,6 +26,27 @@ def get(self):
         return result, 200
 
 
+class UsersNoRole(Resource):
+    @admin_required
+    def get(self):
+        users = [user.json() for user in UserModel.find_by_role(None)]
```
aedwardg: This works fine.  I'm tempted to ask for a new classmethod in `UserModel` to return these (e.g., `find_unassigned_users`), rather than using `find_by_role`.  Since technically these users no longer have a role, using `find_by_role` just seems less correct.

If you feel like going that route, that would be cool, but I'll still approve it if you leave this line as-is.

resources/user.py
```diff
@@ -26,6 +26,27 @@ def get(self):
         return result, 200
 
 
+class UsersNoRole(Resource):
```
aedwardg: I think I'd prefer `UsersPending` for the name of this resource.  Although we're removing the actual PENDING role, these are still conceptually "pending users" since they can't use the app until they're assigned a role.  Plus, `/api/users/pending` makes a nicer URL 🙂

conftest.py
```diff
@@ -128,13 +142,14 @@ def pm_header(create_property_manager):
...
 @pytest.fixture
-def test_database(app, admin_user, new_user, property_manager_user):
+def test_database(app, admin_user, new_user, new_user_no_role, property_manager_user):
     db.create_all()
 
     seedData()
 
     db.session.add(admin_user)
     db.session.add(new_user)
+    db.session.add(new_user_no_role)
```
aedwardg: See my comment re: testing.

tests/integration/test_users.py
aedwardg: So, many of our tests still need to be refactored 😅 
However, in this case, I'd still like to use either the `test_database` or `empty_test_db` fixtures for the tests rather than the real database.

`test_database` seeds itself with all the same info as the normal database and then adds some new users via `db.session.add()` (see my comment on `conftest.py`).  This is why it looks like the assertion is checking both.  That said, if you delete the `new_user_no_role` fixture and then remove that from the `test_database` fixture, this assertion will just check for the number of users w/no role that are seeded (since an extra one won't be added in upon creating the test db).

Alternatively, you could use the `empty_test_db` and `create_unauthorized_user` fixtures and then check for just the one new user.  Either way would be acceptable.

aedwardg: One final change I'd like on this test is to switch it over to the ~~`admin_header`~~ `valid_header` fixture.  

It's hard to tell from looking through the codebase, but we're trying to migrate to these individual headers rather than the `auth_header` fixture, since `auth_header` hits (and thus tests) the login endpoint every single time.
 
```suggestion
def test_user_no_roles(client, test_database, valid_header):
```

You'll also need to change line 148 over to `headers=admin_header`

models/user.py
```diff
@@ -87,7 +87,7 @@ def json(self):
             "lastName": self.lastName,
             "email": self.email,
             "phone": self.phone,
-            "role": self.role.value,
+            "role": self.role.value if self.role else None,
```
aedwardg, line 91: Looks good! 👍

data/seedData.py
```diff
@@ -140,6 +140,36 @@ def seedData():
...
+    user_joe_zero = UserModel(
+        email="joez@joinpdx.org",
+        firstName="Joe",
+        lastName="Zero",
+        password="1234",
+        phone="555-345-5555",
+        archived=False,
+    )
+    user_joe_zero.save_to_db()
+
```
aedwardg: We don't really need more users in the seed file, we should probably just delete the `role=RoleEnum.Pending,` line from the existing ones that have that role. But feel free to leave your favorite one in there if you'd like 🙂 

Also, it looks like the reason there were no schema validation errors is because this file hasn't been fully switched over to use the schemas 😬.  Ideally, these should look more like the property entries (line 178+) except without variable names since the variables will no longer be used:
```python
# Pending users
    UserModel.create(
        payload={
            "email": "pending1@dwellingly.org",
            "firstName": "Anthony",
            "lastName": "Redding",
            "password": "1234",
            "phone": "555-555-5555",
        },
        schema=UserRegisterSchema,
    )
```

If you're up for it, I'd ask that all the "pending" users get updated to use this format.  I will open up an issue for refactoring the rest of the `seedData.py` entries to use schema validations as well.

aedwardg (changes requested):
Thanks for the excellent PR @ovsyankimo & @russellneville!

Everything you two added works well as-is, but since we're in the home stretch of having a working MVP for this app I'm being extra picky on code additions. I'm also pushing hard to enforce refactoring _pre_-merge so we don't have to spend as long re-working files as we slowly shift over from the previous backend design.  

Thanks in advance for your understanding! Let me know if you have any questions regarding my comments/suggestions below.

conftest.py
NickSchimek: This fixture can be removed. A fixture already exists that creates a user without a role. You can use `create_unauthorized_user` see https://github.com/codeforpdx/dwellinglybackend/blob/development/tests/factory_fixtures/user.py#L56

resources/user.py
NickSchimek: One thing we'll need to be sure of is that we only return users that are `pending` authorization. So we'll need to search for Users that do not have a role and are NOT archived.

resources/user.py
NickSchimek: I agree we should use this endpoint name `/api/users/pending` as a user with no role is different than a user that is pending in the context of this app. As there can be users without a role that had their request for access denied.

tests/integration/test_users.py
NickSchimek: > One final change I'd like on this test is to switch it over to the admin_header fixture

Do you mean `valid_header` ?

I would suggest that we should create a new test file to put these tests in since there is a new class that is being created. Probably put the new class in a separate file too to keep things organized.
`app/resources/users/pending_users.py` <- possible location of new class
`tests/integration/users/test_pending_users.py` <- possible location of test

I would suggest taking a look at the class based tests that use the `empty_test_db` as the pattern for emulating anything new that is written.

models/user.py
NickSchimek: yep looks good. I wonder if we can override the `value` method on the role enum to take care of this for us. Then we don't have to remember to add this conditional wherever we want to use `role.value`

tests/integration/test_users.py
aedwardg: > Do you mean valid_header ?

Yes! Thanks for catching that @NickSchimek! 
I saw `auth_headers["admin"]` being used and I immediately wrote `admin_header`, but you're right &mdash; this endpoint doesn't specifically require admin authorization. I've changed my above comment to reflect that.

tests/integration/users/test_pending_users.py
```diff
@@ -0,0 +1,19 @@
...
+        assert pending_user.role is None
+        assert pending_user.archived is False
+
+        """The get pending users returns one user with no role
+        and a successful response code."""
+        response = client.get(
+            "/api/users/pending",
+            headers=valid_header,
+        )
+        print(response)
```
aedwardg: Let's remove the print statement here.

Otherwise, looks good!

data/seedData.py
```diff
@@ -68,45 +68,48 @@ def seedData():
...
+        payload={
+            "email": "pending1@dwellingly.org",
+            "firstName": "Anthony",
+            "lastName": "Redding",
+            "password": "1234",
+            "phone": "555-555-5555",
+        },
+        schema=UserRegisterSchema,
     )
     user_anthony_redding.save_to_db()
```
aedwardg: For each of these four users that you changed over to use the schema, we don't need the `save_to_db()` line anymore, since `UserModel.create()` already saves it to the db.  Let's remove this line.

Once that line is removed, flake8 will complain that you have unused variables, so just remove the variable assignments:
```python
UserModel.create(
    payload={
        "email": "pending1@dwellingly.org",
        "firstName": "Anthony",
        "lastName": "Redding",
        "password": "1234",
        "phone": "555-555-5555",
    },
    schema=UserRegisterSchema,
)
```

aedwardg (commented):
@ovsyankimo thanks for making all those changes! This looks really good.

I just have two very quick additional changes (deleting things) in the comments and then this should be ready to merge 😄

aedwardg (dismissed):
Looks good to me! Thanks again @ovsyankimo

models/user.py
```diff
@@ -128,6 +129,10 @@ def find_by_role_and_name(cls, role, name):
             & (UserModel.firstName.ilike(likeName) | UserModel.lastName.ilike(likeName))
         ).all()
 
+    @classmethod
+    def find_unassigned_users(cls):
```
NickSchimek: I think this method name is a bit vague. Unassigned can mean a multitude of things depending on the context, and I have to look at the code in the method to determine what this method should do. Perhaps we can rename this to `pending` or `pending_authorization` or `without_assigned_role` or something that is a bit more specific as what the method should return.

The query `User.pending` returns pending users in contrast to the query `User.find_unassigned_users` returns unassigned users, but what exactly are they not assigned to?

Depending on how this app evolves we may actually assign other things other than roles to users. For instance our Staff Users can be assigned tenants, so at first I wasn't sure if this should return a list of Users who are not assigned any tenants, and had to look at the code to determine what was going on here.

tests/integration/users/test_pending_users.py
```diff
@@ -0,0 +1,18 @@
+import pytest
+
+
+@pytest.mark.usefixtures("empty_test_db")
+class TestPendingUsers:
+    def test_users_pending(self, client, valid_header, create_unauthorized_user):
+        pending_user = create_unauthorized_user()
+        assert pending_user.role is None
+        assert pending_user.archived is False
```
NickSchimek: These assertions are not testing the endpoint here. These can be removed as the `create_unauthorized_user` already has its own tests in `tests/unit/test_user.py` and it does not need to be retested here.

tests/integration/users/test_pending_users.py
```diff
@@ -0,0 +1,18 @@
...
+
+        """The get pending users returns one user with no role
+        and a successful response code."""
+        response = client.get(
+            "/api/users/pending",
+            headers=valid_header,
+        )
+        assert len(response.get_json()["users"]) == 1
```
NickSchimek: What we want to test here is that we get back the expected json response, not the length of the response. We want to assert the actual response so that we can be confident that we are returning the correct data to the frontend.

resources/users/pending_users.py
```diff
@@ -0,0 +1,24 @@
...
+        ret = [
+            {
+                "id": user["id"],
+                "firstName": user["firstName"],
+                "lastName": user["lastName"],
+                "email": user["email"],
+                "phone": user["phone"],
+                "role": user["role"],
+                "tickets": "TODO",
+                "tenants": "TODO",
```
NickSchimek: There is no need to return this data. Whatever the `.json()` method returns will be good enough.

resources/users/pending_users.py
```diff
@@ -0,0 +1,24 @@
...
+            }
+            for user in users
+        ]
+
+        return {"users": ret}, 200
```
NickSchimek: FYI: No need to explicitly return a status code here, as it will return 200 by default.

NickSchimek (changes requested):
Thanks for the PR! It looks real good. However, there are a couple things I would like to see changed. 
1. Regarding the response of the endpoint. We seem to be returning more data than what is required. All we need to return for each pending user is the data that our `json` method  builds for us.
2. We need a test for the endpoint that checks the actual json response is what we expect it to be.

If the json method is under unit test (if it isn't it would be a good time to add one) then we can assert something like:
```
assert response.get_json() == {"users": [pending_user.json]}
```

If your up for the challenge, I'd like to see the json method in the UserModel work with single user instances or a list of user instances. Then there would be no need to do list comprehension in the endpoints that require us to return a list of Users.

And our endpoint can simply be:
```
return {"users": UserModel.with_unassigned_role().json()}
```

models/user.py
NickSchimek: I think I would lean toward `pending_authorization` or `without_assigned_role` or something similar as the same thing can probably be said about the method name `pending`

NickSchimek:
Actually the endpoint could be this
```
return UserModel.with_unassigned_role().json()
```
If you don't want to update the json method to handle this then it is fine. I just mention it in case it sounds fun.

Most important thing is to test the return is sending back the result we need

ovsyankimo:
> Actually the endpoint could be this
> 
> ```
> return UserModel.with_unassigned_role().json()
> ```
> 
> If you don't want to update the json method to handle this then it is fine. I just mention it in case it sounds fun.
> 
> Most important thing is to test the return is sending back the result we need

Hey Nick,
I'd be happy to work on changing the `json` method. I'm not sure how you all do ticket management, but it feels like it might be nice to have it in a separate ticket though so that #502 can be merged and whatever stuff is waiting on it can proceed. That'd be my vote, and I can figure out how to make the ticket if you'd like. Just let me know!

aedwardg (approved):
Looks good to me!
The name `find_users_without_assigned_role()` seems a little long, but it is clearer, so I think it's fine.

models/user.py
```diff
@@ -128,6 +129,10 @@ def find_by_role_and_name(cls, role, name):
             & (UserModel.firstName.ilike(likeName) | UserModel.lastName.ilike(likeName))
         ).all()
 
+    @classmethod
+    def find_users_without_assigned_role(cls):
+        return cls.query.filter(and_(cls.role.is_(None), cls.archived.is_(False)))
+
```
NickSchimek, line 135: I think this is OK, but I'd like to get this unit tested. If you'd like to do this that would be great; otherwise let's get an issue created to test all happy and sad paths. Also, the name is much better, however the naming contradicts with what it actually does. It says it will return all users without an assigned role, but it returns active users without an assigned role. Although that doesn't really matter. The most important thing is to get some tests so that we can document what this does and build our confidence that it does what it says it does.

NickSchimek (approved):
Thanks @ovsyankimo - There's a lot of code in the repo that conflicts with the style that we're trying to change to. sorry about that. Just one comment below, however this looks good, and looking forward to future PR's from you. Thank you.

PR #209, opened 2021-02-02
Allow property to be updated when the name has not changed

### LINKED ISSUE #521: Property Schema requires unique name for PUT request
**Describe the bug**
Property Schema returns "A property with this name already exists" error in response to PUT requests where property name hasn't changed.  Error encountered on "edit properties" page (ticket #441).  User can update any and all property fields, so long as name has also changed.  If user attempts to change other attributes (address, num of units, etc) when not changing the name, the error occurs.

This functionality makes sense for creating a page, but seems like a bug when updating certain fields.

The backend code that I believe to be involved is at /schemas/property.py (lines 27:30)

```
    @validates("name")
    def validates_uniqueness_of_name(self, value):
        if PropertyModel.find_by_name(value):
            raise ValidationError("A property with this name already exists")
```

Is it possible to have this test only run when creating new property?

**Expected behavior**
User should be able to update property info without changing the property name.

**Screenshots**
Screen Shot 2021-02-01 at 6 12 26 PM

 - OS: Catalina 10.15
 - Browser: Chrome
 - Version 88

**Additional context**

aedwardg:

Closes codeforpdx/dwellingly-app/issues/521
~Closes issue 522~ --moved to a new PR

Assuming we want to handle this on the backend rather than requiring the frontend to only send updated info in payload, this PR solves issue of unique name validation in the PropertySchema.

- Updates property endpoint PUT function to validate using a context dictionary 
- New `PropertySchema` name validation uses a conditional based on the schema's context. This allows for non-unique names in payload if name is unchanged, but validates for uniqueness otherwise
- Refactors existing `update_property_by_id` integration test
- Adds schema tests for new validations

### DIFF
--- a/models/base_model.py
+++ b/models/base_model.py
@@ -36,9 +36,9 @@ class BaseModel(db.Model):
         return obj
 
     @classmethod
-    def update(cls, schema, id, payload):
+    def update(cls, schema, id, payload, context=None):
         obj = cls.find(id)
-        attrs = cls.validate(schema, payload, partial=True)
+        attrs = cls.validate(schema, payload, context=context, partial=True)
 
         for k, v in attrs.items():
             setattr(obj, k, v)
@@ -48,9 +48,14 @@ class BaseModel(db.Model):
         return obj
 
     @staticmethod
-    def validate(schema, payload, partial=False):
+    def validate(schema, payload, context=None, partial=False):
         try:
-            return schema().load(payload, unknown=EXCLUDE, partial=partial)
+            if context:
+                return schema(context=context).load(
+                    payload, unknown=EXCLUDE, partial=partial
+                )
+            else:
+                return schema().load(payload, unknown=EXCLUDE, partial=partial)
         except ValidationError as err:
             abort(400, err.messages)
 
--- a/models/base_model.py
+++ b/models/base_model.py
@@ -50,12 +50,9 @@ class BaseModel(db.Model):
     @staticmethod
     def validate(schema, payload, context=None, partial=False):
         try:
-            if context:
-                return schema(context=context).load(
-                    payload, unknown=EXCLUDE, partial=partial
-                )
-            else:
-                return schema().load(payload, unknown=EXCLUDE, partial=partial)
+            return schema(context=context).load(
+                payload, unknown=EXCLUDE, partial=partial
+            )
         except ValidationError as err:
             abort(400, err.messages)
 
--- a/schemas/property.py
+++ b/schemas/property.py
@@ -2,7 +2,7 @@ from ma import ma
 from models.property import PropertyModel
 from models.user import UserModel
 from schemas.property_assignment import PropertyAssignSchema
-from marshmallow import fields, validates, ValidationError, post_load
+from marshmallow import fields, validates, ValidationError, post_load, pre_load
 from utils.time import time_format
 
 
@@ -36,4 +36,14 @@ class PropertySchema(ma.SQLAlchemyAutoSchema):
                 UserModel.find(manager) for manager in data["propertyManagerIDs"]
             ]
             del data["propertyManagerIDs"]
+        return data
+
+
+class PropertyUpdateSchema(PropertySchema):
+    @pre_load
+    def ignore_name_if_unchanged(self, data, **kwargs):
+        property_to_update = PropertyModel.find(data["id"])
+        if "name" in data and data["name"] == property_to_update.name:
+            del data["name"]
+        del data["id"]
         return data
--- a/schemas/property.py
+++ b/schemas/property.py
@@ -26,9 +26,12 @@ class PropertySchema(ma.SQLAlchemyAutoSchema):
 
     @validates("name")
     def validates_uniqueness_of_name(self, value):
-        if "name" in self.context and value == self.context["name"]:
-            return value
-        if PropertyModel.find_by_name(value):
+        def _assigning_name():
+            if "name" in self.context and value == self.context["name"]:
+                return False
+            return True
+
+        if _assigning_name() and PropertyModel.find_by_name(value):
             raise ValidationError("A property with this name already exists")
 
     @post_load
--- a/schemas/property.py
+++ b/schemas/property.py
@@ -2,7 +2,7 @@ from ma import ma
 from models.property import PropertyModel
 from models.user import UserModel
 from schemas.property_assignment import PropertyAssignSchema
-from marshmallow import fields, validates, ValidationError, post_load, pre_load
+from marshmallow import fields, validates, ValidationError, post_load
 from utils.time import time_format
 
 
@@ -26,6 +26,8 @@ class PropertySchema(ma.SQLAlchemyAutoSchema):
 
     @validates("name")
     def validates_uniqueness_of_name(self, value):
+        if "name" in self.context and value == self.context["name"]:
+            return value
         if PropertyModel.find_by_name(value):
             raise ValidationError("A property with this name already exists")
 
@@ -36,14 +38,4 @@ class PropertySchema(ma.SQLAlchemyAutoSchema):
                 UserModel.find(manager) for manager in data["propertyManagerIDs"]
             ]
             del data["propertyManagerIDs"]
-        return data
-
-
-class PropertyUpdateSchema(PropertySchema):
-    @pre_load
-    def ignore_name_if_unchanged(self, data, **kwargs):
-        property_to_update = PropertyModel.find(data["id"])
-        if "name" in data and data["name"] == property_to_update.name:
-            del data["name"]
-        del data["id"]
         return data
--- a/tests/integration/test_properties.py
+++ b/tests/integration/test_properties.py
@@ -1,8 +1,9 @@
 import pytest
+import json
+from conftest import is_valid
 from models.property import PropertyModel
 from models.user import UserModel
 from models.tenant import TenantModel
-import json
 
 
 def test_get_properties(client, test_database):
@@ -215,15 +216,20 @@ def test_delete_property_by_id(client, auth_headers, test_database):
     assert response == 404
 
 
-def test_update_property_by_id(client, auth_headers, create_property, test_database):
+def test_update_property_by_id(client, empty_test_db, valid_header, create_property):
     test_property = create_property()
     new_property_address = "123 NE Flanders St"
-    test_property.address = new_property_address
+
+    property_json = test_property.json()
+
+    property_json["address"] = new_property_address
+
+    response = client.put(
+        f"/api/properties/{test_property.id}",
+        json=property_json,
+        headers=valid_header,
+    )
+    assert is_valid(response, 200)
 
     """The property should have a new address"""
-    test_changed_property = client.get(
-        f"/api/properties/{test_property.id}", headers=auth_headers["admin"]
-    )
-    test_changed_property = json.loads(test_changed_property.data)
-
-    assert test_changed_property["address"] == new_property_address
+    assert response.json["address"] == new_property_address
--- a/tests/integration/test_properties.py
+++ b/tests/integration/test_properties.py
@@ -1,6 +1,5 @@
 import pytest
 import json
-from conftest import is_valid
 from models.property import PropertyModel
 from models.user import UserModel
 from models.tenant import TenantModel
@@ -216,7 +215,7 @@ def test_delete_property_by_id(client, auth_headers, test_database):
     assert response == 404
 
 
-def test_update_property_by_id(client, empty_test_db, valid_header, create_property):
+def test_update_property(client, empty_test_db, valid_header, create_property):
     test_property = create_property()
     new_property_address = "123 NE Flanders St"
 
@@ -229,7 +228,7 @@ def test_update_property_by_id(client, empty_test_db, valid_header, create_prope
         json=property_json,
         headers=valid_header,
     )
-    assert is_valid(response, 200)
+    assert response == 200
 
     """The property should have a new address"""
     assert response.json["address"] == new_property_address
--- a/tests/integration/test_properties.py
+++ b/tests/integration/test_properties.py
@@ -223,12 +223,31 @@ def test_update_property(client, empty_test_db, valid_header, create_property):
 
     property_json["address"] = new_property_address
 
+    """
+    Updating property info should be successful when the payload includes
+    the property's current name. The response JSON should reflect these changes.
+    """
     response = client.put(
         f"/api/properties/{test_property.id}",
         json=property_json,
         headers=valid_header,
     )
     assert response == 200
+    assert response.json == property_json
 
-    """The property should have a new address"""
-    assert response.json["address"] == new_property_address
+    new_property = create_property()
+    duplicate_name = test_property.name
+
+    new_property_json = new_property.json()
+    new_property_json["name"] = duplicate_name
+
+    """
+    The server responds with a 400 error when attempting to update a
+    property with the name of another existing property.
+    """
+    response = client.put(
+        f"/api/properties/{new_property.id}",
+        json=new_property_json,
+        headers=valid_header,
+    )
+    assert response == 400
--- a/tests/schemas/test_property_schema.py
+++ b/tests/schemas/test_property_schema.py
@@ -61,7 +61,7 @@ class TestPostLoadDeserialization:
         prop = create_property()
         pm_2 = create_property_manager()
         pm_3 = create_property_manager()
-        payload = {"id": prop.id, "propertyManagerIDs": [pm_2.id, pm_3.id]}
+        payload = {"propertyManagerIDs": [pm_2.id, pm_3.id]}
         context = {"request_type": "PUT", "name": prop.name}
 
         PropertyModel.update(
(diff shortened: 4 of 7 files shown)


aedwardg:
Working through the possibility of using original `PropertySchema` and passing in context instead of creating a new schema for updates.  Hold off on reviewing for now.

aedwardg:
Currently waiting on some quasi-related changes from @DaveTrost that will solve https://github.com/codeforpdx/dwellingly-app/issues/522 as well. Do not merge yet.

tests/factory_fixtures/lease.py
```diff
@@ -21,11 +21,11 @@ def _lease_attributes(unitNum, tenant, property):
 
 @pytest.fixture
 def create_lease(faker, lease_attributes, create_property, create_tenant):
-    def _create_lease(unitNum=None):
-        if not unitNum:
-            unitNum = faker.building_number()
-        tenant = create_tenant()
-        lease = LeaseModel(**lease_attributes(unitNum, tenant, create_property()))
+    def _create_lease(tenant=None, property=None, unitNum=None):
```
aedwardg: Pytest has some built-in witchcraft that allows kwargs to be passed in that aren't in the fixture parameters. Discovered `tenant` and `property` only needed to be in the parameters of this inner function.

resources/user.py
```diff
@@ -44,17 +44,11 @@ def get(self, user_id):
...
-                )
-            )
+            properties_list = PropertyModel.find_by_manager(user_id)
+            user_info["properties"] = [p.json() for p in properties_list]
 
-            tenant_IDs = [tenant.id for sublist in tenant_list for tenant in sublist]
-            tenants_list = [TenantModel.find_by_id(t) for t in set(tenant_IDs)]
-            user_info["tenants"] = [t.json() for t in tenants_list if t]
+            tenants_json_2D = [p.tenants() for p in properties_list]
+            user_info["tenants"] = list(chain.from_iterable(tenants_json_2D))
```
aedwardg: Much cleaner solution. Props to @DaveTrost!

aedwardg (commented):
This looks ready to go, whenever you have the time to look over the newest changes @NickSchimek! 🎉

models/base_model.py
```diff
@@ -48,9 +48,14 @@ def update(cls, schema, id, payload):
...
-    def validate(schema, payload, partial=False):
+    def validate(schema, payload, context=None, partial=False):
         try:
-            return schema().load(payload, unknown=EXCLUDE, partial=partial)
+            if context:
+                return schema(context=context).load(
+                    payload, unknown=EXCLUDE, partial=partial
+                )
+            else:
+                return schema().load(payload, unknown=EXCLUDE, partial=partial)
```
NickSchimek: I wonder if we can get rid of this logic here and simply do this: 
`return schema(context=context).load(payload, unknown=EXCLUDE, partial=partial)`

resources/user.py
NickSchimek: Since this method is being modified, perhaps it can be refactored and cleaned up a bit too. Update it to use the `find` method instead of `find_by_id`.

Also this serialization logic doesn't belong here in the endpoint. Perhaps we can move it to the `json` method. That way we don't have to duplicate this wherever we want to serialize a property manager.

schemas/property.py
```diff
@@ -26,6 +26,8 @@ def validates_property_manager_ids(self, value):
 
     @validates("name")
     def validates_uniqueness_of_name(self, value):
+        if "name" in self.context and value == self.context["name"]:
+            return value
         if PropertyModel.find_by_name(value):
             raise ValidationError("A property with this name already exists")
```
NickSchimek: I see that we're passing context "PUT" everywhere, but I don't actually see it used anywhere. Is that context info necessary?

I find this logic a bit confusing, although after some thought I realize that we are only returning early when the name has been passed in which only occurs on the PUT action. So it seems we can remove passing in the `PUT` context. I think if we do something like this it might be a bit more readable.
```suggestion
        if updating_name and PropertyModel.find_by_name(value):
            raise ValidationError("A property with this name already exists")
```

and move the necessary log to a private method for `updating_name`

tests/factory_fixtures/lease.py
NickSchimek: Thanks for cleaning this up 👍

tests/integration/test_properties.py
```diff
@@ -215,15 +216,20 @@ def test_delete_property_by_id(client, auth_headers, test_database):
...
+
+    property_json["address"] = new_property_address
+
+    response = client.put(
+        f"/api/properties/{test_property.id}",
+        json=property_json,
+        headers=valid_header,
     )
-    test_changed_property = json.loads(test_changed_property.data)
+    assert is_valid(response, 200)
```
NickSchimek: This seems good and all, but I don't think is_valid is testing anything useful, and it re-tests what the framework already tests. Testing the status code is great, but testing the headers are correct doesn't need to be done everywhere as that is done automatically by the tools we're using, and does not need to be retested in every endpoint test, especially since we're always returning json.

tests/integration/test_properties.py
```diff
@@ -215,15 +216,20 @@ def test_delete_property_by_id(client, auth_headers, test_database):
     assert response == 404
 
 
-def test_update_property_by_id(client, auth_headers, create_property, test_database):
+def test_update_property_by_id(client, empty_test_db, valid_header, create_property):
```
NickSchimek: I think this method can be renamed to `test_update`. The `by id` part is a bit misleading and confusing.

NickSchimek: Thanks for updating to use the new faster factories! 🎉

tests/integration/test_users.py
```diff
@@ -81,21 +81,52 @@ def test_refresh_user(client, test_database, admin_user):
     assert responseRefreshToken.status_code == 200
 
 
-def test_get_user_by_id(client, auth_headers, admin_user):
+def test_get_user_by_id(client, empty_test_db, create_admin_user, admin_header):
```
NickSchimek: same here.

tests/integration/test_users.py
```diff
@@ -81,21 +81,52 @@ def test_refresh_user(client, test_database, admin_user):
...
     assert responseBadUserId.json == {"message": "User not found"}
 
 
+def test_get_pm_by_id(
+    client,
+    empty_test_db,
+    create_property_manager,
+    create_property,
+    create_lease,
+    admin_header,
```
NickSchimek: I think I would prefer to use valid_header here. And build separate authorization tests that use admin_header elsewhere. That way we're not testing authorization here, and can focus on what the method does.

tests/integration/test_users.py
```diff
@@ -81,21 +81,52 @@ def test_refresh_user(client, test_database, admin_user):
...
+    lease_1 = create_lease(property=prop)
+    lease_2 = create_lease(property=prop)
+
+    response = client.get(f"/api/user/{user.id}", headers=admin_header)
+
+    property_list = response.json["properties"]
+    tenants_list = response.json["tenants"]
+
+    """The get user by id route returns a successful response code for PMs"""
+    assert response.status_code == 200
```
NickSchimek: I don't understand this comment. The admin_header made the request so shouldn't it be a successful response for admins? Am I missing something here?

This looks like an authorization test. Perhaps we can put this with the authorization tests. If a folder for authorization tests don't exist yet, perhaps now would be a good time to create one.

tests/integration/test_users.py
```diff
@@ -81,21 +81,52 @@ def test_refresh_user(client, test_database, admin_user):
...
+
+    """The PM's properties are returned as a list of JSON objects"""
+    assert property_list == [prop.json()]
+
+    """
+    Tenants are retreived through the leases on each
+    property and returned as a list of JSON objects
+    """
+    assert tenants_list == [lease_1.tenant.json(), lease_2.tenant.json()]
```
NickSchimek: This is good! I like it. I'd probably want to expand on this to see the whole structure of the response. Depending on if we have tests for the json methods we can do something more or less like.

`assert request.json == user.json()`

Of course that would mean that the serialization in the endpoint needs to be moved to the json method, and that json method should be unit tested.

IF (or if not) the serialization remains in the endpoint then I would like to see the other values asserted as well.

something like this:

```
assert request.json == {
  "name": user.name
  "phone": user.phone
  ...
  "properties": [prop.json()]
  "tenants": [lease_1.tenant.json(), lease_2.tenant.json()]
}
```

tests/schemas/test_property_schema.py
```diff
@@ -62,16 +62,50 @@ def test_manager_update(self, create_property, create_property_manager):
...
 
-        assert PropertyModel.update(schema=PropertySchema, id=prop.id, payload=payload)
+        assert PropertyModel.update(
+            schema=PropertySchema, id=prop.id, payload=payload, context=context
+        )
+
+
+@pytest.mark.usefixtures("empty_test_db")
+class TestUpdatePropertyValidations:
+    def test_property_can_update_with_same_name(self, create_property):
```
NickSchimek: These look good. I don't think they need to be in a separate class though. Let's move them to the validation section of the class `TestPropertyManagerValidations` -> and while your in there rename that class name to `TestPropertyValidations`

NickSchimek (changes requested):
Thanks @aedwardg - I think this look good. I made some comments as you've already saw :-) Can this be split into two separate PR's? This PR is a bit big and it is solving what seems to be two different issues.

tests/integration/test_users.py
aedwardg: @NickSchimek, sorry this was a confusing comment.  

I meant that it returned a successful response code when the user being queried was a PM. It's just testing that the endpoint now works when querying a PM user, since a previous PR broke that functionality (the endpoint was successful when querying admin, staff, etc, just not PMs).

tests/integration/test_users.py
aedwardg: I agree.  I'll make an issue to make an auth tests directory and make/move auth tests over there.

tests/integration/test_users.py
aedwardg: You bring up a great point here, and in your comment about having the PM's `tenants` serialization moved into the json method.  

If you're okay with it, I'd like to address both of these in a separate issue/PR, just so we get something that works merged quickly since right now that functionality is broken in the frontend.

schemas/property.py
aedwardg: I changed this so that it now uses a private method called `_assigning_name()`.  Since this validation should could return an error from a property name update OR the creation for a new property, I thought that name might be more appropriate.  Let me know what you think of this new implementation.

aedwardg:
@NickSchimek Thanks for the great feedback (as always)! 🙏 

I made the changes you suggested, so let me know if there's any more fine-tuning you think I should do.

resources/user.py
NickSchimek: This can be removed now right?

tests/factory_fixtures/property.py
```diff
@@ -23,10 +23,10 @@ def _property_attributes(archived=False, manager_ids=None):
 
 
 @pytest.fixture
-def create_property(property_attributes):
-    def _create_property():
+def create_property(property_attributes, **kwargs):
```
NickSchimek: I think I saw this in the other pr. Does it belong here?

tests/integration/test_users.py
NickSchimek: Thanks. Since this is testing the bug. I think this should be in a separate test function that is well named so that it doesn't get removed accidentally by someone in the future not knowing what it's here for.

tests/integration/test_users.py
NickSchimek: Ok, but this shouldn't be holding up the frontend work.

NickSchimek (commented):
It looks like the changes that were moved to the other PR are still in this PR. Can we get this cleaned up and resubmitted. Ideally when submitting a bug fix I would only like to see changes that fix the bug. 
One change to write a test to expose the bug.
A second change to fix the bug.
And any necessary refactoring prior to and/or post fixing the bug.

Anything else makes it harder to review and can make the process last longer.

tests/integration/test_users.py
NickSchimek: I'm fine with doing this in a separate or though

schemas/property.py
```diff
@@ -26,7 +26,12 @@ def validates_property_manager_ids(self, value):
...
     @validates("name")
     def validates_uniqueness_of_name(self, value):
-        if PropertyModel.find_by_name(value):
+        def _assigning_name():
+            if "name" in self.context and value == self.context["name"]:
+                return False
+            return True
+
+        if _assigning_name() and PropertyModel.find_by_name(value):
             raise ValidationError("A property with this name already exists")
```
NickSchimek, line 35: I know sqlalchemy isn't that friendly when it comes to building queries. I think providing the context here is a good solution, and might be the best option. Another option would be to build a query that checks if the name exists in the database for any row that is not this property. Without looking at the documentation I'm not to sure how to do that, but I know it can be done.

Although from what you mentioned it seems we would still need to pass in a context, which is too bad. So this is probably the best solution.

aedwardg:
Thanks @NickSchimek, should be all cleaned up now!

tests/integration/test_properties.py
```diff
@@ -215,15 +215,20 @@ def test_delete_property_by_id(client, auth_headers, test_database):
...
+        f"/api/properties/{test_property.id}",
+        json=property_json,
+        headers=valid_header,
     )
-    test_changed_property = json.loads(test_changed_property.data)
+    assert response == 200
 
-    assert test_changed_property["address"] == new_property_address
+    """The property should have a new address"""
+    assert response.json["address"] == new_property_address
```
NickSchimek: This test is kind of ok. A better test would be to confirm we send the correct message to update the resource.

The test is ok, but can you think of anyways the test would still pass when the behavior is broken?

NickSchimek: Also if testing the json response we should test the entire response, although I can't tell if we're doing that somewhere else as I'm on my phone.

tests/schemas/test_property_schema.py
```diff
@@ -41,6 +41,31 @@ def test_uniqueness_of_property_name(self, create_property):
...
 
+    def test_property_can_update_with_same_name(self, create_property):
+        new_property = create_property()
+        payload = new_property.json()
+
+        payload["num_units"] = 42
+
+        context = {"name": new_property.name}
+
+        assert PropertyModel.update(
```
NickSchimek: I don't like hitting the database in these unit tests. We can accomplish the same tests with the validate method. There should be no need to hit the database in validation tests.

Other than that these tests look good. I think I would like one more for the sad path using the context. Since I don't think that line of code is being executed yet.

NickSchimek (changes requested):
Looks good. A couple comments. The only thing I would like to see changed is using the validate method in the validation tests, because these are unit tests and there is no need to actually update or create objects.

tests/schemas/test_property_schema.py
aedwardg: Thanks for recommending a sad path test w/the context.  It actually exposed a bug in how I was assigning the context.  Updating now.

aedwardg: @NickSchimek did you need me to also update these PostLoadDeserialization tests to use `validate()`, or do those need to hit the database?
https://github.com/codeforpdx/dwellinglybackend/blob/b9457c3c0b715cfe5dca4d6131a3fcfef1ed65e4/tests/schemas/test_property_schema.py#L70-L108

NickSchimek: Sure, They should be using `PropertySchema().load`.  Did I write these? I can't remember. I don't think anything in the schema test files should hit a method that is not part of the schema interface. `(validate, load, dump)`. There is actually very good reasoning behind this.

aedwardg: Yes, you did write these 😆
I've updated all the tests in this file so that they use either `PropertySchema().load()` or `PropertySchema().validate()` now, so hopefully they are good to go 🤞

tests/schemas/test_property_schema.py
```diff
@@ -89,12 +86,9 @@ def test_manager_update(self, create_property, create_property_manager):
         payload = {"propertyManagerIDs": [pm_2.id, pm_3.id]}
         context = {"name": prop.name}
 
-        PropertyModel.update(
-            schema=PropertySchema, id=prop.id, payload=payload, context=context
-        )
-        db.session.rollback()
+        updated_prop = PropertySchema(context=context).load(payload, partial=True)
```
NickSchimek: excellent!

tests/integration/test_properties.py
```diff
@@ -215,15 +215,39 @@ def test_delete_property_by_id(client, auth_headers, test_database):
     assert response == 404
 
 
-def test_update_property_by_id(client, auth_headers, create_property, test_database):
+def test_update_property(client, empty_test_db, valid_header, create_property):
```
NickSchimek, line 218: These tests seem like they belong in the Unit tests for the schemas since they are testing the validation on name.

NickSchimek: oh it looks like there are unit tests too, so I don't think we need these integration tests.

NickSchimek (commented):
Just one comment, but I think this is probably good to merge, what do you think?

NickSchimek (approved):

PR #210, opened 2021-02-11
Add tickets relationship to TenantModel

### LINKED ISSUE #498: Add tickets relationship to the TenantModel
**Purpose**
A tenant can have many tickets. Add the tickets relationship to the TenantModel so that we can access tickets through a tenant object.

**Proposed Solution**
something like:
`tickets = db.relationship(...)`

cecillesalazar:

Closes codeforpdx/dwellingly-app/issues/498

See commit messages and comment on b0206ec. Can't figure out why we StaffTenantLink and TenantModel needed to be globally imported in app.py with the `# noqa: F401` (unused imported module error) in order for the one liner in models/tenant.py to compile and not spit out a sqlalchemy invalid request error.

This PR accomplishes the objective - but kicks the can down the road on the import issue. 

Nope

Nope

Nope

### DIFF
--- a/app.py
+++ b/app.py
@@ -8,6 +8,7 @@ from models.user import UserModel
 # This should not be imported here, but for now we can force Flake8 to ignore the error
 # The "noqa" comment can be deleted when we figure out how to get around this.
 from models.staff_tenant_link import StaffTenantLink  # noqa: F401
+from models.tickets import TicketModel  # noqa: F401
 from models.revoked_tokens import RevokedTokensModel
 from resources.user import (
     UserRegister,
--- a/app.py
+++ b/app.py
@@ -8,7 +8,6 @@ from models.user import UserModel
 # This should not be imported here, but for now we can force Flake8 to ignore the error
 # The "noqa" comment can be deleted when we figure out how to get around this.
 from models.staff_tenant_link import StaffTenantLink  # noqa: F401
-from models.tickets import TicketModel  # noqa: F401
 from models.revoked_tokens import RevokedTokensModel
 from resources.user import (
     UserRegister,
--- a/models/tenant.py
+++ b/models/tenant.py
@@ -1,7 +1,6 @@
 from sqlalchemy.orm import relationship
 from db import db
 from models.base_model import BaseModel
-from models.tickets import TicketModel
 from utils.time import Time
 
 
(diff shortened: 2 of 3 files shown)


app.py
```diff
@@ -8,6 +8,7 @@
 # This should not be imported here, but for now we can force Flake8 to ignore the error
 # The "noqa" comment can be deleted when we figure out how to get around this.
 from models.staff_tenant_link import StaffTenantLink  # noqa: F401
+from models.tickets import TicketModel  # noqa: F401
```
aedwardg: What do you think about moving these into the `models` directory's `__init__.py` file instead?

aedwardg (commented):
@cecillesalazar, sorry this issue ended up being such a hassle! 

I was playing around with it and it looks like everything still functions properly if we move the `TicketModel` and `StaffTenantLink` imports from `app.py` and put them in the models `__init__.py` instead.

What do you think about doing it that way? I feel like it's probably a little better for two reasons:

1. We're not actually using either of these in `app.py` and then we can get rid of all those comments as well (both the comments on line 8-9 and the `# noqa: F401` comments)
2. We're already utilizing another `__init__.py` for importing the schemas, so it's more in line with our current setup.

It's still not ideal, but I think it's better than an app-level import.

NickSchimek (approved):
This looks good. I think it is exactly what we want.  Thanks @cecillesalazar

app.py
NickSchimek: I don't have a deep understanding of how Flask builds the app so I can't answer this question or even why we would need to import TicketModel in the app.py file now. I think further discussion/investigation may be warranted to see why this is needing to be done in the first place before making any decisions on what to do.

aedwardg: @NickSchimek It was a temporary band-aid to get around the circular import problem.  `models/tickets.py` imports `TenantModel`, so `models/tenant.py` can't import `TicketModel`.  But `TicketModel` needed to be imported somewhere in order to create the relationship.

NickSchimek: ah I see. Thank you. The correct way to fix this is that we should not be importing the `TenantModel` in the `TicketModel`

NickSchimek: So. Let's remove this line, and update the ticket model so that it is not importing the tenant model.

NickSchimek (changes requested):
OK. Here is what we need to do to resolve the issues that were seen. 

Let's remove the added line in `app.py` and remove the `Tenant` import statement in the `Ticket` model. 

We'll then need to update the `json` method to not depend on the `Ticket` model.

Something like this should work since we now have the `backref` on the relationship you defined on the `tenant` model.

Replace:
```
        tenantData = TenantModel.find_by_id(self.tenantID)
        tenantName = "{} {}".format(tenantData.firstName, tenantData.lastName)
```

with some like:
```
        tenantData = self.tenant
        tenantName = "{} {}".format(tenantData.firstName, tenantData.lastName)
```

Or depending on the style I would just remove those two lines and do this in the return statement:
```
            "tenant": "{} {}".format(self.tenant.firstName, self.tenant.lastName),
```

Or preferably define a full_name method on the tenant and do:
```
            "tenant": self.tenant.full_name,
```

However, it may be best to just serialize all the tenant data here instead. I don't know though. We can check with @Chris-Boe or @DaveTrost on this.
```
           "tenant": self.tenant.json(),
```

aedwardg:
Assuming that after @NickSchimek's suggested changes `TicketModel` is imported inside of `models/tenant.py`, you may need to change this line
```python
tickets = db.relationship("TicketModel", backref="tenant", lazy=True)
```
to not use a string:
```python
tickets = db.relationship(TicketModel, backref="tenant", lazy=True)
```

Otherwise, Flake8 will likely consider it an unused import.  We had to do this two months ago for the managers relationship in the `PropertyModel`.

cecillesalazar:
@aedwardg @NickSchimek - thanks for the feedback - I'll get on this sometime tmrw - losing power's set me back a few days 😿

NickSchimek:
No worries - Thanks @cecillesalazar !

cecillesalazar:
@aedwardg, @NickSchimek - I think this should do it, let me know if I missed/misunderstood something 👍

aedwardg (approved):
Looks good to me @cecillesalazar! Thanks again!

NickSchimek (approved):
This looks great! Thanks Cecile! I think this is about the only time I don't require tests.

PR #211, opened 2021-02-11
Tenant json includes only the active lease

### LINKED ISSUE #512: Resource `tenants/<int:tenant_id>` should only include current/active lease info
## Purpose
When retrieving a tenant's JSON information, the lease provided to the front end should only be the currently active lease. Right now, there is a many-to-one relationship (tenants are able to have multiple leases). It makes sense to keep it this way if we want to retain previous leases, but the endpoint should only return the current lease unless specifically asked for all the leases.

This was discussed to some extent [here](https://github.com/codeforpdx/dwellingly-app/pull/509).

It looks like most of the groundwork for this was completed in [backend PR 195](https://github.com/codeforpdx/dwellinglybackend/pull/195).  **However**, [this line](https://github.com/codeforpdx/dwellinglybackend/blob/3246d1d7b40620e147d17768107242f91a9914d5/models/tenant.py#L28) only returns the first lease in the list.  If a newer lease gets added for this tenant, there is no guarantee it will be at index 0, so we should instead be filtering the tenant's leases to only the current one.

**NOTE: this is assuming one tenant cannot be on two active leases at once. If we later decide this should be possible (e.g., if a family needs two units and the head of household is on both leases) we'll need to re-think this.**

## Proposed Solution
Filter the tenant's leases and return the lease where datetime.now() falls between the lease start and end dates*
*_Note the assumption above_

## Acceptance Criteria

- [ ] Include appropriate tests (when in doubt, consult the [Project Components doc](https://github.com/codeforpdx/dwellinglybackend/blob/development/CONTRIBUTING.md#project-components) for where to test each component)
- [ ] If a tenant has multiple leases, _only_ the currently active one should be returned

## Mockup Screenshot & Page Number
This will provide the correct property and unit information for the frontend's `TenantView`, as seen on [Figma pg. 10](https://drive.google.com/file/d/1YqboQogczYm1HkyRqEtVSzeQ61T9hWU2/view):

image

gussmith23:

Closes codeforpdx/dwellingly-app/issues/512

Previously, `TenantModel.json()` filled out the `'lease'` field by getting the first lease returned. Now, we filter the list of leases using the new `TenantModel.active()` class method, which returns a SQLalchemy filter filtering out inactive leases. Currently, we just return the first of those active leases.

### DIFF
--- a/models/lease.py
+++ b/models/lease.py
@@ -1,3 +1,4 @@
+from sqlalchemy.sql.functions import now
 from db import db
 from models.base_model import BaseModel
 from utils.time import Time
@@ -24,3 +25,8 @@ class LeaseModel(BaseModel):
             "dateTimeEnd": Time.format_date(self.dateTimeEnd),
             "unitNum": self.unitNum,
         }
+
+    @classmethod
+    def active(cls):
+        time = now()
+        return (time > cls.dateTimeStart) & (time < cls.dateTimeEnd)
--- a/models/tenant.py
+++ b/models/tenant.py
@@ -2,6 +2,7 @@ from sqlalchemy.orm import relationship
 from db import db
 from models.base_model import BaseModel
 from utils.time import Time
+from models.lease import LeaseModel
 
 
 class TenantModel(BaseModel):
@@ -16,17 +17,19 @@ class TenantModel(BaseModel):
     # relationships
     staff = relationship("UserModel", secondary="staff_tenant_links")
     leases = db.relationship(
-        "LeaseModel", backref="tenant", lazy=True, cascade="all, delete-orphan"
+        "LeaseModel", backref="tenant", lazy="dynamic", cascade="all, delete-orphan"
     )
 
     def json(self):
+        first_active_lease = self.leases.filter(LeaseModel.active()).first()
+        active_lease_json = first_active_lease.json() if first_active_lease else ""
         return {
             "id": self.id,
             "firstName": self.firstName,
             "lastName": self.lastName,
             "fullName": "{} {}".format(self.firstName, self.lastName),
             "phone": self.phone,
-            "lease": self.leases[0].json() if self.leases else "",
+            "lease": active_lease_json,
             "staff": [user.json() for user in self.staff] if self.staff else [],
             "created_at": Time.format_date(self.created_at),
             "updated_at": Time.format_date(self.updated_at),
--- a/tests/factory_fixtures/lease.py
+++ b/tests/factory_fixtures/lease.py
@@ -4,15 +4,13 @@ from models.lease import LeaseModel
 
 @pytest.fixture
 def lease_attributes(faker):
-    def _lease_attributes(unitNum, tenant, property):
+    def _lease_attributes(unitNum, tenant, property, dateTimeStart, dateTimeEnd):
         return {
             "unitNum": unitNum,
             "tenantID": tenant.id,
             "propertyID": property.id,
-            "dateTimeStart": faker.date_time_this_decade(),
-            "dateTimeEnd": faker.date_time_this_decade(
-                before_now=False, after_now=True
-            ),
+            "dateTimeStart": dateTimeStart,
+            "dateTimeEnd": dateTimeEnd,
             "occupants": faker.random_number(digits=2),
         }
 
@@ -21,11 +19,23 @@ def lease_attributes(faker):
 
 @pytest.fixture
 def create_lease(faker, lease_attributes, create_property, create_tenant):
-    def _create_lease(tenant=None, property=None, unitNum=None):
+    def _create_lease(
+        tenant=None,
+        property=None,
+        unitNum=None,
+        dateTimeStart=None,
+        dateTimeEnd=None,
+    ):
         unitNum = unitNum or faker.building_number()
         tenant = tenant or create_tenant()
         property = property or create_property()
-        lease = LeaseModel(**lease_attributes(unitNum, tenant, property))
+        dateTimeStart = dateTimeStart or faker.date_time_this_decade()
+        dateTimeEnd = dateTimeEnd or faker.date_time_this_decade(
+            before_now=False, after_now=True
+        )
+        lease = LeaseModel(
+            **lease_attributes(unitNum, tenant, property, dateTimeStart, dateTimeEnd)
+        )
         lease.save_to_db()
         return lease
 
--- a/tests/factory_fixtures/lease.py
+++ b/tests/factory_fixtures/lease.py
@@ -1,5 +1,7 @@
 import pytest
 from models.lease import LeaseModel
+from schemas.lease import LeaseSchema
+from utils.time import Time
 
 
 @pytest.fixture
@@ -9,8 +11,8 @@ def lease_attributes(faker):
             "unitNum": unitNum,
             "tenantID": tenant.id,
             "propertyID": property.id,
-            "dateTimeStart": dateTimeStart,
-            "dateTimeEnd": dateTimeEnd,
+            "dateTimeStart": Time.to_iso(dateTimeStart),
+            "dateTimeEnd": Time.to_iso(dateTimeEnd),
             "occupants": faker.random_number(digits=2),
         }
 
@@ -33,10 +35,10 @@ def create_lease(faker, lease_attributes, create_property, create_tenant):
         dateTimeEnd = dateTimeEnd or faker.date_time_this_decade(
             before_now=False, after_now=True
         )
-        lease = LeaseModel(
-            **lease_attributes(unitNum, tenant, property, dateTimeStart, dateTimeEnd)
+        lease = LeaseModel.create(
+            LeaseSchema,
+            lease_attributes(unitNum, tenant, property, dateTimeStart, dateTimeEnd),
         )
-        lease.save_to_db()
         return lease
 
     yield _create_lease
--- a/tests/unit/test_tenant.py
+++ b/tests/unit/test_tenant.py
@@ -1,7 +1,9 @@
+from models.property import PropertyModel
 import pytest
 from tests.unit.base_interface_test import BaseInterfaceTest
 from models.tenant import TenantModel
 from schemas.tenant import TenantSchema
+import datetime
 
 
 class TestBaseTenantModel(BaseInterfaceTest):
@@ -9,6 +11,28 @@ class TestBaseTenantModel(BaseInterfaceTest):
         self.object = TenantModel()
         self.custom_404_msg = "Tenant not found"
         self.schema = TenantSchema
+
+    def test_only_active_lease_returned(self, faker, create_lease, create_tenant):
+        tenant = create_tenant()
+
+        # Create an expired lease for the tenant
+        lease_expired_end = faker.date_time_this_decade(
+            before_now=True, after_now=False
+        )
+        lease_expired_start = lease_expired_end - datetime.timedelta(days=365)
+        lease_expired = create_lease(
+            tenant=tenant,
+            dateTimeStart=lease_expired_start,
+            dateTimeEnd=lease_expired_end,
+        )
+
+        # Create an active lease for the tenant
+        lease_active = create_lease(
+            tenant=tenant, property=PropertyModel.find_by_id(lease_expired.propertyID)
+        )
+
+        # Active lease should be the only one that shows up
+        assert tenant.json()["lease"] == lease_active.json()
 
 
 @pytest.mark.usefixtures("empty_test_db")
--- a/tests/unit/test_tenant.py
+++ b/tests/unit/test_tenant.py
@@ -28,7 +28,7 @@ class TestBaseTenantModel(BaseInterfaceTest):
 
         # Create an active lease for the tenant
         lease_active = create_lease(
-            tenant=tenant, property=PropertyModel.find_by_id(lease_expired.propertyID)
+            tenant=tenant, property=PropertyModel.find(lease_expired.propertyID)
         )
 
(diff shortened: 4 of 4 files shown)


gussmith23:
I still need to write tests for this. Perhaps unsettlingly (or maybe this is actually a good thing!) no tests failed once I made this change!

aedwardg:
> 
> 
> I still need to write tests for this. Perhaps unsettlingly (or maybe this is actually a good thing!) no tests failed once I made this change!

I don't think we've built anything in yet that tested for multiple leases, so that's expected behavior 😂

gussmith23:
Alright, I added a unit test, but I had to mess with fixtures a bit. Let me know if the changes are too intrusive.

In the spirit of test-driven development, I reordered my commits so that the test is added in the first commit. If you want to check that it fails, check out that commit and run the tests.

models/tenant.py
```diff
@@ -16,17 +17,24 @@ class TenantModel(BaseModel):
...
+        first_active_lease = self.leases.filter(LeaseModel.active()).first()
+        active_lease_json = first_active_lease.json() if first_active_lease else ""
         return {
             "id": self.id,
             "firstName": self.firstName,
             "lastName": self.lastName,
             "fullName": "{} {}".format(self.firstName, self.lastName),
             "phone": self.phone,
-            "lease": self.leases[0].json() if self.leases else "",
+            "lease": active_lease_json,
```
aedwardg: I like this 👍 
I didn't know about `lazy="dynamic" before, so that's neat.

models/lease.py
```diff
@@ -24,3 +25,8 @@ def json(self):
             "dateTimeEnd": Time.format_date(self.dateTimeEnd),
             "unitNum": self.unitNum,
         }
+
+    @classmethod
+    def active(cls):
+        time = now()
+        return (time > cls.dateTimeStart) & (time < cls.dateTimeEnd)
```
aedwardg, line 32: I don't think we need the bitwise `&` operator here, since we're comparing two booleans.  I think the logical `and` is preferable unless you had a very specific reason for `&`.

Also, how sure are we that these comparisons are between Datetime objects and not strings? 
I just want to make sure, because if they are strings, we'll need to convert them to Datetime objects to do a proper comparison.

tests/factory_fixtures/lease.py
```diff
@@ -21,11 +19,28 @@ def _lease_attributes(unitNum, tenant, property):
...
+            dateTimeEnd = faker.date_time_this_decade(before_now=False, after_now=True)
+        if not tenantId:
+            tenantId = create_tenant().id
+        if not propertyId:
+            propertyId = create_property().id
+        lease = LeaseModel(
+            **lease_attributes(
+                unitNum, tenantId, propertyId, dateTimeStart, dateTimeEnd
+            )
+        )
```
aedwardg: I actually had to go in and alter these fixtures in my own PR last week, and it looks like we had similar ideas 😄 

Here's how I did it in that PR:
https://github.com/codeforpdx/dwellinglybackend/blob/0dbcf73acf7fdd40475e468259e45d61928e3114/tests/factory_fixtures/lease.py#L6-L32

Basically, the main difference is that in mine I passed in the actual tenant and property rather than the ids, and used short-circuit evaluation in `_create_lease()` to make it more concise.  I think it looks a little better without all the `if not` blocks, but if you think they make the code more understandable, I'm open to hearing the reasoning.

I wouldn't bring this up, except I think this will probably result in some type of merge conflict, so it's probably best if we match them.  

Would you be open to using my style, with your added datetime parameters?

aedwardg (changes requested):
@gussmith23 awesome job, thanks for taking care of this!

I really like how you handled the lease filtering and the test.

I left a couple of comments with some suggested changes, let me know what you think!

models/tenant.py
NickSchimek: Nice! excellent work! 🎉

models/tenant.py
```diff
@@ -16,17 +17,24 @@ class TenantModel(BaseModel):
     # relationships
     staff = relationship("UserModel", secondary="staff_tenant_links")
     leases = db.relationship(
-        "LeaseModel", backref="tenant", lazy=True, cascade="all, delete-orphan"
+        "LeaseModel",
+        backref="tenant",
+        # lazy="dynamic" is required so that we can filter() over leases in
+        # json(). See: https://stackoverflow.com/questions/11578070
+        lazy="dynamic",
```
NickSchimek: I tend to not like comments in code. I think this comment would be good to put in the commit that added this, then if anyone has any question as to what this does they can either look it up in the documentation or review the commit history and see what the contributor wrote in the commit log as to why this was added.

models/tenant.py
```diff
@@ -16,17 +17,24 @@ class TenantModel(BaseModel):
...
+        cascade="all, delete-orphan",
     )
 
     def json(self):
+        first_active_lease = self.leases.filter(LeaseModel.active()).first()
+        active_lease_json = first_active_lease.json() if first_active_lease else ""
```
NickSchimek, line 25: It seems that we want to eagerly load this instead of lazily. Perhaps it would be more correct to eagerly load a custom filtered collection result.
https://docs.sqlalchemy.org/en/14/orm/loading_relationships.html#using-contains-eager-to-load-a-custom-filtered-collection-result

I'm not sure if this would work, but it sounds like this might be what we want. I'm not opposed to the solution you provided either, just providing an alternate method that has the possibility to only hit the database once, but this might be more work and might not work out of the box with the `User.find` method.

NickSchimek: I'm curious if removing the lazy argument works. Then we should be able to use filter on the returned data. Since it might default to eagerly loading. I'm not sure though, I'm not that well read up on sqlalchemy. I just don't think we should lazily load the data when we know we will be using it all the time.

models/lease.py
gussmith23: Re: the bitwise &: I originally tried `and`. It turns out that won't work, because SQLalchemy can't override python's `and`! The & being used here is not actually Python's bitwise and -- it's SQLalchemy's overridden "and" operator. I could also change this to use the equivalent, explicit `and` function that exists somewhere in SQLalchemy.

Re: strings vs datetimes: I actually don't know 😶 I will have to track down the exact datatype of the column involved. I generally work on Dwellingly work on Wednesdays, so I'll do that then.

tests/factory_fixtures/lease.py
gussmith23: Yes, it's no problem with me! The only reason I used ID was because the ID was currently the only field used, so I figured I'd just simplify things. I will have to confirm that this will also work, but I'm pretty sure it will.

models/lease.py
aedwardg: Huh! I had no idea SQLalchemy did that! In that case, I think it's totally fine to use `&` here.
You learn something new every day...

tests/factory_fixtures/lease.py
gussmith23: Done, using your code now. Still changed the fixtures a bit.

models/tenant.py
gussmith23: Done; moved to commit message.

models/lease.py
NickSchimek: I had no idea either. Thanks for explaining this to us @gussmith23

models/tenant.py
```diff
@@ -16,17 +17,19 @@ class TenantModel(BaseModel):
     # relationships
     staff = relationship("UserModel", secondary="staff_tenant_links")
     leases = db.relationship(
-        "LeaseModel", backref="tenant", lazy=True, cascade="all, delete-orphan"
+        "LeaseModel", backref="tenant", lazy="dynamic", cascade="all, delete-orphan"
     )
 
     def json(self):
+        first_active_lease = self.leases.filter(LeaseModel.active()).first()
+        active_lease_json = first_active_lease.json() if first_active_lease else ""
```
NickSchimek, line 25: I think it might be better to return None instead of an empty string when there isn't an active lease, but it depends on what the frontend wants.

tests/unit/test_tenant.py
```diff
@@ -10,6 +12,28 @@ def setup(self):
...
+        lease_expired_start = lease_expired_end - datetime.timedelta(days=365)
+        lease_expired = create_lease(
+            tenant=tenant,
+            dateTimeStart=lease_expired_start,
+            dateTimeEnd=lease_expired_end,
+        )
+
+        # Create an active lease for the tenant
+        lease_active = create_lease(
+            tenant=tenant, property=PropertyModel.find_by_id(lease_expired.propertyID)
```
NickSchimek: FYI: We shouldn't be using the `find_by_id` method. I wonder if there is a way to put a deprecation warning for whenever a developer users that method :-) All new code should be using the `find` method as I would like to remove the `find_by_id` method some day.

tests/unit/test_tenant.py
```diff
@@ -10,6 +12,28 @@ def setup(self):
...
+        # Create an expired lease for the tenant
+        lease_expired_end = faker.date_time_this_decade(
+            before_now=True, after_now=False
+        )
+        lease_expired_start = lease_expired_end - datetime.timedelta(days=365)
+        lease_expired = create_lease(
+            tenant=tenant,
+            dateTimeStart=lease_expired_start,
+            dateTimeEnd=lease_expired_end,
+        )
```
NickSchimek, line 27: Maybe some day we can extract this out to a `create_expired_lease` factory, but if this is the only place we need to do this then perhaps it's not necessary.

tests/factory_fixtures/lease.py
```diff
@@ -21,11 +19,23 @@ def _lease_attributes(unitNum, tenant, property):
...
         property = property or create_property()
-        lease = LeaseModel(**lease_attributes(unitNum, tenant, property))
+        dateTimeStart = dateTimeStart or faker.date_time_this_decade()
+        dateTimeEnd = dateTimeEnd or faker.date_time_this_decade(
+            before_now=False, after_now=True
+        )
+        lease = LeaseModel(
+            **lease_attributes(unitNum, tenant, property, dateTimeStart, dateTimeEnd)
+        )
         lease.save_to_db()
```
NickSchimek: We should probably switch this to use the create method. Just something to be aware of. Whenever we modify old code or create new code.. it should use the `create` method so that we can ensure our objects are valid.

NickSchimek (dismissed):
Thanks @gussmith23 - You did an excellent job. Thank you 🙏

aedwardg (dismissed):
Great job, thanks @gussmith23!

gussmith23:
This isn't done -- don't merge quite yet :) I will finish it up on Wednesday.

NickSchimek:
Ok! Request review when ready

models/tenant.py
gussmith23: Removing lazy argument does not work. Errors look like this:
```
FAILED tests/integration/test_properties.py::test_get_property_by_id - AttributeError: 'InstrumentedList' object has no attribute 'filter'
```
I'm honestly still a little confused by the docs; seems like there are other options other than `dynamic`, and I'm unsure if `dynamic` is the only option that will work here. Maybe there's something better, but maybe those things would require us to restructure other things. For now, this works---if there's something especially egregious about this, someone should call it out, but otherwise we'll let this lie!

tests/unit/test_tenant.py
gussmith23: Done!

gussmith23: Done!

tests/factory_fixtures/lease.py
gussmith23:
```suggestion
        lease = LeaseModel.create(
            LeaseSchema,
            lease_attributes(unitNum, tenant, property, dateTimeStart, dateTimeEnd),
        )
```

When I do this, I get the following error:
```
E       werkzeug.exceptions.BadRequest: 400 Bad Request: {'dateTimeEnd': ['Not a valid datetime.'], 'dateTimeStart': ['Not a valid datetime.']}
```

gussmith23: What those objects look like:
```
> /Users/gus/dwellinglybackend/tests/factory_fixtures/lease.py(9)_lease_attributes()->{'dateTimeEnd': datetime.date...3, 13, 54, 25), 'dateTimeStart': datetime.date...0, 20, 40, 19), 'occupants': 47, 'propertyID': 1, ...}
```

tests/unit/test_tenant.py
```diff
@@ -10,6 +12,28 @@ def setup(self):
...
+
+        # Create an active lease for the tenant
+        lease_active = create_lease(
+            tenant=tenant, property=PropertyModel.find(lease_expired.propertyID)
+        )
+
+        # Active lease should be the only one that shows up
+        assert tenant.json()["lease"] == lease_active.json()
```
NickSchimek, line 35: This is testing the json method, but I would have expected it to test the active method.

NickSchimek: I would expect a separate json test to test the entirety of the json method.

NickSchimek: so when we test the active method here. We can test the happy and sad path. When we test the json method then we can test the happy path.

NickSchimek (approved):
These changes look good. The only issue I have is that we are testing a method on the lease model through a tenant unit test. I would've expected testing to be done in the `tests/unit/test_lease.py` file. And one unit test in the tenant file to test the entirety of the json method instead of a slice of it.

Let me know if this makes sense, I'd be happy to explain more in detail about these nuances. Either way I think these code changes are great.

Thanks for doing this @gussmith23

PR #212, opened 2021-02-15
Fix bug for retrieving PMs and their tenants through the Users endpont

### LINKED ISSUE #522: Cannot GET property manager through users/<id> endpoint
**Describe the bug**
Trying to get user info for Property Managers through the `users/<id>` endpoint fails.
The issue stems from [L46-55](https://github.com/codeforpdx/dwellinglybackend/blob/53b3eb776ee01ae77d2cd447c148b69aeee4b6ba/resources/user.py#L46-L55) in `resources/user.py`

The `tenant_list` variable on L47 is a method, and method objects are not iterable. 

**To Reproduce**
1. Start up front and back ends
2. Click on Manage Property Managers
3. Click on a PM's name
4. See Flask error 

**Expected behavior**
Should see PM details page

**Proposed Solution**
Completely rewrite Lines 46-57 of `resources/user.py`

**Acceptance Criteria**
- [ ] Should return the tenants associated with PM's property
- [ ] Code should be readable
- [ ] Integration test should be updated to actually check this

aedwardg:

Closes codeforpdx/dwellingly-app/issues/522

This was cherry picked over from changes originally made by me and @DaveTrost in #209 

Fixes a recent break where Property managers could no longer be retrieved through the users/<id> endpoint due to a change in how the attached tenants were being handled.  I also added an integration test and updated some pytest fixtures, since the breaking change slipped through the cracks of our previous tests.

The serialization of a Property Manager's tenants data is still being handled in the endpoint.  A new issue will be opened up to migrate that serialization over to the model.

### DIFF
--- a/resources/user.py
+++ b/resources/user.py
@@ -7,6 +7,7 @@ from utils.authorizations import admin_required, admin, pm_level_required
 from models.user import UserModel, RoleEnum
 from models.revoked_tokens import RevokedTokensModel
 import json
+from itertools import chain
 
 from flask_jwt_extended import (
     create_access_token,
@@ -44,17 +45,11 @@ class User(Resource):
         user_info = user.json()
 
         if user.role == RoleEnum.PROPERTY_MANAGER:
-            user_info["properties"], tenant_list = zip(
-                *(
-                    (p.json(), p.tenants)
-                    for p in PropertyModel.find_by_manager(user_id)
-                    if p
-                )
-            )
+            properties_list = PropertyModel.find_by_manager(user_id)
+            user_info["properties"] = [p.json() for p in properties_list]
 
-            tenant_IDs = [tenant.id for sublist in tenant_list for tenant in sublist]
-            tenants_list = [TenantModel.find_by_id(t) for t in set(tenant_IDs)]
-            user_info["tenants"] = [t.json() for t in tenants_list if t]
+            tenants_json_2D = [p.tenants() for p in properties_list]
+            user_info["tenants"] = list(chain.from_iterable(tenants_json_2D))
 
         return user_info, 200
 
--- a/resources/user.py
+++ b/resources/user.py
@@ -2,7 +2,6 @@ from flask_restful import Resource, reqparse
 from flask import request
 from schemas import UserRegisterSchema
 from models.property import PropertyModel
-from models.tenant import TenantModel
 from utils.authorizations import admin_required, admin, pm_level_required
 from models.user import UserModel, RoleEnum
 from models.revoked_tokens import RevokedTokensModel
--- a/resources/user.py
+++ b/resources/user.py
@@ -36,10 +36,7 @@ class UserRegister(Resource):
 class User(Resource):
     @admin_required
     def get(self, user_id):
-        user = UserModel.find_by_id(user_id)
-
-        if not user:
-            return {"message": "User not found"}, 404
+        user = UserModel.find(user_id)
 
         user_info = user.json()
 
--- a/tests/factory_fixtures/lease.py
+++ b/tests/factory_fixtures/lease.py
@@ -20,12 +20,15 @@ def lease_attributes(faker):
 
 
 @pytest.fixture
-def create_lease(faker, lease_attributes, create_property, create_tenant):
-    def _create_lease(unitNum=None):
+def create_lease(
+    faker, lease_attributes, create_property, create_tenant, tenant=None, property=None
+):
+    def _create_lease(tenant=tenant, property=property, unitNum=None):
         if not unitNum:
             unitNum = faker.building_number()
-        tenant = create_tenant()
-        lease = LeaseModel(**lease_attributes(unitNum, tenant, create_property()))
+        tenant = tenant or create_tenant()
+        property = property or create_property()
+        lease = LeaseModel(**lease_attributes(unitNum, tenant, property))
         lease.save_to_db()
         return lease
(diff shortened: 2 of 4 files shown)


tests/factory_fixtures/property.py
```diff
@@ -23,10 +23,10 @@ def _property_attributes(archived=False, manager_ids=None):
 
 
 @pytest.fixture
-def create_property(property_attributes):
-    def _create_property():
+def create_property(property_attributes, **kwargs):
```
NickSchimek, line 26: This is very nice. I don't think we need the Kwargs here though.

tests/integration/test_users.py
```diff
@@ -81,21 +81,55 @@ def test_refresh_user(client, test_database, admin_user):
     assert responseRefreshToken.status_code == 200
 
 
-def test_get_user_by_id(client, auth_headers, admin_user):
+def test_get_user_by_id(client, empty_test_db, create_admin_user, valid_header):
     """The get user by id route returns a successful response code."""
-    user = UserModel.find_by_email(admin_user.email)
-    response = client.get(f"/api/user/{user.id}", headers=auth_headers["admin"])
+    user = create_admin_user()
+    response = client.get(f"/api/user/{user.id}", headers=valid_header)
```
NickSchimek, line 87: I think we need to assert that we get back serialized data. Asserting the status codes can be done in auth tests.

tests/integration/test_users.py
```diff
@@ -81,21 +81,55 @@ def test_refresh_user(client, test_database, admin_user):
...
     assert response.status_code == 200
 
     """
     The server responds with an error if a non-existent user id
     is requested from the get user by id route.
     """
-    responseBadUserId = client.get("/api/user/000000", headers=auth_headers["admin"])
+    responseBadUserId = client.get("/api/user/000000", headers=valid_header)
```
NickSchimek, line 94: This test can be removed. It is retesting the behavior of the find method.

resources/user.py
```diff
@@ -36,25 +36,16 @@ def post(self):
...
-                )
-            )
+            properties_list = PropertyModel.find_by_manager(user_id)
+            user_info["properties"] = [p.json() for p in properties_list]
 
-            tenant_IDs = [tenant.id for sublist in tenant_list for tenant in sublist]
-            tenants_list = [TenantModel.find_by_id(t) for t in set(tenant_IDs)]
-            user_info["tenants"] = [t.json() for t in tenants_list if t]
+            tenants_json_2D = [p.tenants() for p in properties_list]
+            user_info["tenants"] = list(chain.from_iterable(tenants_json_2D))
```
NickSchimek, line 48: Can we move this logic to the serializer? Also it seems like we are missing a relationship here. I don't think it should be this complex to serialized some data.

NickSchimek (changes requested):
Thanks @aedwardg i think a couple small changes and it should be good

tests/factory_fixtures/lease.py
```diff
@@ -21,11 +21,11 @@ def _lease_attributes(unitNum, tenant, property):
...
-    def _create_lease(unitNum=None):
-        if not unitNum:
-            unitNum = faker.building_number()
-        tenant = create_tenant()
-        lease = LeaseModel(**lease_attributes(unitNum, tenant, create_property()))
+    def _create_lease(tenant=None, property=None, unitNum=None):
+        unitNum = unitNum or faker.building_number()
+        tenant = tenant or create_tenant()
+        property = property or create_property()
+        lease = LeaseModel(**lease_attributes(unitNum, tenant, property))
```
NickSchimek, line 28: Very nice refactoring here. me like 👍

tests/integration/test_users.py
```diff
@@ -81,21 +81,55 @@ def test_refresh_user(client, test_database, admin_user):
...
+    assert response.status_code == 200
+
+    """The PM's properties are returned as a list of JSON objects"""
+    assert property_list == [prop.json()]
+
+    """
+    Tenants are retreived through the leases on each
+    property and returned as a list of JSON objects
+    """
+    assert tenants_list == [lease_1.tenant.json(), lease_2.tenant.json()]
```
NickSchimek, line 130: FYI: This can be a unit test once the serialization is moved out of the endpoint then we should move these to the unit test.

NickSchimek (approved):
Overall this looks good. Based on your comments in the other PR I will get this merged for the frontend folks, but I think the comments here should be addressed. Thanks @aedwardg

tests/factory_fixtures/property.py
aedwardg: I put the **kwargs here so I could pass `manager_ids` directly into the `create_property` fixture in the tests.  See line 108 of `tests/integration/test_users.py`

tests/integration/test_users.py
aedwardg: Ok, will do.

tests/integration/test_users.py
aedwardg: Alright, that makes sense.

resources/user.py
DaveTrost: In hindsight, this may be better encapsulated as a method in the tenants model. Find tenants by property. 
I wonder about your comment about a missing relationship. Tenant and Lease have a relationship. Lease contains a key for property. A new relationship would have to be between Tenant and Property. Is this possible? It seems redundant?

PR #213, opened 2021-02-18
Test4save to db

### LINKED ISSUE #426: Create tests for save_to_db() method
The following two tests are needed in the following cases:
 - UserModel overrides the save_to_db() method. 
 - The BaseModel's create and update must call the save_to_db() method incase any subclasses hook into it. (ex: UserModel)

yunnie:

Closes codeforpdx/dwellingly-app/issues/426

### DIFF
--- a/tests/unit/test_user.py
+++ b/tests/unit/test_user.py
@@ -1,5 +1,6 @@
 import pytest
 from models.user import UserModel, RoleEnum
+from schemas.user import UserSchema
 from unittest.mock import patch
 import jwt
 import time
@@ -65,3 +66,24 @@ class TestFixtures:
             return create_unauthorized_user()
 
         assert test_multiple_users_can_be_created()
+
+
+@pytest.mark.usefixtures("empty_test_db")
+class TestOverwrittenBaseClassMethods:
+    def test_user_save_to_db(self, user_attributes):
+        user = UserModel(**user_attributes())
+        user.save_to_db()
+        lookedup = UserModel.find_by_id(user.id)
+        assert lookedup == user
+
+    def test_update_class_method(self, create_join_staff, faker):
+        user = create_join_staff()
+        email = faker.unique.email()
+        UserModel.update(UserSchema, user.id, {"email": email})
+        lookedup = UserModel.find_by_id(user.id)
+        assert lookedup.email == email
+
+    def test_create_class_method(self, user_attributes):
+        user = UserModel.create(UserSchema, user_attributes(role=RoleEnum.STAFF.value))
+        lookedup = UserModel.find_by_id(user.id)
+        assert lookedup == user
--- a/tests/unit/test_user.py
+++ b/tests/unit/test_user.py
@@ -71,9 +71,14 @@ class TestFixtures:
 @pytest.mark.usefixtures("empty_test_db")
 class TestOverwrittenAndInheritedMethods:
     def test_user_save_to_db(self, user_attributes):
-        user = UserModel(**user_attributes())
+        attrs = UserModel.validate(
+            UserSchema, user_attributes(role=RoleEnum.STAFF.value)
+        )
+        user = UserModel(**attrs)
         user.save_to_db()
         lookedup = UserModel.find_by_id(user.id)
+        assert lookedup.password is None
+        user.password = None
         assert lookedup == user
 
     def test_update_class_method(self, create_join_staff, faker):
@@ -82,8 +87,13 @@ class TestOverwrittenAndInheritedMethods:
         UserModel.update(UserSchema, user.id, {"email": email})
         lookedup = UserModel.find_by_id(user.id)
         assert lookedup.email == email
+        assert lookedup.password is None
+        assert lookedup.hash_digest is not None
+        assert lookedup.hash_digest == user.hash_digest
 
     def test_create_class_method(self, user_attributes):
         user = UserModel.create(UserSchema, user_attributes(role=RoleEnum.STAFF.value))
         lookedup = UserModel.find_by_id(user.id)
+        assert lookedup.password is None
         assert lookedup == user
+        assert lookedup.hash_digest is not None
--- a/tests/unit/test_user.py
+++ b/tests/unit/test_user.py
@@ -78,8 +78,6 @@ class TestOverwrittenAndInheritedMethods:
         user.save_to_db()
         lookedup = UserModel.find_by_id(user.id)
         assert lookedup.password is None
-        user.password = None
-        assert lookedup == user
 
     def test_update_class_method(self, create_join_staff, faker):
         user = create_join_staff()
@@ -94,4 +92,3 @@ class TestOverwrittenAndInheritedMethods:
         user = UserModel.create(UserSchema, user_attributes(role=RoleEnum.STAFF.value))
         lookedup = UserModel.find_by_id(user.id)
         assert lookedup.password is None
-        assert lookedup == user
--- a/tests/unit/test_user.py
+++ b/tests/unit/test_user.py
@@ -76,19 +76,19 @@ class TestOverwrittenAndInheritedMethods:
         )
         user = UserModel(**attrs)
         user.save_to_db()
-        lookedup = UserModel.find_by_id(user.id)
+        lookedup = UserModel.find(user.id)
         assert lookedup.password is None
 
     def test_update_class_method(self, create_join_staff, faker):
         user = create_join_staff()
         email = faker.unique.email()
         UserModel.update(UserSchema, user.id, {"email": email})
-        lookedup = UserModel.find_by_id(user.id)
+        lookedup = UserModel.find(user.id)
         assert lookedup.email == email
         assert lookedup.password is None
         assert lookedup.hash_digest == user.hash_digest
 
     def test_create_class_method(self, user_attributes):
         user = UserModel.create(UserSchema, user_attributes(role=RoleEnum.STAFF.value))
-        lookedup = UserModel.find_by_id(user.id)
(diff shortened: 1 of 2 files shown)


tests/unit/test_user.py
```diff
@@ -65,3 +66,24 @@ def test_multiple_users_can_be_created():
...
+        user = create_join_staff()
+        email = faker.unique.email()
+        UserModel.update(UserSchema, user.id, {"email": email})
+        lookedup = UserModel.find_by_id(user.id)
+        assert lookedup.email == email
+
+    def test_create_class_method(self, user_attributes):
+        user = UserModel.create(UserSchema, user_attributes(role=RoleEnum.STAFF.value))
+        lookedup = UserModel.find_by_id(user.id)
+        assert lookedup == user
```
aedwardg: At first glance these look great, but when you dig deeper it turns out these aren't actually testing that the overridden `save_to_db` method from `UserModel` is the one that is executed.  These tests still pass even if that method is deleted.

The only difference between the `UserModel`'s method and the one in the `BaseModel` is this conditional:
https://github.com/codeforpdx/dwellinglybackend/blob/53c761cef25ae14efb7acefad824d216da7c16e9/models/user.py#L134-L136

However, we _also_ can't just assert that the user's `hash_digest` exists, because it turns out that `hash_digest` is created in two places:
- in `UserModel`'s `save_to_db` method as mentioned above;
- and also in the `UserModel` init method:
  https://github.com/codeforpdx/dwellinglybackend/blob/53c761cef25ae14efb7acefad824d216da7c16e9/models/user.py#L48-L51

This means that the `hash_digest` is created on instantiation so long as a password is in the attributes, and it is then _changed_ with each `save_to_db` (or at least it should be if the overriding method is properly called).

To demonstrate this, run the tests by checking if the `hash_digests` are the same between `user` (pre-update/save) and `lookedup`: 

```python
class TestOverwrittenAndInheritedMethods:
    def test_user_save_to_db(self, user_attributes):
        user = UserModel(**user_attributes())

        hashed = user.hash_digest

        user.save_to_db()
        lookedup = UserModel.find_by_id(user.id)
        
        assert lookedup == user
        assert lookedup.hash_digest == hashed

    def test_update_class_method(self, create_join_staff, faker):
        user = create_join_staff()
        email = faker.unique.email()

        hashed = user.hash_digest

        UserModel.update(UserSchema, user.id, {"email": email})
        lookedup = UserModel.find_by_id(user.id)

        assert lookedup.email == email
        assert lookedup.hash_digest == hashed

    def test_create_class_method(self, user_attributes):
        user = UserModel.create(UserSchema, user_attributes(role=RoleEnum.STAFF.value))

        hashed = user.hash_digest

        lookedup = UserModel.find_by_id(user.id)
        assert lookedup == user
        assert lookedup.hash_digest == hashed
```
The first 2 tests will fail because the `hash_digest` is changed by the `save_to_db` and `update` methods. The third test will pass because the `hash_digest` is only created once &mdash; through the `create` method.

aedwardg: Now, it's a good sign they fail, because that means that `UserModel`'s method is indeed the one being called each time.

However, it begs the question: should we really be creating a new `hash_digest` every time we save or update the user, _even when_ the `hash_digest` already exists? 

It might make more sense to only create a `hash_digest` when the user is created and when their password is updated...but that might also be more work than it's worth since creating a new `hash_digest` on every save doesn't seem to be hurting anything.  Interested to hear what you and @NickSchimek think!

aedwardg (commented):
@yunnie, I looked over this and the code looks good but I don't think it's testing precisely what we need to test here.  I wrote a super-long comment below on it, so sorry in advance! 😅 🙏

tests/unit/test_user.py
NickSchimek: I haven't read what you wrote, but I saw this PR come up and was meaning to comment on it. At first glance everything looked good, and does what the description said. However, I wrote that description for me as I had planned to do this. And when I unassigned myself I never updated the description. This is all my fault for not updating the description for what kind of tests need to be written. I apologize. The description I put in was for reasons we are not removing the `save_to_db` method, and not what kind of tests to write. **I will take a look at this tomorrow**, but I think what we do need are some base interface tests.

NickSchimek (commented):
Hey @yunnie . You actually interpreted the issue quite accurately. You implemented the save_to_db test and the inherited tests. Ultimately we'd want to test all the inherited methods. This is to ensure that the UserModel does nothing weird to inadvertently overwrite methods that collaborators depend on. 

The tests that need to be included are already written. We just need to tell this UserModel unit test to run them. Additionally it would be good to test the extra behavior the save_to_db method has in the UserModel which is the conditional here:
```
        if self.password:
            self.hash_digest = UserModel._hash_pw(self.password)
```

The other two statements (the actual saving) would already be under test once the base interface tests are included, although however you implement is up to you as long as we cover the happy and sad path of the conditional.

To reiterate. If we include this file `tests/unit/base_interface_test.py` then all the interface/inherited tests will be covered, and we'll just need to write a test for the additional behavior that the save_to_db method has. There is an example here: `tests/unit/test_lease.py` for including the base interface tests.

NickSchimek:
Come to think of it. I think we already test the save_to_db method somewhere, maybe the integration test, So I'm not too worried about a test here, but it might be good to have, although that might violate my over testing policy, feel free to check it out and make whatever decision you think is best. I think the most important thing is to get the base interface tests hooked up. And if we already have a UserSchema it might be relatively straight forward

tests/unit/test_user.py
NickSchimek: hey @aedwardg this is a good write up and your spot on how we want to check the conditional in the save_to_db method. 

Good catch on the hashing of the password twice, although I think this is there to be backward compatible while we do all the refactoring. Once all the places have been updated to use the `create` method then I believe we can remove the hashing of the password in the User init method. I'm not 100% sure about the reason, but I remember there was a reason why I did this, and it is only done on create, not any other time. Maybe @cecillesalazar remembers. In your test you used the User init method to create a user `UserModel(**user_attributes())` <- We need to move away from this, because it bypasses validations, and all the checks that we have put in place to ensure that a user is properly created. So yes, I think currently it hashes a password on init, and again on the save_to_db method, but maybe the one in the `init` can be removed once the refactoring for creating all the User's have been updated to use the `create` method. 

I also noticed the User factories are not using the `create` method, which they should be.

The second test is failing, because the db config is failing here. I'm not sure why 🤦 ; otherwise the password are not rehashed when the user does not update the password. Actually I think I might know why the db config is failing. <- This is going to be a high priority fix. It is real simple to fix, but I'd like to tighten up if we can. I'd be happy to go over the architecture put in place, and come up with a way to improve it with whoever wants to take this on.

Good analysis @aedwardg - Lets get the second issue fixed yesterday, and the first issue will hopefully take care of itself in time.

yunnie:
First, thank you both @aedwardg and @NickSchimek  for the careful analysis and feedback. 

It looks like the UserModel is saving <code>UserModel.password</code> in plain text in the database. I believe that's what's causing the the password to get re-hashed with <code>UserModel.update</code>. 

When I change the test to 
```
def test_user_save_to_db(self, user_attributes): 
  user  = UserModel(**user_attributes())
  pword = user.password
  user.save_to_db()
  lookedup = UserModel.find_by_id(user.id)
  assert lookedup.password  is None
```

The test fails.

I think we want to change password to "None" after it is hashed. Or we remove password from the UserModel to eliminate the possibility of us storing the password in plain text.

yunnie:
The ```update``` method in the ```BaseModel``` looks up the record by id, validates the payload, updates the value based on the payload, then saves the object back to the database.

```
@classmethod
def update(cls, schema, id, payload):
    obj = cls.find(id)
    attrs = cls.validate(schema, payload, partial=True)
  
    for k, v in attrs.items():
        setattr(obj, k, v)

    obj.save_to_db()

    return obj
```

So when the ```save_to_db()``` is called from the ```UserModel``` there is a value in ```self.password```, which is what is causing the password to get rehashed.

yunnie:
So I think the appropriate test for ```UserModel.save_to_db()``` is to make sure that the password is not saved, and that the hashed password is saved.

tests/unit/test_user.py
```diff
@@ -65,3 +66,32 @@ def test_multiple_users_can_be_created():
...
+@pytest.mark.usefixtures("empty_test_db")
+class TestOverwrittenAndInheritedMethods:
+    def test_user_save_to_db(self, user_attributes):
+        attrs = UserModel.validate(
+            UserSchema, user_attributes(role=RoleEnum.STAFF.value)
+        )
+        user = UserModel(**attrs)
+        user.save_to_db()
+        lookedup = UserModel.find_by_id(user.id)
+        assert lookedup.password is None
```
NickSchimek, line 80: Perfect 👍  - Thanks for fixing this

tests/unit/test_user.py
```diff
@@ -65,3 +66,32 @@ def test_multiple_users_can_be_created():
...
+        user.password = None
+        assert lookedup == user
```
NickSchimek: What are we testing here?

yunnie: Just making sure we don’t save the password as plain text and getting the user attributes back. I could just test to make sure we don’t save the password.

NickSchimek: Ah yeah, it might be over testing here.

aedwardg (commented):
This is great @yunnie, thanks for getting it done! The only change I'd suggest is using the newer `UserModel.find()` instead of `UserModel.find_by_id()`

Other than that, this looks good to go!

aedwardg (approved):
LGTM! 🚀

PR #216, opened 2021-03-11
Separate authorization tests from endpoint logic tests

### LINKED ISSUE #545: Separate authorization tests from endpoint logic tests
### Purpose

### Proposed Solution

### Acceptance Criteria

gussmith23:

Closes https://github.com/codeforpdx/dwellingly-app/issues/545

Finds tests that test authorization, and moves them to a new file.


gussmith23:
Talking to Nick:

At the same time we do this, we should be refactoring authorization tests to look like the following:
https://github.com/codeforpdx/dwellinglybackend/blob/87cd36203cbd4033caf64aa86b8be39ccfb589e5/tests/integration/test_leases.py#L76

Nick says: as we move forward with this, we should refactor the tests to the class-based style as we move them. So I should refactor what I've done so far, and then move on.

ALSO  I should split them up into multiple PRs for everyone's sake. A different PR for each file.

gussmith23:
Going to close this one in favor of smaller PRs.

PR #218, opened 2021-03-15
Issue532 archive property endpoint update

### LINKED ISSUE #532: Add ability to unarchive properties
**Purpose**
User can unarchive a property from the Property Details view

**Proposed Solution**
* Hide the Edit button on the property details page for Archived properties so the user cannot edit the archived property
* Add an Unarchive button to the right of the Property name where the edit icon was (Is there a "restore" icon we can use for the button?)
* On click, send a message to the backend to unarchive the property and then show the edit button again
* To the right of the Edit button, add an "Archive" button with the archive icon

**Acceptance Criteria**
- [x] Hide the Edit button on the property details page for Archived properties so the user cannot edit the archived property
- [x] Add an Unarchive button to the right of the Property name where the edit icon was (Is there a "restore" icon we can use for the button?)
- [ ] On click, send a message to the backend to unarchive the property and then show the edit button again
- [x] Add an "Archive" button with the archive icon that will archive the property when clicked

image

**Additional context**

Chris-Boe:

Closes codeforpdx/dwellingly-app/issues/532

Not directly related to issue 532, but close enough. The property archive endpoint should just return a success or error message instead of the whole property json. It should also be a 200 instead of a 201 since we are not "Creating" anything with this endpoint.

### DIFF
--- a/resources/property.py
+++ b/resources/property.py
@@ -42,8 +42,12 @@ class ArchiveProperty(Resource):
         property.archived = not property.archived
 
         property.save_to_db()
-
-        return property.json(), 201
+        message = (
+            "Property archived successfully"
+            if property.archived
+            else "Property unarchived successfully"
+        )
+        return {"message": message}, 200
 
 
 class ArchiveProperties(Resource):
--- a/tests/integration/test_properties.py
+++ b/tests/integration/test_properties.py
@@ -73,11 +73,11 @@ def test_archive_property_by_id(client, auth_headers, create_property, test_data
     assert responseNoAdmin == 401
     assert responseNoAdmin.json == {"message": "Missing authorization header"}
 
-    """The archive property endpoint should return a 201 code when successful"""
+    """The archive property endpoint should return a 200 code when successful"""
     responseSuccess = client.post(
         f"/api/properties/archive/{test_property.id}", headers=auth_headers["admin"]
     )
-    assert responseSuccess.status_code == 201
+    assert responseSuccess.status_code == 200
 
     """The property should have its 'archived' key set to True"""
     responseArchivedProperty = client.get(


NickSchimek (commented):
I'm not sure this make sense to have separate routes for each resource to handle updating a single attribute.

This could be handled through the put action on the property resource without adding/editing any code. Is there a reason for not doing that?

If we are going maintain a separate route for this then the one thing I'd suggest is to make the request a bit more predictable. For instance, this route does two different things depending on the state. I would separate that so the state doesn't impact what is being requested of it to do.

A delete action to archive the resource.

And a post action to unarchive.

That way we guarantee what work will be done when a request is made.

Chris-Boe:
> I'm not sure this make sense to have separate routes for each resource to handle updating a single attribute.
> 
> This could be handled through the put action on the property resource without adding/editing any code. Is there a reason for not doing that?
> 
> If we are going maintain a separate route for this then the one thing I'd suggest is to make the request a bit more predictable. For instance, this route does two different things depending on the state. I would separate that so the state doesn't impact what is being requested of it to do.
> 
> A delete action to archive the resource.
> 
> And a post action to unarchive.
> 
> That way we guarantee what work will be done when a request is made.

This PR wasn't to add this endpoint but just to make it make a little more sense. I do agree though that we have some misaligned API's when it comes to verb usage and purpose. I'm fine with redoing some of these API's, including those in the Tenant resource as well, but would rather not do it in this PR

NickSchimek (approved):
> This PR wasn't to add this endpoint but just to make it make a little more sense.

Sorry, I had thought it added a feature. I wasn't aware we were supporting unarchiving. And when I saw that I assumed a feature was added. 

Yes, this does improve the endpoint. Ultimately, If we are going to support unarchiving too. Then I think we should open up an issue to discuss/tighten up these endpoints, because this endpoint assumes that the frontend knows the state of the backend, which it does not.

aedwardg (approved):
LGTM! I agree though that we may want to revisit this endpoint to tighten it up in the future.  It feels a little strange that for archiving/unarchiving properties there is a separate endpoint, but then for tenants it is just the DELETE action.

PR #219, opened 2021-03-18
Separate emergency contacts auth tests

### LINKED ISSUE #545: Separate authorization tests from endpoint logic tests

gussmith23:

Moves emergency contact auth tests into their own file.

### DIFF
--- a/tests/authorizations/test_emergency_contact_authorization.py
+++ b/tests/authorizations/test_emergency_contact_authorization.py
@@ -1,0 +1,43 @@
+from conftest import is_valid
+import pytest
+
+
+@pytest.mark.usefixtures("client_class", "empty_test_db")
+class TestEmergenyContactsAuthorizations:
+    def setup(self):
+        self.endpoint = "/api/emergencycontacts"
+
+    def test_emergency_contacts_POST(self, auth_headers):
+        endpoint = "/api/emergencycontacts"
+
+        newContact = {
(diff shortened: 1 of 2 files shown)


tests/authorizations/test_emergency_contact_authorization.py
```diff
@@ -0,0 +1,43 @@
...
+import pytest
+
+
+@pytest.mark.usefixtures("client_class", "empty_test_db")
+class TestEmergenyContactsAuthorizations:
+    def setup(self):
+        self.endpoint = "/api/emergencycontacts"
+
+    def test_emergency_contacts_POST(self, auth_headers):
+        endpoint = "/api/emergencycontacts"
```
aedwardg, line 11: I think here, if you want to use `endpoint` instead of `self.endpoint`, then that variable should be defined as 
```python
endpoint = self.endpoint
```
Otherwise there doesn't seem to be any reason to have the `setup()` in there.

Same thing on line 32.

aedwardg (approved):
This is exactly what we want 👍   
I have one suggested change, but it's not a deal-breaker for me.  I'll approve this in advance so that if you do make the change you won't have to re-request my review.

NickSchimek (commented):
Thank you! This is an excellent start to getting our authorizations in order, and well documented.

PR #221, opened 2021-03-18
Move email auth tests to auth tests file

### LINKED ISSUE #545: Separate authorization tests from endpoint logic tests

gussmith23:

### DIFF
--- a/tests/authorizations/test_email_authorization.py
+++ b/tests/authorizations/test_email_authorization.py
@@ -1,0 +1,21 @@
+from conftest import is_valid
+import pytest
+
+
+@pytest.mark.usefixtures("client_class", "empty_test_db")
+class TestEmailAuthorizations:
+    def setup(self):
+        self.endpoint = "/api/user/message"
+
+    def test_auth_header_is_required(self):
+        response = self.client.post(self.endpoint)
+
+        assert is_valid(response, 401)
+        assert response.json == {"message": "Missing authorization header"}
+
+    def test_all_roles_except_admin_are_denied_access(self, auth_headers):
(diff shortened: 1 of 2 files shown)


aedwardg (approved):
LGTM! 👍

tests/authorizations/test_email_authorization.py
```diff
@@ -0,0 +1,21 @@
+from conftest import is_valid
+import pytest
+
+
+@pytest.mark.usefixtures("client_class", "empty_test_db")
+class TestEmailAuthorizations:
+    def setup(self):
+        self.endpoint = "/api/user/message"
```
NickSchimek, line 8: I want to take a closer look at this. We're testing a user endpoint here, yet we're stating that we are testing email authorizations. I'm a bit confused by what we are testing here.

aedwardg: It looks like that's the url for the endpoint.  Not sure why.
https://github.com/codeforpdx/dwellinglybackend/blob/448df61f6ea81f00dde7dab751f8b2106f0eb7b1/app.py#L55

NickSchimek:
Re-opening to trigger CI

tests/authorizations/test_email_authorization.py
NickSchimek: I wonder if the frontend devs would notice if I deleted this endpoint. I think at the very least we should create a message endpoint on the UserModel so that this authorization test wont confuse me the next time I see it. I really don't like hitting the email endpoint directly. I think the user controller should be responsible for authorizing the user action and requesting a message to be sent. A reason to do this is that we can reuse the method to send an email in different contexts.

PR #225, opened 2021-03-28
Property serialization includes only tenants with active lease

### LINKED ISSUE #520: Property endpoint GET response should only include tenants with active leases
## Purpose
Because tenants can have more than one lease (1 active, plus past leases) we need to make sure that when we [list the tenants for a property](https://github.com/codeforpdx/dwellinglybackend/pull/205), we only list those tenants who have an _active_ lease with that property.

~We'll keep this temporarily on hold until we've dealt with issue #512, as the two are highly correlated.~
EDIT: #512 is now completed, so this issue is ready to work on.

## Proposed Solution
Implementation will depend on how issue 512 is resolved, as mentioned above. 

## Acceptance Criteria
- [x] A GET request for any property should only return the tenants with _active_ leases to the property in the response JSON.
- [x] Include appropriate tests (when in doubt, consult the [Project Components doc](https://github.com/codeforpdx/dwellinglybackend/blob/development/CONTRIBUTING.md#project-components) for where to test each component)

aedwardg:

Closes codeforpdx/dwellingly-app/issues/520

Serialization of a property's tenants has been moved out of the endpoint and into the `json()` method. 
The `tenants()` method in `PropertyModel` now only includes tenants with currently active leases.

Added another unit test for the `tenants()` method and refactored the relevant integration test.

### DIFF
--- a/models/lease.py
+++ b/models/lease.py
@@ -2,6 +2,7 @@ from sqlalchemy.sql.functions import now
 from db import db
 from models.base_model import BaseModel
 from utils.time import Time
+from datetime import datetime
 
 
 class LeaseModel(BaseModel):
@@ -25,6 +26,10 @@ class LeaseModel(BaseModel):
             "dateTimeEnd": Time.format_date(self.dateTimeEnd),
             "unitNum": self.unitNum,
         }
+
+    def is_active(self):
+        now = datetime.now()
+        return now > self.dateTimeStart and now < self.dateTimeEnd
 
     @classmethod
     def active(cls):
--- a/models/property.py
+++ b/models/property.py
@@ -52,7 +52,8 @@ class PropertyModel(BaseModel):
     def tenants(self):
         tenants = []
         for lease in self.leases:
-            tenants.append(lease.tenant.json())
+            if lease.is_active():
+                tenants.append(lease.tenant.json())
         return tenants
 
     @classmethod
--- a/tests/unit/test_property.py
+++ b/tests/unit/test_property.py
@@ -2,6 +2,8 @@ import pytest
 from tests.unit.base_interface_test import BaseInterfaceTest
 from models.property import PropertyModel
 from schemas.property import PropertySchema
+from utils.time import Time
+from dateutil.relativedelta import relativedelta
 
 
 class TestBasePropertyModel(BaseInterfaceTest):
@@ -46,3 +48,22 @@ class TestPropertyModel:
         tenant = lease.tenant
 
         assert property.json(include_tenants=True)["tenants"] == [tenant.json()]
+
+    def test_only_active_tenants_included(self, create_lease):
+        active_lease = create_lease()
+
+        old_lease_end = Time._yesterday()
+        old_lease_start = old_lease_end - relativedelta(years=1)
+        expired_lease = create_lease(
+            property=active_lease.property,
+            dateTimeStart=old_lease_start,
+            dateTimeEnd=old_lease_end,
+        )
+
+        property = PropertyModel.find(expired_lease.propertyID)
+        property_json = property.json()
+
+        assert len(property_json["tenants"]) == 1
+        assert len(property_json["lease"]) == 2
+        assert active_lease.tenant.json() in property_json["tenants"]
+        assert expired_lease.tenant.json() not in property_json["tenants"]
--- a/tests/unit/test_property.py
+++ b/tests/unit/test_property.py
@@ -61,9 +61,9 @@ class TestPropertyModel:
         )
 
         property = PropertyModel.find(expired_lease.propertyID)
-        property_json = property.json()
+        property_json = property.json(include_tenants=True)
 
         assert len(property_json["tenants"]) == 1
-        assert len(property_json["lease"]) == 2
+        assert len(property_json["leases"]) == 2
         assert active_lease.tenant.json() in property_json["tenants"]
         assert expired_lease.tenant.json() not in property_json["tenants"]


NickSchimek:
Re-opening to trigger CI

tests/integration/test_properties.py
```diff
@@ -38,31 +36,25 @@ def test_post_property(
...
+    def setup(self):
+        self.endpoint = "/api/properties"
+
+    def test_get_property_by_id(self, create_lease, valid_header):
+        """The get property by name returns a successful response code."""
+        lease = create_lease()
+        property = lease.property
+        response = self.client.get(f"{self.endpoint}/1", headers=valid_header)
+        property_info = response.get_json()
+        assert property_info == property.json()
```
NickSchimek: There are a lot of temp variables here that make this hard to parse. 
Can we change this to:
```python
        property =create_lease().property
        response = self.client.get(f"{self.endpoint}/{property.id}", headers=valid_header)
        assert response.get_json() == property.json()
```

Also replace the hardcoded :id as that lowers my confidence in this test.

models/lease.py
```diff
@@ -26,6 +27,10 @@ def json(self):
             "unitNum": self.unitNum,
         }
 
+    def is_active(self):
+        now = datetime.now()
+        return now > self.dateTimeStart and now < self.dateTimeEnd
+
```
NickSchimek, line 33: How is this different from the existing `active` method? Do we need both?

NickSchimek (commented):
This looks good, and it seems like it would solve the issue. However, it would potentially cause a lot of unnecessary database hits. I think if we had something where we can ask for `property.tenants` that returned all tenants with an active lease.

Thinking this through. A Tenant who has an active lease belongs to a property. If we join the tenants table and the lease table to return all tenants with an active lease and then join that with the property table where tenant.property.id = property.id then that should give us the result we need. see https://docs.sqlalchemy.org/en/14/orm/queryguide.html

You could also go the other direction which might actually be better. Join Property and lease table to get all active leases for that property and then lease has a direct association to tenant. Yeah, I think I like that better.

It also seems like we're missing an association here. A property has many tenants through leases. We could build a table for this, but I think building a query is an easier solution.

I'm fine either way too. We could stick with the current changes too. I'm not too concerned about performance at this time.

models/property.py
```diff
@@ -50,7 +51,8 @@ def json(self):
     def tenants(self):
         tenants = []
         for lease in self.leases:
-            tenants.append(lease.tenant.json())
+            if lease.is_active():
+                tenants.append(lease.tenant.json())
```
NickSchimek, line 56: There will always be way more inactive leases than active leases. We should be able to build a query that only returns active leases, and plug that in here. So we could do something like `self.leases.active:` Once you have all the active leases you could ask the database for those tenants as well instead of iterating through all active leases and asking the db for the tenant.

aedwardg:
@NickSchimek
 
Good point about the multiple DB hits, I wasn't even thinking about that when I wrote this PR. The main reason I created the `is_active()` method was to avoid using the `active()` class method. Using the `active()` class method would require importing the LeaseModel inside `models/tenant.py`, and as we keep importing one model for another model's methods we continue to run into the circular import problem.

I'll take a closer look and go over the sqlalchemy docs to see if I can come up with a better solution this week based on your suggestions.

NickSchimek:
It would be nice to ask the stakeholder the qualifications for being considered a tenant of a property. For instance, you could let your lease expire and still be living at the property. In that case I would still expect the list of tenants to include tenants who do not have a current lease and are still occupying a dwelling at the property. I think to keep things simple and decrease the risk of assumptions interfering with the app's usability we should treat all leases as being active. Currently, I think the frontend only supports one lease anyway as there isn't a UI to create a new lease that is separate from the original lease or even inactivate a lease, for instance a lease might end early even if its end date hasn't came yet. So, before spending time implementing features that treat expired leases as inactive, and make it so residents don't appear in a list of tenants we should simply treat all leases as active knowing that once we get more information then our future selves can improve upon this accordingly. With that said it probably makes more sense to add some kind of inactive attribute instead of doing it off of dates. And to ensure that only one lease can be active at any time then there are a couple ways to handle this.

NickSchimek:
Ok. After thinking about this some more and seeing what the frontend wants/has. They only have knowledge of 1 lease being in existance without any way to remove or add leases. I think lets move forward with this and get this app in front of the user so that we can find out what they really want.

models/lease.py
NickSchimek: oh, I realized this is an instance method, and the other one is a class method. I'm still curious if we need both of them. I think we could get all the active leases by doing something like:
`active_leases = self.leases.filter(LeaseModel.active())`
And from all the active_leases, pull all the tenant_ids out into an Array.

And then ask the database to retrieve all tenants with those ids.
TenantModel.query`.get_by_ids`(tenant_ids) <- I'm sure they have a method for this.

We should also be able to skip those last two steps and ask the db to get all the tenants for active_leases.

tests/unit/test_property.py
```diff
@@ -46,3 +48,22 @@ def test_json_with_tenants(self, create_property, create_lease):
...
+            dateTimeEnd=old_lease_end,
+        )
+
+        property = PropertyModel.find(expired_lease.propertyID)
+        property_json = property.json(include_tenants=True)
+
+        assert len(property_json["tenants"]) == 1
+        assert len(property_json["leases"]) == 2
+        assert active_lease.tenant.json() in property_json["tenants"]
+        assert expired_lease.tenant.json() not in property_json["tenants"]
```
NickSchimek, line 69: This comment is kinda picky. It seems like we have more assertions than what is needed. 
These assertions are confusing because I have to understand the context before knowing what the assertions mean.
```
        assert len(property_json["tenants"]) == 1
        assert len(property_json["leases"]) == 2
```
I don't think we need an assertion on the leases if we're testing the tenants.

For the assertions I think we could simplify it to
`assert property_json["tenants"]) == [active_lease.tenant.json()]`

I believe this one test confirms what three of your tests are testing for.

NickSchimek (approved):
Thanks for the PR. This looks good! I have a couple comments as usual. If I review my own PRs then I 'd litter them with comments too.

PR #227, opened 2021-04-01
Dockerized

### LINKED ISSUE #320: Docker for developer setup
**Purpose**
Docker can be used to set up projects for new developers to easily start developing -- you can run one command and get a server w/ hot reloading running in a docker container. See how Seattle's [Conviction Vacation](https://github.com/openseattle/convictionvacation#development-environment-with-docker) project handles this! I heard there was discussion of using Docker to set up the Dwellingly projects. I just wanted to note the fact that Docker can make developer setup a lot easier too, if it's set up in the right way! 

**Proposed Solution**
See [this link](https://github.com/openseattle/convictionvacation#development-environment-with-docker) to see how another project sets up Docker for development.

**Acceptance Criteria**
- [ ]

**Additional context**

donovan-PNW:
Closes codeforpdx/dwellingly-app/issues/320
Optional Docker install

- No. However, it will automatically re-seed the database if one does not already exist
- Docker for desktop, but no new dependencies within the app. Note that the docker image is reading from requirements.txt and so any later changes to the pipfile will need to be copied into  requirements.txt
- Able to create a new user, but unable to sign in. This behavior seemed to exist while using pipenv as well.
- Not tested on windows.

- [🎆] Code builds successfully
- [🎆] Code has been tested and does not produce errors
- [🎆] Code is readable and formatted
- [🎆] There isn't any unnecessary commented-out code

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,19 @@
+FROM python:3.8
+
+ENV SECRET_KEY='dwellingly'
+ENV PORT=5000
+ENV LC_ALL C.UTF-8
+ENV LANG C.UTF-8
+ENV PYTHONFAULTHANDLER 1
+ENV PYTHONDONTWRITEBYTECODE 1
+ENV FLASK_SERVER_PORT=5000
+EXPOSE 3000
+EXPOSE 5000
+
+WORKDIR /root
+RUN apt-get update -y && apt-get install -y
+COPY ./requirements.txt /root/requirements.txt
+RUN pip3 install -r requirements.txt
+COPY . .
+
+ENTRYPOINT ["./docker_bash.sh"]
--- a/docker_bash.sh
+++ b/docker_bash.sh
@@ -1,0 +1,6 @@
+#!/bin/bash
+
+if [ ! -f ./data-dev.sqlite3 ]; then
+    flask db create
+fi
+flask run --reload --host=0.0.0.0
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,0 +1,27 @@
+# [dev-packages]
+pytest-flask
+coverage
+pytest-cov
+flake8
+freezegun
+pytest-watch
+python-dotenv
+pytest
+
+# [packages]
+Flask==1.1.1
+Flask-JWT-Extended==3.24.1
+Flask-RESTful==0.3.7
+Flask-SQLAlchemy==2.5.1
+Jinja2==2.11.1
+PyJWT==1.7.1
+Werkzeug==1.0.0
+Flask-mail==0.9.1
+Flask-Cors
+psycopg2-binary
+bcrypt==3.1.7
+flask-marshmallow
+marshmallow-sqlalchemy
+# not sure how to handle
+# [scripts]
+# view_coverage = "coverage report -m"


NickSchimek:
Re-opening to trigger CI

NickSchimek:
Thanks for the PR. I didn't realize there was an issue for this. Any idea on how to automatically keep the requirements.txt file up to date?

donovan-PNW:
Sure thing! Not sure if it's really necessary but it was a neat exercise.

I _think_ that there's a way to do that -- I can try to get it working and push another commit.

aedwardg:
I wonder if it would be better to use pipenv directly in Docker to install the deps rather than a `requirements.txt` ?  The `requirements.txt` in this PR is missing a few things that are in the Pipfile.
There's quick video here showing how it can be done:
https://jonathanmeier.io/using-pipenv-with-docker/

The main difference between that tutorial and this project is that the RUN line would need to be changed to this:
`RUN pipenv install -d --pre --system --deploy` to ensure that our dev dependencies and Black (which is still pre-release) get installed.

I'm not familiar with Docker, so I don't know if our pre-commit hook needs to be run from there, but I'd assume so. Otherwise, we'd need a virtual environment outside of docker as well.

If it gets run inside docker, maybe another line at the end of the Dockerfile that says `RUN pre-commit install` to make sure that that gets set up.  I'm not sure whether or not the config files would also need to be copied before that command.

donovan-PNW:
@aedwardg I think that running pipenv within a docker container would work, but my main argument against doing so is that a docker container is meant to be a complete and self-enclosed environment. So running pipenv inside of a python3 container would somewhat defeat the purpose. I'm not sure if there are any real-world performance implications except that it would reintroduce any extra quirks/bugginess from pipenv.

That said, I think that I can just install pipenv and then use it to export the contents of Pipfile to requirements.txt, per [this article](https://pythonspeed.com/articles/pipenv-docker/) (toward the bottom). Each RUN command should be cached, and so the command 
`RUN pipenv lock --keep-outdated --requirements > requirements.txt` 
would only run if the Pipfile has changed, and flask could run within an isolated environment.

aedwardg:
@donovan-PNW, 

The `--system` flag I included above tells pipenv to install the packages directly into the system Python (in this case, the Docker container's Python), so it wouldn't be in an extra virtual environment inside the Docker container, if that's what you're worried about.

I'm on board with the idea of having an optional Docker set up, but I'm mostly concerned about 2 things:
1. Maintaining a separate `requirements.txt` file:
   - My first preference would be to not have one, and just use the Pipfile and Pipfile.lock we already have to keep track of dependencies
   - If for some reason this doesn't work or doesn't work as well, I'd like the `requirements.txt` to be automatically updated when a new dependency is added or an existing one is updated.
2. The Docker setup needs to work with our current pre-commit hook
   - The hook does quite a bit to keep incoming PRs clean and consistent before they ever get reviewed so that needs to be working with this Docker setup as well.

NickSchimek:
This may be helpful. Here is pipenv's documentation for generating a requirements.txt file. https://pipenv.pypa.io/en/latest/advanced/#generating-a-requirements-txt

donovan-PNW:
@aedwardg Hmm, very cool. I wasn't familiar enough with pipenv to know that you could install packages at a system level. I'll give it a shot and post an update.

Regarding a separate requirements.txt file: I agree that it should be unnecessary, so if we can work directly with the Pipfile then it would be nice to circumvent it. If we do find that the docker image needs a pip-friendly requirements.txt file, then a happy medium might be to simply generate it within the container and store it somewhere outside of the working directory (like `/etc/requirements.txt`). The way that the image is put together now, a running container has read/write access to the host system's dwellinglybackend directory, so any changes it makes within its own working directory are reflected in a user's actual filesystem. But anything outside of the container's home directory is isolated from the host system, so even though we would produce a requirements.txt file inside of a container it wouldn't exist within dwellinglybackend's git repository IRL. I did something similar with `node_modules/` on the frontend.

Re the pre-commit hook: Interesting! pre-commit hooks are new to me. I might be able to get it to trigger the hook without needing to open a shell in the container (which is a bit tedious) but I'll have to play around with it to see. Otherwise, running the backend with docker might only be useful as a way of supporting devs who are working on dwellingly-app without making changes to this repo.

NickSchimek:
Are you still working on this @donovan-PNW ?

aedwardg:
Closing due to inactivity.  If you continue working on this, let us know and we can open this back up.

PR #228, opened 2021-04-02
Add API for deleting multiple tickets at once

Chris-Boe:

### DIFF
--- a/resources/tickets.py
+++ b/resources/tickets.py
@@ -71,6 +71,7 @@ class Tickets(Resource):
     parser.add_argument("urgency")
     parser.add_argument("issue")
     parser.add_argument("assignedUserID")
+    parser.add_argument("ids", action="append")
 
     @pm_level_required
     def get(self):
@@ -86,9 +87,24 @@ class Tickets(Resource):
             return {"tickets": [ticket.json() for ticket in TicketModel.query.all()]}
 
     @pm_level_required
-    def post(self):
+    def put(self):
         data = Tickets.parser.parse_args()
         ticket = TicketModel(**data)
 
         ticket.save_to_db()
         return ticket.json(), 201
+
+    @pm_level_required
+    def post(self):
+        data = Tickets.parser.parse_args()
+
+        if not ("ids" in data and type(data["ids"]) is list):
+            return {"message": "Ticket IDs missing in request"}, 400
+
+        for id in data["ids"]:
+            ticket = TicketModel.find_by_id(id)
+            if not ticket:
+                return {"message": "Ticket cannot be deleted"}, 400
+            ticket.delete_from_db()
+
+        return {"message": "Tickets successfully deleted"}, 200
--- a/tests/integration/test_tickets.py
+++ b/tests/integration/test_tickets.py
@@ -51,35 +51,6 @@ def test_tickets_GET_one(client, test_database, auth_headers):
     assert response.json == {"message": "Ticket not found"}
 
 
-def test_tickets_POST(client, auth_headers):
-    newTicket = {
-        "senderID": 1,
-        "tenantID": 1,
-        "status": "New",
-        "urgency": "low",
-        "issue": "Lead paint issue",
-        "assignedUserID": 4,
-    }
-
-    response = client.post(endpoint, json=newTicket, headers=auth_headers["admin"])
-
-    assert is_valid(response, 201)
-    assert response.json["id"] != 0
-    assert response.json["issue"] == "Lead paint issue"
-    assert response.json["tenant"] == "Renty McRenter"
-    assert response.json["senderID"] == 1
-    assert response.json["tenantID"] == 1
-    assert response.json["assignedUserID"] == 4
-    assert response.json["sender"] == "user1 tester"
-    assert response.json["assigned"] == "Mr. Sir"
-    assert response.json["status"] == TicketStatus.New
-    assert response.json["urgency"] == "low"
-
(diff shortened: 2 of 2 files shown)


Chris-Boe:
This is the first ticket API that is actually being used from the frontend (with #528) but there were several ticket APIs already created in the backend. I think we'll likely remove or redo most of the ticket APIs and tests currently created

NickSchimek:
Re-opening to trigger CI

resources/tickets.py
```diff
@@ -86,9 +87,24 @@ def get(self):
             return {"tickets": [ticket.json() for ticket in TicketModel.query.all()]}
 
     @pm_level_required
-    def post(self):
+    def put(self):
         data = Tickets.parser.parse_args()
         ticket = TicketModel(**data)
 
         ticket.save_to_db()
         return ticket.json(), 201
```
NickSchimek, line 95: This looks like a post action to me. Can you explain this change.

resources/tickets.py
```diff
@@ -86,9 +87,24 @@ def get(self):
...
+        data = Tickets.parser.parse_args()
+
+        if not ("ids" in data and type(data["ids"]) is list):
+            return {"message": "Ticket IDs missing in request"}, 400
+
+        for id in data["ids"]:
+            ticket = TicketModel.find_by_id(id)
+            if not ticket:
+                return {"message": "Ticket cannot be deleted"}, 400
+            ticket.delete_from_db()
```
NickSchimek, line 108: This looks like it should be a delete action instead of a post action. Can you explain these changes

resources/tickets.py
```diff
@@ -86,9 +87,24 @@ def get(self):
...
+
+    @pm_level_required
+    def post(self):
+        data = Tickets.parser.parse_args()
+
+        if not ("ids" in data and type(data["ids"]) is list):
+            return {"message": "Ticket IDs missing in request"}, 400
+
+        for id in data["ids"]:
+            ticket = TicketModel.find_by_id(id)
```
NickSchimek, line 105: find_by_id is deprecated. Please use the `find` method

NickSchimek (changes requested):
The issue says to add ability to delete tickets, yet these changes are doing other things as well. Please explain/update

tests/integration/test_tickets.py
```diff
@@ -51,35 +51,6 @@ def test_tickets_GET_one(client, test_database, auth_headers):
     assert response.json == {"message": "Ticket not found"}
 
 
-def test_tickets_POST(client, auth_headers):
```
NickSchimek, line 54: We shouldn't be deleting tests when adding a feature. Please explain why these deletions are necessary.

resources/tickets.py
NickSchimek: When doing a batch update:
 - How should we handle if one of the ids does not exist. Do we abort all, or do we update the ones that exist?

Chris-Boe:
@NickSchimek to address most of your Q's, none of the API's in this Tickets resource were being used at all, so I've changed some around and deleted some tests that were unnecessary or were replaced. The reason for using a POST instead of a DELETE (and changing the previous post api to a put because of its behavior) was to utilize the list of id's that are being sent from the frontend in the body of the request. POST is very often used for every single endpoint in a REST style API when you are object + verb oriented instead of just object + CRUD. When the endpoint was a DELETE, the request body was being ignored for some reason. I'm still unfamiliar with flask so I may be missing something!

tests/integration/test_tickets.py
Chris-Boe: It was replaced by the one added below

resources/tickets.py
Chris-Boe: I was going to bring up that batch update question on Wednesday 😄 As for the "find" vs "find_by_id", I see "find_by_id" being used in a bunch of places throughout the app, should I update all of them?

NickSchimek:
hey chris, these endpoints were created before my involvement; hence I am unaware of the context for why they were created. With that said, if there are endpoints that are currently not being used I would be hesitant to remove them. Obviously someone thought they were needed. And even if this issue is discussed and it's decided that they're not needed. I would still be hesitant to remove functionality that enables CRUD actions on a resource. An example of this would be the Lease resource.

Generally speaking, a PUT request can be used to create a new resource, however, it has a different meaning than POST. PUT also updates a resources. Since the method you changed to a `PUT` is strictly creating a resource then it is not doing what most people would think it should do.

It sounds like you had some issues creating the feature you need so some things were changed around to get it to work with the reasoning that these endpoints are not being used anyway. That would normally be fine, but these changes add more confusion rather than clarification. Can you let me know what you need for the frontend, and I can outline how we can implement that for the backend. We have a lot of options available to us without needing to change existing behavior.

FYI: The endpoint documentation will need to be updated when endpoints are added, behavior is modified, or removed.

NickSchimek:
One other point. You linked this PR to close an issue that is labeled as frontend. I'm a bit confused about this. If you need behavior from the backend that is currently not provided then lets get a separate issue created tagged with a backend label so that we can discuss the best way to go about this. thank you.

Chris-Boe:
@NickSchimek I had some weird behavior from the frontend that then caused me to reevaluate the backend approach but I've resolved it, so I'm going to open a new PR under a new backend specific issue so we can have a clean slate! Sorry for the confusion. I have some questions I want to ask about our API patterns so I'll bring them up in the meeting

PR #229, opened 2021-04-03
modified user schema to serialize proprty managers

### LINKED ISSUE #543: UserModel serialization should include properties & tenants for PMs
## Purpose

Right now, PM information includes a list of assigned properties and tenants when retrieved through the User endpoint, but not anywhere else. So that we don't have to re-implement that logic every place we want to serialize a PM, it should be moved from the User endpoint (resource) to the UserModel.

## Proposed Solution

Move the serialization of a PM's assigned properties and tenants to the UserModel `json` method.
The corresponding test should be moved out of the integration tests and into the unit tests.

## Acceptance Criteria
- [ ] The UserModel `json` method should include assigned properties and tenants if the user is a Property Manager.
- [ ] The serialization logic should no longer be in the endpoint
- [ ] The corresponding test should be under the unit tests for the UserModel
- [ ] If any additional methods are created, there should be unit tests for those as well

## Additional context

The serialization logic is here:

https://github.com/codeforpdx/dwellinglybackend/blob/985561ecbe21d0d38b0abbf852c1776bee482863/resources/user.py#L43-L48

And the test that should become a unit test is here:
https://github.com/codeforpdx/dwellinglybackend/blob/985561ecbe21d0d38b0abbf852c1776bee482863/tests/integration/test_users.py#L126-L130

yunnie:

Closes codeforpdx/dwellingly-app/issues/543

### DIFF
--- a/resources/user.py
+++ b/resources/user.py
@@ -1,13 +1,12 @@
 from flask_restful import Resource, reqparse
 from flask import request
 from schemas import UserRegisterSchema
+from schemas import UserSchema
 from models.property import PropertyModel
 from utils.authorizations import admin_required, admin, pm_level_required
 from models.user import UserModel, RoleEnum
 from models.revoked_tokens import RevokedTokensModel
 import json
-from itertools import chain
-
 from flask_jwt_extended import (
     create_access_token,
     create_refresh_token,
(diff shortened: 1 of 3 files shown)


yunnie:
The challenge was the circular relationship between the UserModel and the PropertyModel. Adding serialization of the properties in the `UserModel.json()` method can result in an infinite mutual recursion. To deal with this I used the `UserSchema` to serialize the user information in the get user resource.

NickSchimek:
Re-opening to trigger CI

NickSchimek (approved):
Nice work!

PR #231, opened 2021-04-04
Add backref for StaffTenant relationship

### LINKED ISSUE #567: DB relationship overlap in StaffTenantsLink
**Describe the bug**
After bumping Flask-SQLAlchemy up to version 2.5.1, when you re-seed the db you get two SQLAlchemy warnings about overlapping relationships in the `StaffTenantLink` model:
- `StaffTenantLink.tenant` and `TenantModel.staff` both write values to `staff_tenant_links.tenant_id`; and 
- `StaffTenantLink.staff` and `TenantModel.staff` both write values to `staff_tenant_links.staff_id`

It's unclear why this warning didn't show before, but it looks like you can make the warning go away by using the new [`overlaps`](https://docs.sqlalchemy.org/en/14/orm/relationship_api.html#sqlalchemy.orm.relationship.params.overlaps) kwarg.  An example of how to use it can be found [here](https://docs.sqlalchemy.org/en/14/errors.html#error-qzyx).

That said, I think using `overlaps` would be a sub-optimal solution.  I believe this issue stems from improperly setting up the many-to-many relationship and could probably be easily solved using this as an example:
https://docs.sqlalchemy.org/en/14/orm/tutorial.html#orm-tutorial-many-to-many

However, if we go that route, we'll need to move stuff from the `StaffTenantSchema` to the `UserSchema` and `TenantSchema` and update all tests as necessary.

**To Reproduce**
1. If [backend PR 226](https://github.com/codeforpdx/dwellinglybackend/pull/226) isn't merged yet, checkout that branch run the `pipenv run dev-install` command to upgrade Flask-SQLAlchemy
2. If that PR is merged, run `pipenv run dev-install`
3. Re-seed the db with `pipenv run flask db recreate`

**Expected behavior**
The db should be re-seeded with no warnings.  Instead, the two warnings below show up 👇 

```
SAWarning: relationship 'StaffTenantLink.tenant' will copy column tenants.id to column staff_tenant_links.tenant_id, 
which conflicts with relationship(s): 'TenantModel.staff' (copies tenants.id to staff_tenant_links.tenant_id). 
If this is not the intention, consider if these relationships should be linked with back_populates, or if viewonly=True 
should be applied to one or more if they are read-only. 
For the less common case that foreign key constraints are partially overlapping, the orm.foreign() annotation can be used to
isolate the columns that should be written towards. The 'overlaps' parameter may be used to remove this warning.
```
```
SAWarning: relationship 'StaffTenantLink.staff' will copy column users.id to column staff_tenant_links.staff_id, 
which conflicts with relationship(s): 'TenantModel.staff' (copies users.id to staff_tenant_links.staff_id). 
If this is not the intention, consider if these relationships should be linked with back_populates, or if viewonly=True 
should be applied to one or more if they are read-only. 
For the less common case that foreign key constraints are partially overlapping, the orm.foreign() annotation can be used to
isolate the columns that should be written towards. The 'overlaps' parameter may be used to remove this warning.
```

 - OS: mac OS Big Sur, Windows 10
 - Version: Flask-SQLAlchemy 2.5.1, SQLAlchemy 1.4.4

**Additional context**

NickSchimek:
Added backref so that we can ask these queries. `staff_user.tenants` and `tenant.staff`

Removed the staff and tenant relationship on the StaffTenant table. It is not necessary to have these and they were generating the below warnings. If we need to work with the table directly in the future and need to make such queries then we can adjust at that time.

```
python3.8/site-packages/sqlalchemy/orm/relationships.py:3435: SAWarning: relationship 'StaffTenantLink.tenant' will copy column tenants.id to column staff_tenant_links.tenant_id, which conflicts with relationship(s): 'TenantModel.staff' (copies tenants.id to staff_tenant_links.tenant_id). If this is not the intention, consider if these relationships should be linked with back_populates, or if viewonly=True should be applied to one or more if they are read-only. For the less common case that foreign key constraints are partially overlapping, the orm.foreign() annotation can be used to isolate the columns that should be written towards.   The 'overlaps' parameter may be used to remove this warning.
  util.warn(
python3.8/site-packages/sqlalchemy/orm/relationships.py:3435: SAWarning: relationship 'StaffTenantLink.staff' will copy column users.id to column staff_tenant_links.staff_id, which conflicts with relationship(s): 'TenantModel.staff' (copies users.id to staff_tenant_links.staff_id). If this is not the intention, consider if these relationships should be linked with back_populates, or if viewonly=True should be applied to one or more if they are read-only. For the less common case that foreign key constraints are partially overlapping, the orm.foreign() annotation can be used to isolate the columns that should be written towards.   The 'overlaps' parameter may be used to remove this warning.
  util.warn(
```

Fixes codeforpdx/dwellingly-app/issues/567

### DIFF
--- a/models/staff_tenant_link.py
+++ b/models/staff_tenant_link.py
@@ -1,7 +1,4 @@
 from db import db
-from sqlalchemy.orm import relationship
-from models.tenant import TenantModel
-from models.user import UserModel
 from models.base_model import BaseModel
 
 
@@ -9,8 +6,6 @@ class StaffTenantLink(BaseModel):
     __tablename__ = "staff_tenant_links"
     tenant_id = db.Column(db.Integer, db.ForeignKey("tenants.id"), primary_key=True)
     staff_id = db.Column(db.Integer, db.ForeignKey("users.id"), primary_key=True)
-    tenant = relationship(TenantModel)
-    staff = relationship(UserModel)
 
     # might want to set up relationships and backref's?
     # import from sqlalchemy.orm
--- a/models/tenant.py
+++ b/models/tenant.py
@@ -16,7 +16,7 @@ class TenantModel(BaseModel):
     archived = db.Column(db.Boolean, default=False, nullable=False)
 
     # relationships
-    staff = relationship("UserModel", secondary="staff_tenant_links")
+    staff = relationship("UserModel", secondary="staff_tenant_links", backref="tenants")
     leases = db.relationship(
         "LeaseModel", backref="tenant", lazy="dynamic", cascade="all, delete-orphan"
     )


aedwardg:
@NickSchimek I included some other SQLAlchemy doc links in the issue for this problem:
https://github.com/codeforpdx/dwellingly-app/issues/567

I was tinkering around with it when I wrote the issue and I'm pretty sure I got it working by following the structure used at this link:
https://docs.sqlalchemy.org/en/14/orm/tutorial.html#orm-tutorial-many-to-many

However, in doing so, StaffTenantLink only ends up having `tenant_id` and `staff_id`, so we'd have to change the StaffTenantSchema and all the related tests

aedwardg:
Basically, in `StaffTenantLink`, delete `tenant` and `staff`.

Then in the `TenantModel`, change `staff` to:
```python
staff = relationship("UserModel", secondary="staff_tenant_links", back_populates="tenant")
```

And in `UserModel`, add a relationship field called `tenant`:
```python
tenant = db.relationship("TenantModel", secondary="staff_tenant_links", back_populates="staff")
```

I didn't do any testing on this solution, but it seems to work at first glace.  Like I said though, the schemas/tests would have to change as well.

aedwardg (approved):
This seems to solve our problem, and succinctly at that!

I'll will want to discuss further via Slack or at our next meeting whether we want to keep the StaffTenantSchema and its tests where they are or whether they should be moved after this change.

PR #242, opened 2021-04-05
Create a PyTest config file

NickSchimek:
Created config file for running pytest.
Turn off capture in config file to allow for printing to stdout.

This has two advantages. It makes it easier to add print debug messages for developers, and allows us to add deprecation message for the `find_by_id` method :)


aedwardg (dismissed):
Looks good to me! I like the idea of being able to add deprecation warnings to things we haven't been able to remove yet.

Also, not having to make the test fail to see `print` statements will be nice!

PR #246, opened 2021-04-08
Add delete API for deleting multiple tickets at once

### LINKED ISSUE #573: Add delete endpoint for multiple tickets
**Purpose**
An endpoint at /tickets that will allow the deletion of a list of Tickets for issue #528 

**Proposed Solution**
* A new endpoint that will take an array of ticket ids and delete all of them from the database

**Acceptance Criteria**
- [ ] Unit Tests should be created
- [ ] Should be usable from the Tickets page in issue #528 
- [ ] Protected by Property Manager permissions

**Additional context**

Chris-Boe:

Closes codeforpdx/dwellingly-app/issues/573

### DIFF
--- a/resources/tickets.py
+++ b/resources/tickets.py
@@ -2,6 +2,7 @@ from flask_restful import Resource, reqparse
 from models.tickets import TicketModel
 from models.notes import NotesModel
 from utils.authorizations import pm_level_required
+from flask import request
 
 
 class Ticket(Resource):
@@ -95,8 +96,7 @@ class Tickets(Resource):
 
     @pm_level_required
     def delete(self):
-        self.parser.add_argument("ids", action="append")
-        data = Tickets.parser.parse_args()
+        data = request.json
 
         if not ("ids" in data and type(data["ids"]) is list):
             return {"message": "Ticket IDs missing in request"}, 400
--- a/resources/tickets.py
+++ b/resources/tickets.py
@@ -101,10 +101,8 @@ class Tickets(Resource):
         if not ("ids" in data and type(data["ids"]) is list):
             return {"message": "Ticket IDs missing in request"}, 400
 
-        for id in data["ids"]:
-            ticket = TicketModel.find(id)
-            if not ticket:
-                continue
-            ticket.delete_from_db()
+        TicketModel.query.filter(TicketModel.id.in_(data["ids"])).delete(
+            synchronize_session="fetch"
+        )
 
         return {"message": "Tickets successfully deleted"}, 200
--- a/resources/tickets.py
+++ b/resources/tickets.py
@@ -92,3 +92,19 @@ class Tickets(Resource):
 
         ticket.save_to_db()
         return ticket.json(), 201
+
+    @pm_level_required
+    def delete(self):
+        self.parser.add_argument("ids", action="append")
+        data = Tickets.parser.parse_args()
+
+        if not ("ids" in data and type(data["ids"]) is list):
+            return {"message": "Ticket IDs missing in request"}, 400
+
+        for id in data["ids"]:
+            ticket = TicketModel.find(id)
+            if not ticket:
+                continue
+            ticket.delete_from_db()
+
+        return {"message": "Tickets successfully deleted"}, 200
--- a/tests/integration/test_tickets.py
+++ b/tests/integration/test_tickets.py
@@ -141,7 +141,16 @@ def test_tickets_DELETE(client, auth_headers, create_ticket):
     assert is_valid(response, 404)
     assert response.json == {"message": "Ticket not found"}
 
+
+def test_tickets_DELETE_list(client, auth_headers, create_ticket):
     ticketsToDelete = [create_ticket().id, create_ticket().id]
+    deleteIds = {"ids": ticketsToDelete}
+    response = client.delete(endpoint, json=deleteIds, headers=auth_headers["pm"])
+    assert is_valid(response, 200)
+    assert response.json == {"message": "Tickets successfully deleted"}
+
+    nonExistentTicketId = 999
+    ticketsToDelete = [create_ticket().id, nonExistentTicketId]
     deleteIds = {"ids": ticketsToDelete}
     response = client.delete(endpoint, json=deleteIds, headers=auth_headers["pm"])
     assert is_valid(response, 200)
--- a/tests/integration/test_tickets.py
+++ b/tests/integration/test_tickets.py
@@ -125,7 +125,7 @@ def test_tickets_PUT(client, auth_headers):
     assert response.json == {"message": "Ticket not found"}
 
 
-def test_tickets_DELETE(client, auth_headers):
+def test_tickets_DELETE(client, auth_headers, create_ticket):
 
     response = client.delete(f"{endpoint}/{validID}")
     # UNAUTHORIZED - Missing Authorization Header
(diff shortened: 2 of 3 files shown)


resources/tickets.py
```diff
@@ -92,3 +92,19 @@ def post(self):
...
         ticket.save_to_db()
         return ticket.json(), 201
+
+    @pm_level_required
+    def delete(self):
+        self.parser.add_argument("ids", action="append")
+        data = Tickets.parser.parse_args()
+
+        if not ("ids" in data and type(data["ids"]) is list):
+            return {"message": "Ticket IDs missing in request"}, 400
```
NickSchimek, line 102: This is a validation, and should be handled in the schema. Since we wouldn't be using our normal model actions to make the update you would just call validate on the schema directly from here.
`TenantModel.validate(TenantSchema, request.json)`.

resources/tickets.py
```diff
@@ -92,3 +92,19 @@ def post(self):
 
         ticket.save_to_db()
         return ticket.json(), 201
+
+    @pm_level_required
+    def delete(self):
+        self.parser.add_argument("ids", action="append")
+        data = Tickets.parser.parse_args()
```
NickSchimek: The parser is being removed, let's not add to our tech debt. You can access the endpoints params with `request.json`

resources/tickets.py
```diff
@@ -92,3 +92,19 @@ def post(self):
...
+
+        if not ("ids" in data and type(data["ids"]) is list):
+            return {"message": "Ticket IDs missing in request"}, 400
+
+        for id in data["ids"]:
+            ticket = TicketModel.find(id)
+            if not ticket:
+                continue
+            ticket.delete_from_db()
```
NickSchimek: I have the same question from the previous PR #228 concerning how to handle id's that are non-existent. Do we ignore or do we abort the batch delete? I'm asking again, because it is still not clear to me based on the code changes alone.

tests/integration/test_tickets.py
```diff
@@ -141,6 +141,17 @@ def test_tickets_DELETE(client, auth_headers):
...
+    deleteIds = {"ids": ticketsToDelete}
+    response = client.delete(endpoint, json=deleteIds, headers=auth_headers["pm"])
+    assert is_valid(response, 200)
+    assert response.json == {"message": "Tickets successfully deleted"}
+
+    missingIds = {"notIds": [5]}
+    response = client.delete(endpoint, json=missingIds, headers=auth_headers["pm"])
+    assert is_valid(response, 400)
+    assert response.json == {"message": "Ticket IDs missing in request"}
+
```
NickSchimek: Can we put these tests into a different function? It makes it difficult to read here since they're testing a different endpoint then when we're deleting a single resource.

resources/tickets.py
```diff
@@ -92,3 +92,19 @@ def post(self):
 
         ticket.save_to_db()
         return ticket.json(), 201
+
+    @pm_level_required
+    def delete(self):
```
NickSchimek, line 98: hmmm, I think I like this. I probably would've suggested to create a new endpoint called batch-delete or something, but this works too.

NickSchimek (changes requested):
Thanks for the PR Chris, this is looking good. 

Since we're adding a new endpoint the Endpoint documentation needs to be updated accordingly.

resources/tickets.py
Chris-Boe: To me, it doesn't make sense to abort the deletion of a group of tickets if one of the tickets requested to delete is already gone (i.e. concurrent users deleting the same resources). The frontend just needs to fix the possible sync issue by re-requesting the resources or adjusting the displayed items based on the response of this call

resources/tickets.py
Chris-Boe: Is there documentation on things we have deprecated but haven't replaced yet? I seem to keep using functions or patterns that are "deprecated" but are still used abundantly throughout the app

resources/tickets.py
NickSchimek: If that's the case then there are tests missing here to document this behavior. The logic here will abort on the first id that is non-existent in the db. Also, I believe a batch delete can be done in one go with something like this.
```
        TicketModel.query.filter(
            TicketModel.id.in_(data["ids"])
        ).delete(synchronize_session="fetch")
```

resources/tickets.py
NickSchimek: Yes, thanks for asking. The CONTRIBUTING.md document

resources/tickets.py
Chris-Boe: "ids" is not part of the TicketModel so I am not sure why we would be validating it with the Model's schema, or am I misunderstanding something? If I put a validator on the ids parameter and validate with 
`TicketModel.validate(TicketSchema, request.json)`
I get 
```
File "C:\Users\cboe1\.virtualenvs\dwellinglybackend-vISDP1tA\Lib\site-packages\marshmallow\schema.py", line 1116, in _invoke_field_validators
    raise ValueError(
ValueError: "ids" field does not exist.```
On top of that, we don't want to validate that each id matches a ticket in the database because I would think we'd want to go ahead with deleting all the ones in the request that exist and ignore the others. If not I can change the behavior of this API.

File: resources/tickets.py (hunk 3 above)
Comment (Chris-Boe) [reply resources/tickets.py]: Ah. I did not realize "find" threw a 404 if it did not get a match, while the function "find_by_id" I was using before was really what I wanted to use, but we are getting rid of that? I've replaced the find with the code you've given

File: resources/tickets.py (hunk 1 above)
Comment (NickSchimek) [reply resources/tickets.py:102]: I don't think its necessary to validate that the id exists before deleting it. And while "ids" are not part of the ticket table. The `id` in the list of 'ids' are. Marshmallow doesn't allow validations on any fields that are not declared. Here we can tell Marshmallow that there is a field called ids and it will allow us to validate it. 

Since you mentioned that you rather just ignore it when an id does not exist, then I don't think we need the validation and can simply execute the batch delete statement.

File: resources/tickets.py (hunk 3 above)
Comment (NickSchimek) [reply resources/tickets.py]: yeah, the code base was littered with find and return a 404 if it couldn't find it. So that has been abstracted away so that we're not repeating the same logic in every controller. "find_by_id" is removed because I don't think it was doing anything useful. There are sometimes cases where we want to look up an object by the id and not throw an error if it is not found, and that can be done easily with the `query.get(:id)` method.

File: resources/tickets.py (hunk 1 above)
Comment (NickSchimek) [reply resources/tickets.py:102]: oh I see, we're checking to see if the request is in the right format. Yeah this should be moved to validation. I'm not too concerned about this type of validation though. I would just delete this. If we were having other parties consume the API then it might be worth validating so that we can respond with an appropriate failure message.

**Review by NickSchimek — CHANGES_REQUESTED** [review]
This looks good. I would just remove the validation check
```
if not ("ids" in data and type(data["ids"]) is list):
            return {"message": "Ticket IDs missing in request"}, 400
```
If it turns out we need to have some form of validation then we can update later, or if you know of some scenarios where the FE may send bad requests to the backend. I would say that we should update the fe so that it doesn't happen. Although the bad requests would cause 500 errors which is not ideal, but I'm not to concerned about it since this is not an open API. I don't know. What do you think? if you think this validation is worth it then I can help get it moved to marshmallow or we can create a separate ticket for that.

Chris-Boe:
> This looks good. I would just remove the validation check
> ```
> if not ("ids" in data and type(data["ids"]) is list):
>             return {"message": "Ticket IDs missing in request"}, 400
> ```
> If it turns out we need to have some form of validation then we can update later, or if you know of some scenarios where the FE may send bad requests to the backend. I would say that we should update the fe so that it doesn't happen. Although the bad requests would cause 500 errors which is not ideal, but I'm not to concerned about it since this is not an open API. I don't know. What do you think? if you think this validation is worth it then I can help get it moved to marshmallow or we can create a separate ticket for that.

I probably would leave the validation in since it doesn't hurt and it gives a better 400 and error message if somehow this gets called without the 'ids' parameter.
I'm fine with leaving it here but would probably need help moving it to the schema validation since I'm not too familiar with it, or we can create an issue to clean up the other tech debt pieces in the Ticket resource like the use of the parser

NickSchimek (approved):
Agreed. 

The tests look good. Thanks for updating those.

Chris-Boe:
> Agreed. 
> 
> The tests look good. Thanks for updating those. 

Thanks for your help and patience! 😄

PR #247, opened 2021-04-10
special delivery: flask-mailman!

### LINKED ISSUE #564: Remove Flask Mail library replace with Flask-Mailman
**Purpose**
Remove the defunct Flask-mail library from project (may it rest in peace) 

**Proposed Solution**
replace with Flask-Mailman
Documentation: https://www.waynerv.com/flask-mailman/

**Acceptance Criteria**
- [ ] Code passes all tests
- [ ] No lingering comments or unnecessary logging
- [ ] functionality is maintained, emails can be successfully sent from backend.

**Additional context**
Flaskmail has not been updated for  6 years,  so we need to replace it.

yunnie:

Closes codeforpdx/dwellingly-app/issues/564

Any new dependencies to install? Yes (Flask-Mailman)

- [x ] Code builds successfully
- [x ] There isn't any unnecessary commented-out code


config.py
```diff
@@ -27,7 +27,6 @@ class Default(object):
     MAIL_DEFAULT_SENDER = os.environ.get("MAIL_DEFAULT_SENDER")
     MAIL_MAX_EMAILS = os.environ.get("MAIL_MAX_EMAILS", 3)
     MAIL_ASCII_ATTACHMENTS = os.environ.get("MAIL_ASCII_ATTACHMENTS", False)
-    MAIL_SUPPRESS_SEND = os.environ.get("MAIL_SUPPRESS_SEND", False)
```
NickSchimek, line 30: Can you explain this removal for me?

NickSchimek (commented):
This looks good. In addition to the one question I posted above. Has this configuration been tested manually, and does everything work? It's OK if it hasn't.

config.py
yunnie: `MAIL_SUPRESS_SEND` is not a configuration option in [Flask-Mailman.](https://www.waynerv.com/flask-mailman/). And neither are `MAIL_MAX_EMAILS` and `MAIL_ASCII_ATTCHMENTS`. There's no harm in leaving unused configurations in the application, but I think it can cause confusion.

yunnie:
I just checked it manually, using the "Forgot My Password" function. In the process I discovered that Gmail has a security feature that won't allow you to log in to your account through a "less secure application". After toggling the security off to run the test, I was able to send this email: 

```
Hello user1,

To reset your password click on the following link: Reset password link

Alternatively, you can paste the following link in your browser's address bar:

http://localhost:5000/api/reset-password/eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJ1c2VyX2lkIjoxLCJleHAiOjE2MTgxMTA5MzkuNDM0ODQ2OX0.5N6dCcG-2RvRCi_jj0_fEZY6E7yQyOybI0rL2Rt4ZFs

If you have not requested a password reset simply ignore this message.

Best,

Team Dwellingly.
```

And I got this response: 
```
The response was:
DNS Error: 6915027 DNS type 'mx' lookup of dwellingly.org responded with code NXDOMAIN Domain name not found: dwellingly.org
```

NickSchimek:
That’s great! Yeah I usually temporarily turn that off to test. Did it come
through as a text email or was it the html version?

config.py
NickSchimek: Yep, agreed. Let's remove all unneeded configs from all **three** dev environments then. 

It looks like Flask-Mailman suppresses mails in the testing environment without the use of configs, which is great. It probably be great to have a tool that catches mail in the dev environment and allows you to view them. But I'm not sure how involved that would be and it sounds like a different issue then this one. I think the Flask-Mailman docs mentioned something about it, but I'm not sure.

Pipfile
```diff
@@ -27,6 +27,7 @@ Jinja2 = "==2.11.3"
 PyJWT = "==1.7.1"
 Werkzeug = "==1.0.1"
 Flask-mail = "==0.9.1"
+Flask-Mailman = "==0.2.2"
```
NickSchimek: Let's remove the Flask-mail dependency. We no longer want it.

NickSchimek (changes requested):
Thanks for manually testing this! It looks like everything is working and we're good to go. Let's get those unused configs removed from all three environments. And let's remove the Flask-mail dep

config.py
yunnie: I've removed the unused configuration options, and will look into seeing if we can set something up in the dev environment to actually review the emails sent.

yunnie:
The email was text, with an HTML alternative. I've switched the order, so now it sends HTML with a text alternative.

config.py
yunnie: Using `MAIL_BACKEND = 'console'` will display the sent email to the console. I've tested it manually, and it works.

NickSchimek: That works! Thank you 😄

NickSchimek (approved):
Thanks for this! Hopefully Flask-Mailman sticks around and makes it to version 1. It looks promising. I'm curious, what was the reason for going with Flask-Mailman?

I do find it odd that the text version of the email was displayed when the client could handle html version. Thanks for switching the order, I was curious if the order mattered. Their documentation makes it sound like that sending and receiving html emails are fancy stuff that may become the norm in the far distant future, but is currently not well supported by email clients. Perhaps we shouldn't be using javascript in our app if the future isn't here yet. 😕

aedwardg:
@NickSchimek  
> I'm curious, what was the reason for going with Flask-Mailman?

You mentioned it [here](https://github.com/codeforpdx/dwellingly-app/issues/351#issuecomment-694875543) 😄 

Hugh took a look a few weeks ago and liked it, so he closed that issue and made a new one that specifically said to use Flask-Mailman.

NickSchimek:
haha, 😂 that was like a half a year ago. I mentioned that, because it would be the easiest switch, but maybe not the best. We'll see.

@yunnie Would you like to update the CONTRIBUTING doc and remove mentions of Flask-mail and replace it with flask-mailman?

PR #249, opened 2021-04-11
Implement MethodView's into the Tenant endpoint

NickSchimek:
- Initial commit for moving toward MethodViews

   - This switches the tenant endpoint to use MethodViews instead of the now,
or soon to be, deprecated Flask-RESTful library. The only benefit we were
getting from using flask-restful was that all our responses were configured
to be json responses where as by default flask responds with text/html
when given a string.

    - This caused some issues where we call abort, to raise an error, because
those responses are automatically structured to return text instead of
a json response. So to get around this we will not use the abort method
in our app and instead use our own error response class that will return
an appropriate json response.

FYI: This PR is not changing, adding, or removing behavior. It simply moves the Tenant resource off of Flask-RESTful and separates the resource into two classes for readability. The routes have not changed.


CONTRIBUTING.md
```diff
@@ -56,7 +56,7 @@ All schemas should have unit tests, which primarily should be validation tests.
 
 ### Resources
 
-Flask-RESTful uses the term "resources" in place of "controllers" in an MVC framework and that is also the term we use in this project.
+Flask-RESTful used the term "resources" in place of "controllers" in an MVC framework and that is also the term we use in this project. If your familiar with Django this would be a View.
```
aedwardg:
```suggestion
Flask-RESTful used the term "resources" in place of "controllers" in an MVC framework and that is also the term we use in this project. If you're familiar with Django this would be a View.
```

models/base_model.py
```diff
@@ -14,7 +14,11 @@ class BaseModel(db.Model):
 
     @classmethod
     def find(cls, id):
-        return cls.query.get_or_404(id, f"{cls._name()} not found")
+        object = cls.query.get(id)
+        if object:
+            return object
+        else:
+            raise InvalidRequest(404, f"{cls._name()} not found")
```
aedwardg, line 21: Just double-checking to make sure I understand this correctly --
We're using a custom `InvalidRequest` Exception here instead of using the flask-sqlalchemy `get_or_404()` method because without flask_restful, the error will be in text rather than json. Is that correct?

aedwardg (dismissed):
Just a quick typo fix and this looks good to go! Making the switch over to the built-in `MethodView`s seems simple enough.

models/base_model.py
NickSchimek: Correct, but I don't like this implementation.

poppa-nick:
@NickSchimek are you still working on this and looking for help? What are your goals here?

NickSchimek:
> @NickSchimek are you still working on this and looking for help? What are your goals here?

Thanks for asking @poppa-nick ! My immediate concern is that flask-restful may not receive regular updates anymore. Although the author of the library does not work on it that much anymore they are still merging PR's and releasing updates. It was just updated to support Flask 2.0

So my main reasoning for moving off of Flask-restful is not much of a concern anymore. The main thing we use flask-restful for is ensuring that all responses are in a json format. MethodView default response format is html, and it doesn't appear to be straightforward to change that. Flask-Restful's default configuration only supports json. Which is very convenient for us since we only ever want to return json to the frontend. As of now that is mainly the only feature we are using in flask-restful. Someday I would like to migrate off of flask-restful, but the one thing it does for us, it does it very well. So until we find or build a better alternative I think it is better and safer to stick with flask-restful for now. If you or anyone else has any ideas then please share them here. In the meantime lets close this issue.

PR #252, opened 2021-04-14
Update JWT packages and code

### LINKED ISSUE #585: Update pyjwt to the latest version
pyjwt's latest 2.0 release has had some breaking changes, and our app needs to be updated to work with it.
See codeforpdx/dwellinglybackend/pull/240 for more information.

aedwardg:

Closes codeforpdx/dwellingly-app/issues/585

Initially I was just going to update PyJWT, but it wouldn't work without also updating Flask-JWT-Extended.
Both packages had a significant amount of breaking changes, but everything should be working correctly after this.

For more information on the changes to both packages, see:
- https://pyjwt.readthedocs.io/en/stable/changelog.html#v2-0-0
- https://flask-jwt-extended.readthedocs.io/en/stable/v4_upgrade_guide/

Any new dependencies to install? YES
Any special requirements to test? Not sure, I don't think so. All tests are passing on my end.

### DIFF
--- a/app.py
+++ b/app.py
@@ -85,7 +85,7 @@ def create_app(env):
     app.mail = Mail(app)
 
     # check the user role in the JSON Web Token (JWT)
-    @app.jwt.user_claims_loader
+    @app.jwt.additional_claims_loader
     def role_loader(identity):
         user = UserModel.find(identity)
         return {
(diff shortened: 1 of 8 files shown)


app.py
```diff
@@ -98,8 +98,8 @@ def role_loader(identity):
 
     # checking if the token's jti (jwt id) is in the set of revoked tokens
     # this check is applied globally (to all routes that require jwt)
-    @app.jwt.token_in_blacklist_loader
-    def check_if_token_in_blacklist(decrypted_token):
+    @app.jwt.token_in_blocklist_loader
+    def check_if_token_in_blacklist(jwt_header, decrypted_token):
```
NickSchimek: Is this correct? Should the a be an o?

NickSchimek (commented):
Thank you. I think there is one typo

app.py
aedwardg: Sorry, that was the old function name and I left it the same when I updated the decorator.
Changed it now.

NickSchimek (approved):
Awesome, thanks for doing this!

PR #253, opened 2021-04-15
Creates add Note endpoint & tests

Protopaco:

Adds new Notes endpoint /tickets/:ticketID/notes for managing notes.
This is the first step in that process, allowing the user to add a note with a POST request.
Removes edit note functionality from edit Ticket endpoint


resources/notes.py
```diff
@@ -0,0 +1,18 @@
...
+    parser.add_argument("text")
+    parser.add_argument("authorID")
+
+    @pm_level_required
+    def post(self, id):
+        data = Notes.parser.parse_args()
+        note_payload = {"ticketid": id, "text": data.text, "userid": data.authorID}
+
+        created_note = NotesModel.create(schema=NotesSchema, payload=note_payload)
+        return created_note.json()
```
NickSchimek: I think this can be changed to:
```python
class Notes(Resource):
    @pm_level_required
    def post(self, id):
        return NotesModel.create(schema=NotesSchema, payload=request.json).json()
```

The frontend code will need to make sure they're sending the right request.

POST("userid": ?, "ticketid": ?, "text": "some text")

As the backend is not responsible for translating request parameters.

Although I'd argue that the frontend shouldn't be sending the "userid", because the backend knows who the current user is and can assign appropriately. Let me know if you'd like to talk more about this.

NickSchimek (changes requested):
Lookin good. We just need to remove the reqparser as  we're moving away from that.

resources/notes.py
Protopaco: I love the idea of pulling the userid directly from the jwt, but I haven't been able to find out how to pull it from the request.  Do you know off the top of your head, or know where I can look to find that information?

Is it redundant to send the ticketid twice?  Inside the POST request as well as in the endpoint?  Would this make sense?

```python
class Notes(Resource):
    @pm_level_required
    def post(self, ticketid):
        return NotesModel.create(schema=NotesSchema, payload={**request.json, ticketid: ticketid}).json()
```

Protopaco: @NickSchimek  No rush, but I was curious if you saw my response to your comment.  I'm excited to get this endpoint finished so I can finish this issue and work on the rest of the Note functionality.  Thanks!

NickSchimek: Sorry for the delayed response. Your code:
```python
class Notes(Resource):
    @pm_level_required
    def post(self, ticketid):
        return NotesModel.create(schema=NotesSchema, payload={**request.json, ticketid: ticketid}).json()
```

Looks correct. good catch. The payload parameter would need to include the `ticket_id`. It is not redundant since I don't think `request.json` should include the `ticket_id` since it is already in the endpoint. Good question. Thanks for catching that.

It looks like we can get the current user by calling `get_jwt_identity()` - you may want to look up that method to confirm.
https://github.com/codeforpdx/dwellinglybackend/blob/acc9a65fbdb9b9d3b43250170f6f30b250f3db57/resources/user.py#L229

Protopaco:
```python
get_jwt_identity()
```

Was exactly what I was looking for.  Thank you for that.  I believe I made all of the necessary changes.  Please let me know if there are any other improvements I can make.  Thanks!

tests/integration/test_notes.py
```diff
@@ -0,0 +1,35 @@
...
+class TestCreate:
+    def setup(self):
+        self.endpoint = "/api/tickets"
+        self.newNote = {"text": "We don't need no water"}
+
+    def test_creates_note_through_endpoint(
+        self, valid_header, create_admin_user, create_tenant, create_ticket
+    ):
+        create_admin_user()
+        create_tenant()
```
NickSchimek: Do we need these for the test?

tests/integration/test_notes.py
```diff
@@ -0,0 +1,35 @@
+import pytest
+from conftest import is_valid
+from models.notes import NotesModel
+
+
+@pytest.mark.usefixtures("client_class", "empty_test_db")
+class TestCreate:
+    def setup(self):
+        self.endpoint = "/api/tickets"
+        self.newNote = {"text": "We don't need no water"}
```
NickSchimek: minor style: The backend uses snake_case. Please use that style everywhere.

tests/integration/test_notes.py
```diff
@@ -0,0 +1,35 @@
...
+        create_tenant()
+        validTicketID = create_ticket().json()["id"]
+
+        response = self.client.post(
+            f"{self.endpoint}/{validTicketID}/notes",
+            json=self.newNote,
+            headers=valid_header,
+        )
+
+        assert response.json == NotesModel.find(response.json["id"]).json()
```
NickSchimek: This assertion seems a bit odd. I don't like using the response on both sides of the equality check. Is there another way we can assert our note was created?

NickSchimek (changes requested):
Looking good. I left you a couple comments in the test.

tests/integration/test_notes.py
Protopaco: It appears we don't.  I didn't realize create_ticket includes adding admin and tenant.  Good catch!

tests/integration/test_notes.py
Protopaco: Fixed, thanks!

tests/integration/test_notes.py
Protopaco: We could hardcode like this but that seems like it will make the test more fragile.  
```python
        assert response.json == NotesModel.find(1).json()
```


I could pull the number from last note added to the ticket.  It is making some assumptions, but it should be pretty safe.
```python
        note_id = TicketModel.find(valid_ticket_id).json()["notes"][-1]["id"]
        assert response.json == NotesModel.find(note_id).json()
```

Does one of these solutions work better?  I'll keep thinking about this, see if I can come up with a different solution.

NickSchimek: I'd go with your second approach. Although, we already have the ticket object so there is not need to ask the database for it. 

Your test could be a bit more concise and look something like this:
```python
        ticket = create_ticket()

        response = self.client.post(
            f"{self.endpoint}/{ticket.id}/notes",
            json=self.payload,
            headers=valid_header,
        )

        assert response.json == ticket.notes[-1].json()
```

Protopaco: That makes sense.  I made those changes and pushed a new commit.

NickSchimek (approved):
This looks good. Thanks for all the work you put in.

tests/integration/test_notes.py
NickSchimek: Technically we're still asking the db for the notes, but doing so through the association. Just a bit more cleaner and easier to read.

PR #254, opened 2021-04-19
Refactor Property resource

NickSchimek:
- Test property json method
    - This adds test for serializing a property object. It has some breaking
changes as the field names were changed to plural when serializing a
list.
- Add ability to serialize tenants with property.json()
- Refactor property resource
- Refactor tests

Includes partial fixes for: codeforpdx/dwellingly-app/issues/403

### DIFF
--- a/tests/integration/test_properties.py
+++ b/tests/integration/test_properties.py
@@ -1,8 +1,49 @@
 import pytest
 import json
+from unittest.mock import patch
+from schemas.property import PropertySchema
 from models.property import PropertyModel
-from models.user import UserModel
-from models.tenant import TenantModel
+
+
+@pytest.mark.usefixtures("client_class", "empty_test_db")
+class TestProperty:
+    def setup(self):
+        self.endpoint = "/api/properties/"
+
+    def test_get(self, valid_header, create_property):
+        property = create_property()
+        response = self.client.get(
+            f"{self.endpoint}{property.id}", headers=valid_header
+        )
+
+        assert response.json == property.json(include_tenants=True)
+
+    def test_delete(self, valid_header, create_property):
+        with patch.object(PropertyModel, "delete") as mock_delete:
+            response = self.client.delete(f"{self.endpoint}1", headers=valid_header)
+
+        mock_delete.assert_called_once_with(1)
+        assert response.status_code == 200
+        assert response.json == {"message": "Property deleted"}
+
+    def test_put(self, valid_header, create_property):
+        property = create_property()
+
+        with patch.object(
+            PropertyModel, "update", return_value=property
+        ) as mock_update:
+            response = self.client.put(
+                f"{self.endpoint}1", json={"num_units": 2}, headers=valid_header
+            )
+
+        mock_update.assert_called_once_with(
+            schema=PropertySchema,
+            id=1,
+            context={"name": property.name},
+            payload={"num_units": 2},
+        )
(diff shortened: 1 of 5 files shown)


tests/integration/test_properties.py
```diff
@@ -1,8 +1,49 @@
...
+            )
+
+        mock_update.assert_called_once_with(
+            schema=PropertySchema,
+            id=1,
+            context={"name": property.name},
+            payload={"num_units": 2},
+        )
+        assert response.status_code == 200
+        assert response.json == property.json()
```
aedwardg, line 46: I'm assuming the reason you're mocking the `PropertyModel` here is because the update functionality is already tested in the `BaseModel` tests and `PropertySchema` tests.

That said, what's the point in the last line?
```python
assert response.json == property.json()
```

What is that testing? I'm just trying to understand, since `response.json` won't reflect the updated property `num_units` here.

aedwardg (commented):
This looks good, I just have the one question below about the PUT test.

tests/integration/test_properties.py
NickSchimek: Yeah, I have a somewhat love/hate relationship with testing PUT/PATCH methods.
... EDIT: Maybe its just a hate relationship

I want to test the behavior, so to define the behavior I might say. 
Upon updating a property it returns a successful response with serialized data.

So the reason why I test
`assert response.json == property.json()` is because I want to ensure the response is returning serialized property data.

So that is what that is testing. As for your second question. The property was never updated, because I mocked that method out so I cannot test that.

This is more of a unit test rather than an integration/system test. I tend to hate testing methods that update resources, because of the number of different arguments it can have. It's usually not feasible to exhaustively test an update method.

I almost didn't mock this out, but I couldn't get myself to test that when I pass in 2 as the value for num_units then the database will update that column to 2 and I assert that the response has 2 for num_units. And not only that, I would then need to ensure I submit valid data to test, because these will be ran through validations before the database is updated. If I don't mock this and some point in the future we add a validation that a property cannot have 2 units then this test will fail. All I really care about is that when this method is invoked. It calls the update method correctly, and with the correct arguments. 

On line 36, I'm passing in `json={"num_units": 2}` to the put request, and I'm asserting that the update action is being called with the correct paramters on lines 40-43.

I assert 200 success
and I assert the response is serialized. The response is what I told the mock to return which is on line 33. `return_value=property`

So its a bit weird that this method updates a resource, but we're not actually testing that anything is updated, because I have a high level of confidence that if the request to update the object is sent correctly then our object will be updated, and I want to make sure that we are serializing the object we get back from the update method which is what line 46 does. I could've returned a user object and assert that a user object is serialized, and that still would be a good test, it would be a confusing test, and would fail if we deleted the user class in the future, but it would still be a "somewhat" correct test.

aedwardg: > Yeah, I have a somewhat love/hate relationship with testing PUT/PATCH methods.
> ... EDIT: Maybe its just a hate relationship

🤣 I feel you there.

> we're not actually testing that anything is updated, because I have a high level of confidence that if the request to update the object is sent correctly then our object will be updated, and I want to make sure that we are serializing the object we get back from the update method which is what line 46 does.

Thanks for the thorough explanation, I think I understand your reasoning now!

aedwardg (dismissed):
Looks good to me! 🚀

PR #257, opened 2021-04-22
Move ticket auth tests

### LINKED ISSUE #545: Separate authorization tests from endpoint logic tests

gussmith23:


tests/authorizations/test_ticket_authorization.py
```diff
@@ -0,0 +1,15 @@
...
+class TestTicketAuthorization:
+    def setup(self):
+        self.endpoint = "/api/tickets"
+        self.validID = 1
+
+    def test_tickets_DELETE(self):
+        response = self.client.delete(f"{self.endpoint}/{self.validID}")
+        # UNAUTHORIZED - Missing Authorization Header
+        assert is_valid(response, 401)
+        assert response.json == {"message": "Missing authorization header"}
```
aedwardg, line 15: If you're interested in improving this test a little, you could update it to mock the `TicketModel` with unittest.mock and use `.assert_not_called()` similar to how Nick did in his recent PR:
https://github.com/codeforpdx/dwellinglybackend/blob/73e929c7345e4b7860500145734d17f5ad244729/tests/integration/test_properties.py#L21-L27

but since this PR is just for moving the tests over, that's totally optional 🙂

aedwardg (dismissed):
Thanks @gussmith23!
Looks like that was the only authorization test we had for tickets, so I'll go ahead and approve this.

It does feel a little weird to me to be calling something `validID` though, when it's actually not a valid ID (there are no tickets yet in the empty database).  If you have a moment and want to update this to just use `/1`  and get rid of `self.validID` that would probably make more sense.

tests/authorizations/test_ticket_authorization.py
NickSchimek: Did I use assert not called? I'd always be hesitant using that since there is a much wider range for which it will pass than fail.

I think this test is fine. In the future, if feasible, I would like to improve the auth tests, I will get a ticket created for what should be done, and maybe someone smarter than me can figure out how to do it.

aedwardg: No, you used `.assert_called_once_with()`, but in this case the delete method wouldn't be called since there is no valid header, so I figured that was the way to go here. That makes sense though.

NickSchimek (dismissed):
Thanks Gus!

PR #260, opened 2021-04-29
551 unit test users w no role

### LINKED ISSUE #551: Create test for the unauthorized-user query method in User model
## Description

The `find_users_without_assigned_role` method defined in the UserModel should have a unit test (`tests/unit/test_user.py`) that checks the behavior of the method. If a variety of authorized and unauthorized users both active and inactive existed in the db. Does this query return what we would expect it to?

I think this could be done in two different ways. Depending which way you choose is up to you, or if you have another preferred way then that is fine too. As long as it covers the four different possibilities.

This could be done with one test where we create a bunch of objects and check that only the "pending" ones are returned.
OR
This could be done with multiple tests where each test checks that if the object created (user) is returned or not returned.

Possibly create a new class in the `tests/unit/test_user.py` file to define this test inside, which can also be used to put all future UserModel tests inside. Name the class `TestUser` and name the test method after the method your testing. `def test_ find_users_without_assigned_role`.
OR
If you decide to create more than one test to check this method then perhaps name the class `TestFindUsersWithoutAssignedRole` and put all the test methods for that method inside that class.

```
class TestUser:
    def test_ find_users_without_assigned_role
        ...
```
OR
```
class TestFindUsersWithoutAssignedRole:
    def test_a_condition
        ...
    def test_another_condition
        ...
    def test_a_condition
        ...
    def test_etc
        ...
```

I think depending on which way you choose to write the test then one of the two options above will help keep the code organized and easy to read/understand what is going on. I'm open to other suggestions as well. Try them both and see which one you like better, or come up with something different. My main concern is that the code is easy to parse, read, and figure out what this method is doing and give me the confidence that it is correct. One note: I prefer not to have comments in code, and I think one of the suggestions above will not require comments in order to tell the story of what is going on with the `find_users_without_assigned_role` method. Although that method name tells the story pretty well I think 😄 But the test should build our confidence that it is doing what it says it does.

ovsyankimo:
Closes codeforpdx/dwellingly-app/issues/551

### DIFF
--- a/tests/unit/test_user.py
+++ b/tests/unit/test_user.py
@@ -29,34 +29,32 @@ def test_full_name(empty_test_db, create_admin_user):
 @pytest.mark.usefixtures("empty_test_db")
 class TestFindUsersWithoutAssignedRole:
     def test_find_users_authorized_active(self, create_admin_user):
-        user = create_admin_user(archived=False)
-        assert user.role == RoleEnum.ADMIN
-        assert user.archived is False
+        _ = create_admin_user()
 
         found_user = UserModel.find_users_without_assigned_role()
-        # TODO do I need to wipe the DB out after each test?
-        assert found_user
+        assert found_user.first() is None
 
     def test_find_users_authorized_inactive(self, create_admin_user):
-        user = create_admin_user(archived=True, firstName="Sara")
-        assert user.role == RoleEnum.ADMIN
-        assert user.archived is True
+        user = create_admin_user()
+        user.archived = True
+        user.save_to_db()
 
         found_user = UserModel.find_users_without_assigned_role()
-        assert found_user
+        assert found_user.first() is None
 
     def test_find_users_unauthorized_active(self, create_unauthorized_user):
-        user = create_unauthorized_user(archived=False)
-        assert user.role is None
+        user = create_unauthorized_user()
 
         found_user = UserModel.find_users_without_assigned_role()
-        assert found_user
+        assert found_user.first() == user
 
     def test_find_users_unauthorized_inactive(self, create_unauthorized_user):
-        _ = create_unauthorized_user(archived=True)
+        user = create_unauthorized_user()
+        user.archived = True
+        user.save_to_db()
 
         found_user = UserModel.find_users_without_assigned_role()
-        assert found_user
+        assert found_user.first() is None
 
 
 @pytest.mark.usefixtures("empty_test_db")
--- a/tests/unit/test_user.py
+++ b/tests/unit/test_user.py
@@ -26,6 +26,21 @@ def test_full_name(empty_test_db, create_admin_user):
     assert admin.full_name() == "first last"
 
 
+###################################################################################
+# TODO this is what I'm working on
+# def test_find_users_without_assigned_role(
+#    empty_test_db, find_users_without_assigned_role
+# ):
+# user = UserModel.find_by_id(1)
+# users = [user.json() for user in UserModel.find_users_without_assigned_role()]
+# stuff = UserModel.find_users_without_assigned_role()
+# thing_returned = find_users_without_assigned_role()
+
+# TODO 4 cases we want to test:
+
+# TODO what is hte behavior of the method & how do I want ot test it?
+
+
 @pytest.mark.usefixtures("empty_test_db")
 class TestFixtures:
     def test_create_admin_user(self, create_admin_user):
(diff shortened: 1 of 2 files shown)


aedwardg (approved):
Thanks @ovsyankimo, this looks great!

The only thing I might change is making the test function names a little more descriptive of what we should expect from the tests, but that's no biggie. Perhaps if we come up with better names down the line we can update them, but right now this looks good to go.

tests/unit/test_user.py
```diff
@@ -26,6 +26,37 @@ def test_full_name(empty_test_db, create_admin_user):
     assert admin.full_name() == "first last"
 
 
+@pytest.mark.usefixtures("empty_test_db")
+class TestFindUsersWithoutAssignedRole:
+    def test_find_users_authorized_active(self, create_admin_user):
+        _ = create_admin_user()
```
NickSchimek, line 32: Do we need to assign this? If we're not going to use the assignment, can we simply do `create_admin_user()`?

tests/unit/test_user.py
```diff
@@ -26,6 +26,37 @@ def test_full_name(empty_test_db, create_admin_user):
...
+
+        found_user = UserModel.find_users_without_assigned_role()
+        assert found_user.first() is None
+
+    def test_find_users_authorized_inactive(self, create_admin_user):
+        user = create_admin_user()
+        user.archived = True
+        user.save_to_db()
```
NickSchimek, line 40: I wonder, I think the factories are setup to take keyword arguments.

I believe we can remove all this and replace it with `create_admin_user(archived=True)`

Well, I tested this theory out, and while it is setup to take keyword arguments. archived doesn't work as expected, but I think it would be ok to change that.

tests/unit/test_user.py
```diff
@@ -26,6 +26,37 @@ def test_full_name(empty_test_db, create_admin_user):
...
+
+        found_user = UserModel.find_users_without_assigned_role()
+        assert found_user.first() is None
+
+    def test_find_users_unauthorized_active(self, create_unauthorized_user):
+        user = create_unauthorized_user()
+
+        found_user = UserModel.find_users_without_assigned_role()
+        assert found_user.first() == user
```
NickSchimek, line 49: I didn't know about this `first()` method. That is pretty cool. And there shouldn't be a second result, because we only created one user. Hmmm. It looks like there is a `.all()` method - https://flask-sqlalchemy.palletsprojects.com/en/2.x/queries/?highlight=first - I would prefer to use the .all() method in all of the tests.
`assert result.all() == [user]`
and instead of testing for `None` we test the query is empty.
`assert result.all() == []`
Because the query doesn't ever return `None` so I don't think we should be testing for it.

NickSchimek (commented):
This is good. It tests all the right things 👍 I did find it a bit hard to read, and found myself being a bit confused even though I was familiar with the context and what it was doing. I find naming things and naming test methods to be one of the hardest things.

I think moving the positive test case up to the top would improve it a lot. Because then the reader instantly sees what the method does rather than what it doesn't do.

And then perhaps renaming the other tests. For example: the first test: `test_find_users_authorized_active` I was expecting it to find users that were authorized and active, but the result was `None`. I think if it was renamed to be `test_it_does_not_contain_active_users_that_have_been_authorized` would make it a lot easier to understand what the subsequent code should do. I prefer to be concise, but sometimes verbosity is better.

tests/unit/test_user.py
NickSchimek: I was going to say if you wanted to include this in this PR than that would be OK, but I think it would be best to create a separate issue for it. I will get an issue created after this PR is merged.

aedwardg: Yeah, @ovsyankimo  and I already tried altering the fixture to do that.  
However, even when you do, `archived` stays as `False` because the `UserModel` has `self.archived=False` in the init method.
https://github.com/codeforpdx/dwellinglybackend/blob/b02379dffaa64b6352e93e1d45275bd81afa263b/models/user.py#L48-L51

I think that can be removed from the init, but I haven't gotten around to digging through the code to see if there was a reason for it to be there in the first place.  I have the feeling it is just a remnant from before we added `default=False` to the `archived` field.

NickSchimek:
Yep, I remember when that was added. I think it can be removed without
issue, if you want to do that in this pr, I think that would be fine, but
if removing it causes other tests to fail then let’s do it separately.

tests/unit/test_user.py
NickSchimek: Yep, I remember when that was added. I think it can be removed without issue, if you want to do that in this pr, I think that would be fine, but if removing it causes other tests to fail then let’s do it separately.

PR #271, opened 2021-05-03
Remove tenant endpont DELETE action

### LINKED ISSUE #591: Fix tenant archiving logic
**Purpose**
Archiving a tenant should NOT be on a toggle switch. When the frontend hits the API with a request it should be deterministic. Currently the backend logic is written in a way that it expects the frontend to know the state of the backend. The update (PUT) action could handle this, but then the return message will be different.

To keep the response the same the delete method should be deleted/modified and replaced with deterministic logic. I believe there are a few different ways this could be done, but basically we want the frontend to be able to make a request and know for certain that they are archiving a tenant even if the tenant is already archived, as it is now they could inadvertently unarchive a tenant when they want to archive a tenant.

**Proposed Solution**
Whoever picks this up, Let me know what implementation you think is a good option and we can discuss, or if you need me to provide a couple different options then I can do that too.

aedwardg:

Closes codeforpdx/dwellingly-app/issues/591

Removed the `delete()` method from this resource and refactored the
update tenant endpoint test.

The `delete()` method was being used for archiving tenants through a
"toggle" mechanism that would archive the tenant if they were active and
unarchive them if they were already archived.

This required the front end to be aware of a tenant's archival status
before making the request, or else risk accidentally doing the opposite
of what was intended.

Rather than risk this and write additional logic for custom
messages on the back end, we decided to switch to a more deterministic
(and simple) approach by using the existing PUT action to handle
archival.

Success messages can be handled by the front end (as with all update
actions) and the built-in error messages are returned by the back end.

### DIFF
--- a/resources/tenants.py
+++ b/resources/tenants.py
@@ -58,21 +58,3 @@ class Tenants(Resource):
         return TenantModel.update(
             schema=TenantSchema, payload=request.json, id=tenant_id
         ).json()
-
-    @admin_required
-    def delete(self, tenant_id):
-        tenant = TenantModel.find(tenant_id)
-
-        def _toggle_archive():
-            if tenant.archived:
-                tenant.archived = False
-                status = {"message": "Tenant unarchived"}
-            else:
-                tenant.archived = True
-                status = {"message": "Tenant archived"}
-
-            tenant.save_to_db()
-            return status
-
-        response = _toggle_archive()
-        return response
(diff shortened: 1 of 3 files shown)


aedwardg:
@NickSchimek Let me know if you want me to put an actual delete method back in here for permanently deleting from the database.  I know we're planning on just archiving the tenants, but the Property and User endpoints both currently support permanent deletions in addition to archival.

NickSchimek:
> @NickSchimek Let me know if you want me to put an actual delete method back in here for permanently deleting from the database. I know we're planning on just archiving the tenants, but the Property and User endpoints both currently support permanent deletions in addition to archival.

No, We don't have a reason to have it. so lets not support it.

NickSchimek (dismissed):
Looks good. Thanks!

PR #272, opened 2021-05-03
Separate tenant resource out into two classes

### LINKED ISSUE #592: Separate Tenant resource out into two separate classes
**Purpose**
Endpoints that require an id should be in the class Tenant For example the update and get action
Endpoints that do not require an id should be in the class Tenants for example the POST and GET action

See the app.py file for other resources that follow this style. The actual routes and behavior will not change. Only the code organization. So this is more of a refactor which will remove the conditional in the get action when its done.

NickSchimek:
Separated tenant resource out into two classes

Closes codeforpdx/dwellingly-app/issues/592


aedwardg (dismissed):
This looks good to me! 👍

PR #273, opened 2021-05-03
Create json method for use with collections

NickSchimek:
Created json method for use with collections.

### DIFF
--- a/db.py
+++ b/db.py
@@ -1,3 +1,4 @@
 from flask_sqlalchemy import SQLAlchemy
+from utils.nobiru import NobiruQuery
 
-db = SQLAlchemy()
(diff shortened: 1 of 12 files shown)


aedwardg (dismissed):
@NickSchimek,

This is pretty cool, I never would have thought to do this.  I'm not sure it makes understanding the codebase any easier for newer contributors, but it certainly handles the model serialization in a more elegant way.

I'd be okay with merging this in, but I do have one question:  
Does this mean we're completely abandoning the thought of using Marshmallow for serialization down the road?  I know that was always a "maybe", but it seems even less likely if we're investing more into these `json` methods.

PR #275, opened 2021-05-06
adds delete notes endpoint and tests

### LINKED ISSUE #600: Creates delete Note endpoint & tests
**Purpose**
Adds new Notes endpoint /tickets/:id/notes for deleting notes.

**Proposed Solution**

Using new backend code style endpoint should confirm that the note's userid is the same at the request's userid.  Using the NotesModel delete the note based on the id passed.  Create a class test on empty database to ensure notes are deleting successfully.

**Acceptance Criteria**
- [ ] Create DELETE /tickets/:id/notes endpoint
- [ ] Uses NotesModel to delete ticket from database
- [ ] Tests that note was deleted properly

Protopaco:
Closes codeforpdx/dwellingly-app/issues/600


resources/notes.py
```diff
@@ -19,3 +19,8 @@ def post(self, id):
                 "userid": get_jwt_identity(),
             },
         ).json()
+
+    @pm_level_required
+    def delete(self, id):
+        TicketModel.delete(id)
```
aedwardg: Right here, you're actually deleting the Ticket, not the Note.  You'll need to update this to delete the note instead.

tests/integration/test_notes.py
```diff
@@ -28,3 +29,19 @@ def test_it_returns_404_with_invalid_ticket(self, valid_header):
...
         )
         assert is_valid(response, 404)  # Bad Request- 'Invalid Ticket'
+
+    def test_it_deletes_one_note(
+        self, valid_header, create_note, create_admin_user, create_ticket
+    ):
+        note = create_note(create_admin_user(), create_ticket())
+
+        response = self.client.delete(
+            f"{self.endpoint}/{note.id}/notes", headers=valid_header
```
aedwardg: Because we need to access the Note and not the Ticket, this should say `f"{self.endpoint}/{note.ticketid}/notes/{note.id}`

tests/integration/test_notes.py
```diff
@@ -28,3 +29,19 @@ def test_it_returns_404_with_invalid_ticket(self, valid_header):
...
+        )
+
+        assert response.json == {"message": "Note deleted"}
+
+        with pytest.raises(Exception) as e_info:
+            NotesModel.find(note.id)
+
+        assert "404 Not Found: Notes not found" in str(e_info.value)
```
aedwardg: I'm not so sure that we need to test the sad path here -- the 404 response is handled by the inherited `delete()` and `find()` methods in the Base Model and these are already tested in `base_interface_test.py`

Perhaps @NickSchimek has a different preference though.

aedwardg (changes requested):
@Protopaco, thanks for taking this on, this is a good start.

However, there are a couple of things that need updating.
For starters, since we have to access an individual note to delete it, the route for this one should be updated to `tickets/<int:ticket_id>/notes/<int:id>` as mentioned in this comment:
https://github.com/codeforpdx/dwellingly-app/issues/600#issuecomment-831952413

This can go on the same line of `app.py` as the current route: 
https://github.com/codeforpdx/dwellinglybackend/blob/62677dbf1f9dd1be9722f2c0cc1f7c192443e7ff/app.py#L67-L68
just put a comma before the new string (check out the tenants, emergency contacts and reset password routes to see what I mean).

There are also a few related changes in the comments below

resources/notes.py
Protopaco: I can't believe I did that.  Thank you.

aedwardg: 😂 all good, I have moments like that all the time!

tests/integration/test_notes.py
Protopaco: I guess I didn't think this all of the way through, I'm sorry.  This is going to change the endpoint for the POST as well as the DELETE, since the .POST is only accepting the ticket_id.  How do we feel about changing the endpoint in app.py to:

```python
    api.add_resource(Note, "tickets/<int:ticket_id>/notes", "tickets/<int:ticket_id>/notes/<int:note_id>")
```

and then the test code to: 

```python
            f"{self.endpoint}/{note.ticketid}/notes/{note.id}", headers=valid_header
```

as Anthony suggests.

It is sending an unnecessary parameter "ticket_id" but it is adhering closer to the existing style.

tests/integration/test_notes.py
Protopaco: I like to think of it as the "successful" path rather than the "sad" path ;) , but i have no preference on this one.

app.py
```diff
@@ -64,7 +64,11 @@ def create_routes(app):
     api.add_resource(Leases, "lease")
     api.add_resource(Tickets, "tickets")
     api.add_resource(Ticket, "tickets/<int:id>")
-    api.add_resource(Note, "tickets/<int:id>/notes")
+    api.add_resource(
+        Note,
+        "tickets/<int:ticket_id>/notes",
+        "tickets/<int:ticket_id>/notes/<int:note_id>",
+    )
```
NickSchimek: This should be the following:
```suggestion
    api.add_resource(Notes, "tickets/<int:ticket_id>/notes")
    api.add_resource(Note,   "tickets/<int:ticket_id>/notes/<int:id>")
```

Note class acts on a single Note resource.
Notes class acts when a single Note resource is not provided. Such as retrieving multiple notes, or creating a note.

resources/notes.py
```diff
@@ -9,13 +9,18 @@
 
 class Note(Resource):
```
NickSchimek: Change this to `class Notes`

resources/notes.py
```diff
@@ -9,13 +9,18 @@
...
-        TicketModel.find(id)
+    def post(self, ticket_id):
+        TicketModel.find(ticket_id)
         return NotesModel.create(
             schema=NotesSchema,
             payload={
                 "text": request.json["text"],
-                "ticketid": id,
+                "ticketid": ticket_id,
                 "userid": get_jwt_identity(),
```
NickSchimek: changes here should not be needed.

resources/notes.py
```diff
@@ -9,13 +9,18 @@
...
             },
         ).json()
+
+    @pm_level_required
+    def delete(self, ticket_id, note_id):
+        NotesModel.delete(note_id)
+        return {"message": "Note deleted"}
```
NickSchimek: This needs to be in a class. `class Note`
Let's replace `note_id` with `id` - It's a small style thing, but it's how the rest of the app is doing so lets stay consistent

tests/integration/test_notes.py
```diff
@@ -1,9 +1,10 @@
 import pytest
 from conftest import is_valid
+from models.notes import NotesModel
 
 
 @pytest.mark.usefixtures("client_class", "empty_test_db")
-class TestCreate:
+class TestNote:
```
NickSchimek: This should still be `TestCreate`
Let's create a new class for each action. IMO it is easier to read when organized that way.

So we will have `TestCreate` class and `TestDelete` class

tests/integration/test_notes.py
```diff
@@ -28,3 +29,19 @@ def test_it_returns_404_with_invalid_ticket(self, valid_header):
...
+        response = self.client.delete(
+            f"{self.endpoint}/{note.ticketid}/notes/{note.id}", headers=valid_header
+        )
+
+        assert response.json == {"message": "Note deleted"}
+
+        with pytest.raises(Exception) as e_info:
+            NotesModel.find(note.id)
+
+        assert "404 Not Found: Notes not found" in str(e_info.value)
```
NickSchimek: This is testing behavior in the BaseModel which is already thoroughly tested. Let's remove this test.

tests/integration/test_notes.py
```diff
@@ -28,3 +29,19 @@ def test_it_returns_404_with_invalid_ticket(self, valid_header):
             headers=valid_header,
         )
         assert is_valid(response, 404)  # Bad Request- 'Invalid Ticket'
+
+    def test_it_deletes_one_note(
+        self, valid_header, create_note, create_admin_user, create_ticket
+    ):
+        note = create_note(create_admin_user(), create_ticket())
```
NickSchimek: I didn't know we had to pass in arguments to create_note. We'll have to get this updated. Issue codeforpdx/dwellingly-app/issues/639 created to address this.

tests/integration/test_notes.py
```diff
@@ -28,3 +29,19 @@ def test_it_returns_404_with_invalid_ticket(self, valid_header):
...
+
+        response = self.client.delete(
+            f"{self.endpoint}/{note.ticketid}/notes/{note.id}", headers=valid_header
+        )
+
+        assert response.json == {"message": "Note deleted"}
```
NickSchimek: FYI: If I were to remove the following line from the Note endpoint under test.
`NotesModel.delete(note_id)`
Then this test would still pass. We want to test behavior here. Behavior could be stated as follows:
> The delete endpoint removes a Note resource, identified by the provided id, from the database.

The preferred way to test this is to test that the `NotesModel.delete` method is being called with the correct arguments. 

Another way to test this is to check if the created Note no longer exists in the database, but that's technically retesting behavior that is already well tested.

Protopaco: I understand what you are saying, and it makes sense.  The way it is currently setup, this essentially tests that no errors are thrown, rather than NotesModel.delete is called successfully.  That being said, I'm not sure how to test NotesModel.delete when it doesn't return anything, and without testing that the note can no longer be found.

We could change the BaseModel.delete to return a success message, though I'm not sure what the implications of that would be for the rest of the models.  

Is there better way to test this?

NickSchimek: Here is a code example of how to test a method gets called correctly.
https://github.com/codeforpdx/dwellinglybackend/blob/273521e410b0a3c15359b8a9a2e0623d84df712d/tests/integration/test_leases.py#L52-L58

Note: You can literally pass any value as the argument, but I'd probably prefer an actual note id created through a factory is used. Which this example doesn't do.

We're basically stubbing out the model and database here. We know the delete method works, and we know the database works. We just want to assert the endpoint calls the delete method with the correct argument. And I believe we do have a true system test somewhere just in case anything does change with the model/database in the future, so I'm pretty comfortable with this level of testing and nothing more is needed here.

Protopaco: I reworked the test based on test_leases.   Does this look better?

 I used the create_note factory.  For that reason we will want to remove the arguments being passed when #283 is merged in.

NickSchimek: yep, Looks great. I like this because now we'll know if anything changes in the endpoint that causes the delete method to not get called.

NickSchimek (commented):
Thanks @Protopaco - This looks good, and does everything I asked of it.

After taking one more look I noticed a couple red flags. If I remember correctly, the notes requirements are not exactly that clear. And the current logic will let any authenticated user to delete any note in the system. I think this is too trusting. We should either change the delete endpoint to be admin authorization only or change it so that the current_user can only delete notes they created until we can get this out in front of real users and get more feedback on. I think there should also be some way to ensure that the note being deleted belongs to the ticket being passed in. Although I'd say that is a lower priority then the other issue mentioned. I think we should probably get some tickets created to address these issues before merging this in. 

The two questions that need to be answered are:
Who has authorization to delete a note?
Should we ensure the note belongs to the ticket or a ticket? <- Not so important, but depending on the answer we may want to change some things.

Protopaco:
@NickSchimek I'm glad to hear its working correctly.  

 I tend to agree with you that it seems overly trusting to allow any user to delete any note.   We can talk about what direction we want to go with it at Wednesday's meeting, if that makes sense.

NickSchimek:
@Protopaco Can we update this to use the collection delete method?

tests/integration/test_notes.py
NickSchimek: Thanks for updating the factory, can we update this too?

Protopaco:
@NickSchimek Sorry, I spaced this.  Does this look better?

tests/integration/test_notes.py
```diff
@@ -26,4 +28,26 @@ def test_it_returns_404_with_invalid_ticket(self, valid_header):
...
+@pytest.mark.usefixtures("client_class", "empty_test_db")
+class TestDelete:
+    def setup(self):
+        self.endpoint = "/api/tickets"
+
+    def test_it_deletes_one_note(
+        self, valid_header, create_note, create_admin_user, create_ticket
+    ):
+        note = create_note()
+        ticket = TicketModel.find(note.ticket_id)
```
NickSchimek:
```suggestion
        ticket = note.ticket
```

NickSchimek (dismissed):
Perfect this looks good. Thanks for being patient with this. I have one suggestion on the test

Protopaco:
@NickSchimek No problem, it's fun to build something that is so precise.

NickSchimek (approved):

PR #276, opened 2021-05-19
Update ticket model & schema

### LINKED ISSUE #614: Update ticket model/schema
The ticket model accepts an `assignedUserID`, `senderID`, and a `notelog` along with other fields. These two fields `assignedUserID`, and `notelog` can be removed from the model as they're not needed. The note model contains any notes for the ticket and it is not clear what the notelog is to contain when we have a table dedicated for notes. The `assignedUserID` can be removed, because tenants are already assigned to staff. Hence, the staff responsible for the ticket will be the same staff as the ticket.tenant assignment. In other words, a ticket is assigned to staff through the tenant. The senderID can be better named to `creator_id`, which would be the current_user at the time the ticket was created. 

The ticket model will need to have the following fields:
```
    tenant_id
    issue
    urgency ( Currently a string, but looks like it should be an enum. Keep it as a string though unless confirmed by FE )
    status ( This is an enum and should default to the active/open status )
    creator_id ( current_user )
```

 - tasks:
Update the ticket model to include the following fields above.
Update the schema as necessary
The json method will need to be updated necessary
May need to update the ticket factories and seed file.

aedwardg:
## Description

Updates the ticket model and schema.
Renames db fields for clarity and removes `noteslog` and `assignedUserID`.
The assigned users will now be the staff members who are assigned to the ticket's tenant.
Rather than returning a user id, the `json` method will now return a list json objects representing all assigned staff members.

@NickSchimek let me know if I should add some sort of validation to make sure the `creator_id` is the current user.  Also note that the ticket integration tests are pretty bad.  These should probably be addressed in another issue.

Closes codeforpdx/dwellingly-app/issues/614

_NOTE: DB needs to be re-seeded for this._

### DIFF
--- a/models/tickets.py
+++ b/models/tickets.py
@@ -17,12 +17,10 @@ class TicketModel(BaseModel):
 
     id = db.Column(db.Integer, primary_key=True)
     issue = db.Column(db.String(144))
-    tenantID = db.Column(db.Integer, db.ForeignKey("tenants.id"))
-    assignedUserID = db.Column(db.Integer, db.ForeignKey("users.id"))
-    senderID = db.Column(db.Integer, db.ForeignKey("users.id"))
-    status = db.Column(db.Enum(TicketStatus))
+    tenant_id = db.Column(db.Integer, db.ForeignKey("tenants.id"))
+    creator_id = db.Column(db.Integer, db.ForeignKey("users.id"))
+    status = db.Column(db.Enum(TicketStatus), default=TicketStatus.New)
     urgency = db.Column(db.String(12))
-    notelog = db.Column(db.Text)
 
     notes = db.relationship(
         "NotesModel",
@@ -39,11 +37,10 @@ class TicketModel(BaseModel):
             "id": self.id,
             "issue": self.issue,
             "tenant": "{} {}".format(self.tenant.firstName, self.tenant.lastName),
-            "senderID": self.senderID,
-            "tenantID": self.tenantID,
-            "assignedUserID": self.assignedUserID,
+            "creator_id": self.creator_id,
+            "tenant_id": self.tenant_id,
+            "assigned_staff": self.tenant.staff.json(),
             "sender": self.author.full_name(),
-            "assigned": self.assignee.full_name(),
             "status": self.status,
             "minsPastUpdate": minsPastUpdate,
             "urgency": self.urgency,
--- a/models/tickets.py
+++ b/models/tickets.py
@@ -17,9 +17,9 @@ class TicketModel(BaseModel):
 
     id = db.Column(db.Integer, primary_key=True)
     issue = db.Column(db.String(144))
-    tenant_id = db.Column(db.Integer, db.ForeignKey("tenants.id"))
-    creator_id = db.Column(db.Integer, db.ForeignKey("users.id"))
-    status = db.Column(db.Enum(TicketStatus), default=TicketStatus.New)
+    tenant_id = db.Column(db.Integer, db.ForeignKey("tenants.id"), nullable=False)
+    author_id = db.Column(db.Integer, db.ForeignKey("users.id"), nullable=False)
+    status = db.Column(db.Enum(TicketStatus), default=TicketStatus.New, nullable=False)
     urgency = db.Column(db.String(12))
 
     notes = db.relationship(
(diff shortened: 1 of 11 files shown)


models/tickets.py
```diff
@@ -17,12 +17,10 @@ class TicketModel(BaseModel):
...
     id = db.Column(db.Integer, primary_key=True)
     issue = db.Column(db.String(144))
-    tenantID = db.Column(db.Integer, db.ForeignKey("tenants.id"))
-    assignedUserID = db.Column(db.Integer, db.ForeignKey("users.id"))
-    senderID = db.Column(db.Integer, db.ForeignKey("users.id"))
-    status = db.Column(db.Enum(TicketStatus))
+    tenant_id = db.Column(db.Integer, db.ForeignKey("tenants.id"))
+    creator_id = db.Column(db.Integer, db.ForeignKey("users.id"))
+    status = db.Column(db.Enum(TicketStatus), default=TicketStatus.New)
     urgency = db.Column(db.String(12))
```
NickSchimek: I'm not sure if I mentioned it, but we would like all of these fields to be required. Since we're using marshmallow's auto schema we can add the `nullable=False` to achieve that in the validations and at the database level.

models/user.py
```diff
@@ -48,16 +48,11 @@ class UserModel(BaseModel):
...
-        backref="assignee",
-        primaryjoin=id == TicketModel.assignedUserID,
-        collection_class=NobiruList,
-    )
+
     authored_tickets = db.relationship(
         "TicketModel",
         backref="author",
-        primaryjoin=id == TicketModel.senderID,
+        primaryjoin=id == TicketModel.creator_id,
```
NickSchimek: I wasn't aware we were using the term `author` here. Instead of using `creator_id` lets change that to `author_id` to be consistent with the terminology already being used.

schemas/ticket.py
```diff
@@ -12,25 +12,17 @@ class Meta:
         include_fk = True
 
     tenant = fields.Nested("TenantSchema")
-    user = fields.Nested("UserSchema")
+    author = fields.Nested("UserSchema")
```
NickSchimek, line 15: I forget what this does. Is this for serialization? Do we need it? Should we remove it?

tests/factory_fixtures/ticket.py
```diff
@@ -19,16 +18,12 @@ def _ticket_attributes(issue, tenant, assignedUser, sender):
...
-):
+def create_ticket(faker, ticket_attributes, create_tenant, create_join_staff):
     def _create_ticket(issue=None):
         if not issue:
             issue = faker.sentence()
         tenant = create_tenant()
-        ticket = TicketModel(
-            **ticket_attributes(issue, tenant, create_admin_user(), create_join_staff())
-        )
+        ticket = TicketModel(**ticket_attributes(issue, tenant, create_join_staff()))
```
NickSchimek: Since we're modifying this lets update it to use the create method instead of bypassing schema validations.

NickSchimek (changes requested):
This looks good. Just have a few small changes. Thanks!

Your going to need to rebase. Let me know if you need help with this.

schemas/ticket.py
aedwardg: I know that this is used for Marshmallow serialization.  I'm not 100 percent sure if it does anything else.  I know we've been doing this for the fields that have a db relationship with another field but I don't know that it's necessary outside of serialization.

aedwardg:
Made all the requested changes, but I did not remove the nested fields in the schema.

schemas/ticket.py
```diff
@@ -12,25 +12,17 @@ class Meta:
...
-        user = UserModel.query.get(value)
-        if not user:
-            raise ValidationError(f"{value} is not a valid user ID")
-        if user.role != RoleEnum.STAFF:
-            raise ValidationError(f"{value} is not JOIN staff")
-
-    @validates("senderID")
-    def validate_sender(self, value):
+    @validates("author_id")
+    def validate_creator(self, value):
```
NickSchimek: this probably should be validate_author

tests/factory_fixtures/ticket.py
```diff
@@ -1,16 +1,16 @@
 import pytest
 from models.tickets import TicketModel
 from models.tickets import TicketStatus
+from schemas.ticket import TicketSchema
 
 
 @pytest.fixture
 def ticket_attributes(faker):
-    def _ticket_attributes(issue, tenant, assignedUser, sender):
+    def _ticket_attributes(issue, tenant, creator):
```
NickSchimek: probably author here too

NickSchimek (approved):
Thanks!

PR #279, opened 2021-05-19
Issue619 - Use Schema's create function & fix tests

### LINKED ISSUE #619: Refactor the post method in the Tickets resource
Refactor the post method in the Tickets resource to use the new create method provided by the base resource class.

Chris-Boe:
## Description
Fixes some test data and uses the schema's create functionality in the ticket POST method

Closes codeforpdx/dwellingly-app/issues/619

### DIFF
--- a/resources/tickets.py
+++ b/resources/tickets.py
@@ -1,4 +1,4 @@
-from flask_restful import Resource, reqparse
+from flask_restful import Resource
 from models.tickets import TicketModel
 from models.tenant import TenantModel
 from utils.authorizations import pm_level_required
@@ -7,14 +7,6 @@ from schemas.ticket import TicketSchema
 
 
 class Ticket(Resource):
-    parser = reqparse.RequestParser()
-    parser.add_argument("senderID")
-    parser.add_argument("tenantID")
-    parser.add_argument("status")
-    parser.add_argument("urgency")
-    parser.add_argument("issue")
-    parser.add_argument("assignedUserID")
-
     @pm_level_required
     def get(self, id):
         return TicketModel.find(id).json()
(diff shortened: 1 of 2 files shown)


resources/tickets.py
```diff
@@ -70,11 +71,8 @@ def get(self):
 
     @pm_level_required
     def post(self):
-        data = Tickets.parser.parse_args()
-        ticket = TicketModel(**data)
-
-        ticket.save_to_db()
-        return ticket.json(), 201
+        TicketModel.create(schema=TicketSchema, payload=request.json)
+        return {"message": "Ticket successfully created"}, 201
```
Chris-Boe, line 39: We are not using the returned JSON of the created ticket at all in the frontend (PR: https://github.com/codeforpdx/dwellingly-app/pull/611) so I changed this to just return a success message

NickSchimek (changes requested):
Thanks for the PR. We have someone fixing the seed file in a different issue. Rather than modifying the seed file. I would prefer to disconnect the test from the seed file by using factories.

tests/integration/test_tickets.py
```diff
@@ -58,26 +58,13 @@ def test_tickets_POST(client, auth_headers):
...
-    assert response.json["assignedUserID"] == 4
-    assert response.json["sender"] == "user1 tester"
-    assert response.json["assigned"] == "Mr. Sir"
-    assert response.json["status"] == TicketStatus.New
-    assert response.json["urgency"] == "low"
-
-    # verify jwt only
-    response = client.post(endpoint, json=newTicket, headers=auth_headers["admin"])
-    assert is_valid(response, 201)
+    assert response.json == {"message": "Ticket successfully created"}
```
NickSchimek, line 67: There is still some work that needs to be done with this test, but I will go ahead and merge for now. This test isn't actually testing anything useful. It's just testing that we return json, but not actually testing that we called the create method. See the discussion here: https://github.com/codeforpdx/dwellinglybackend/pull/275#discussion_r637637833

NickSchimek (approved):
Thanks!

PR #281, opened 2021-05-20
628 refactor to use role level fixtures

### LINKED ISSUE #628: Use role-level fixtures in `test_email_authorization.py`
### Purpose
**Step 3 of 3** in moving away from using `auth_headers` in the integration tests (sub-issue of #547)

Authorization tests are for testing whether a user has the role/permissions to perform a certain action.  To do this, we don't need to hit the login endpoint every time (which is what `auth_headers` is doing).   For more information, see #545 

### Proposed Solution
Replace the `auth_headers` fixture with with the appropriate role-level fixture (`admin_header`, `staff_header`, `pm_header`) in **`test_email_authorization.py`** tests. This is located in a separate `authorizations` test directory.  

NOTE:
- This may require altering the tests to use `empty_test_db` as well if the test doesn't already.
- There is no header fixture for "pending" users, because these users have no role and will not be authenticated.  To test that a pending user cannot perform an action, write the API call with no header information.

### Acceptance Criteria
- [ ] All integration tests in `test_email_authorization.py` (i.e., checks if the user has permission to perform the action) should use the appropriate role-level fixture instead of `auth_headers`.

ovsyankimo:
## Description
Refactor test_email_authorization.py to use role-level fixtures, not `auth_headers` fixture.

Closes codeforpdx/dwellingly-app/issues/628

### DIFF
--- a/tests/authorizations/test_email_authorization.py
+++ b/tests/authorizations/test_email_authorization.py
@@ -13,9 +13,7 @@ class TestEmailAuthorizations:
         assert is_valid(response, 401)
         assert response.json == {"message": "Missing authorization header"}
 
-    def test_all_roles_except_admin_are_denied_access(self, auth_headers):
+    def test_all_roles_except_admin_are_denied_access(self, pm_header):
         payload = {"user_id": 1, "subject": "Some email subject", "body": "Some body"}
-        for role, token in auth_headers.items():
-            if role != "admin":
-                response = self.client.post(self.endpoint, json=payload, headers=token)
-                assert is_valid(response, 401)
+        response = self.client.post(self.endpoint, json=payload, headers=pm_header)
+        assert is_valid(response, 401)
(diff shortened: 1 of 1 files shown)


tests/authorizations/test_email_authorization.py
```diff
@@ -13,9 +13,7 @@ def test_auth_header_is_required(self):
         assert is_valid(response, 401)
         assert response.json == {"message": "Missing authorization header"}
 
-    def test_all_roles_except_admin_are_denied_access(self, auth_headers):
+    def test_all_roles_except_admin_are_denied_access(self, pm_header):
```
NickSchimek: Can we update the test name. It is saying test all roles, but we are now testing one role. Furthermore, with this new test I'm not sure if join_staff have access or not. With the previous test I knew only admins had access.

Preferably I would like to have 3 tests here. One for each role.

NickSchimek (changes requested):
See my comment below. Previously we were testing that join staff and pms did not have access, and we have now lost one of those tests. If we want to do this with one test then I would assert that join staff do not have access, because the way things are designed we can infer that if join staff do not have access then pms do not have access. However, it doesn't work the other way around.

NickSchimek (approved):
Looks good thanks!

PR #282, opened 2021-05-20
Refactor properties tests

### LINKED ISSUE #595: Use `valid_header` for Property integration tests
### Purpose

### Proposed Solution
Replace the `auth_headers` fixture with `valid_header` in all integration tests in **`tests/integration/test_properties.py`**.  

### Acceptance Criteria
- [x] All integration tests that test _logic_ (i.e., checks if the action functions as intended) should use the `valid_header` fixture instead of `auth_headers`.
- [x] All tests should still pass.

Chris-Boe:
## Description
Refactors the properties integration tests to be class based and use valid_headers

Closes codeforpdx/dwellingly-app/issues/595

### DIFF
--- a/tests/integration/test_properties.py
+++ b/tests/integration/test_properties.py
@@ -6,7 +6,7 @@ from models.property import PropertyModel
 
 
 @pytest.mark.usefixtures("client_class", "empty_test_db")
-class TestProperty:
+class TestPropertyGet:
     def setup(self):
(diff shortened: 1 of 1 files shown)


Chris-Boe:
@aedwardg I'm a little confused on how we're using the mock library, should I be updating the other tests to use that?

NickSchimek (approved):
Thanks

PR #283, opened 2021-05-26
create note fixture no longer needs parameters

### LINKED ISSUE #639: Update create notes fixture
Create note fixture currently requires two positional arguments: User, and ticket_id

The create note fixture should be able to create a note without any arguments. Update the factory so that these arguments are optional.

Protopaco:
## Description

Updated note factory fixture so that no attributes are required, though they still can be passed.

Closes codeforpdx/dwellingly-app/issues/639

### DIFF
--- a/tests/factory_fixtures/notes.py
+++ b/tests/factory_fixtures/notes.py
@@ -1,5 +1,6 @@
 import pytest
 from models.notes import NotesModel
+from schemas.notes import NotesSchema
 
 
 @pytest.fixture
@@ -17,8 +18,9 @@ def create_note(faker, note_attributes, create_admin_user, create_ticket):
         user = user or create_admin_user()
         ticket = ticket or create_ticket()
 
-        note = NotesModel(**note_attributes(text, user, ticket))
-        note.save_to_db()
-        return note
+        return NotesModel.create(
+            schema=NotesSchema,
+            payload={"text": text, "user_id": user.id, "ticket_id": ticket.id},
+        )
 
     yield _create_note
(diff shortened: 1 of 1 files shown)


tests/factory_fixtures/notes.py
```diff
@@ -11,11 +11,13 @@ def _note_attributes(text, user, ticket):
...
-            text = faker.paragraph()
-        note = NotesModel(**note_attributes(text, userid, ticketid))
+def create_note(faker, note_attributes, create_admin_user, create_ticket):
+    def _create_note(user=None, ticket=None, text=None):
+        text = text or faker.paragraph()
+        user = user or create_admin_user()
+        ticket = ticket or create_ticket()
+
+        note = NotesModel(**note_attributes(text, user, ticket))
         note.save_to_db()
```
aedwardg: Since we're updating this file, let's change this to use the NoteModel's `.create` method so we can validate it with the schema.  Check out the `create_lease` fixture for an example of how to do this.

aedwardg (changes requested):
Thanks, @Protopaco, this looks great. I just have one suggested change below to make this fixture a little better since we're updating it anyway.

tests/factory_fixtures/notes.py
Protopaco: Makes perfect sense. I made the change in a new commit.  We should probably hold off on merging it until #284 is pushed however.  The variable names "ticketid" and "userid" will need to be updated to bring it in line with the new terminology.  Thanks!

NickSchimek (approved):
Looks good! Thanks for reviewing @aedwardg

PR #285, opened 2021-05-27
refactors tenant tests

### LINKED ISSUE #593: Refactor tenant tests
**Purpose**
The unit/tenant test needs a method testing the JSON method. 
 - FYI: The `class TestBaseTenantModel` should only have the setup method in it. No other functions should be in that class.

Whoever picks this up. Look at other integration tests that have been refactored for an example of what to do here.

Protopaco:
## Description

Refactors the tenants tests using classes and only affirmative tests.  

Closes codeforpdx/dwellingly-app/issues/593

### DIFF
--- a/tests/integration/test_tenants.py
+++ b/tests/integration/test_tenants.py
@@ -1,89 +1,67 @@
 import pytest
-from conftest import is_valid
 from models.tenant import TenantModel
-from schemas.tenant import TenantSchema
-from utils.time import Time
-from unittest.mock import patch
-
-endpoint = "/api/tenants"
-
-
-def test_tenants_GET_all(client, test_database, auth_headers):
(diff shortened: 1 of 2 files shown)


NickSchimek (commented):
@Protopaco This PR looks like it is moving in the right direction, there are a few things I want to address.

I think my original intent of the issue was to only update the unit test file in `tests/unit/test_tenant.py`. I'm not sure what my intent of the FYI was about. 

I see an open ticket regarding the integration/ tenant test file so I went ahead and assigned that to you codeforpdx/dwellingly-app/issues/596. Can you separate this out into two PR's. One for the integration test, and one for the unit test. And then I can address each of them individually. Thanks!

Protopaco:
I ended up creating two new pull requests to avoid confusion when splitting this one up.

PR #286, opened 2021-05-28
Update CI badge in README.md

aedwardg:
## Description
Updates the README badge to use the GitHub Actions workflow instead of Travis (which hasn't been run in months).
Note that GitHub actions badges automatically use the name of the workflow, so if we don't like "CI Tests" we'll have to update that in the workflow file.


NickSchimek (approved):
That's good to note. I think CI Tests name is fine

PR #287, opened 2021-05-30
Implement collection delete method for NobiruList

NickSchimek:
Implements a delete method that can be called on a NobiruList
collection. The method takes an object instance, and returns None if
successful; otherwise it raises a 404 Not Found error.


Protopaco (approved):
Looks good!

aedwardg (approved):
Saucy! I like it 👍

PR #288, opened 2021-05-31
refactors tenant integration tests

### LINKED ISSUE #596: Use `valid_header` for Tenant integration tests
### Purpose

### Proposed Solution
Replace the `auth_headers` fixture with `valid_header` in all integration tests in **`tests/integration/test_tenants.py`**.  

### Acceptance Criteria

Protopaco:
## Description

Refactors tenant integration tests 

Closes codeforpdx/dwellingly-app/issues/596


tests/integration/test_tenants.py
```diff
@@ -1,89 +1,67 @@
...
 
-        with patch.object(TenantModel, "update", return_value=tenant) as mock_update:
-            response = self.client.put(
-                f"{self.endpoint}/{tenant.id}",
-                json=updated_fields,
-                headers=valid_header,
-            )
+        response = self.client.get(f"{self.endpoint}/{tenant.id}", headers=valid_header)
+
+        assert response.json == tenant.json()
```
aedwardg, line 17: This `test_get_tenant` test looks good 👍

tests/integration/test_tenants.py
```diff
@@ -1,89 +1,67 @@
...
+    def test_get_tenant(self, valid_header, create_tenant):
         tenant = create_tenant()
-        updated_fields = {"archived": True}
 
-        with patch.object(TenantModel, "update", return_value=tenant) as mock_update:
-            response = self.client.put(
-                f"{self.endpoint}/{tenant.id}",
-                json=updated_fields,
-                headers=valid_header,
-            )
```
aedwardg: This update tenant test was using our newer testing strategy, monkey-patching the `update` method and asserting that it had been called, and what it had been called with.  This will need to go back to the way it was.  I like that it's now in the `TestPut` class and uses the decorator for the `empty_test_db` fixture, so those changes can stay, but the rest should be reverted to how it was before.

tests/integration/test_tenants.py
```diff
@@ -1,89 +1,67 @@
...
+        self.endpoint = "/api/tenants"
+
+    def test_get_all_tenants(self, valid_header, create_tenant):
+        create_tenant()
+        create_tenant()
+        all_tenants = TenantModel.query.json()
+
+        response = self.client.get(self.endpoint, headers=valid_header)
+
+        assert response.json["tenants"] == all_tenants
```
aedwardg: So right here, you're not really testing anything. You've written the same exact logic that is in the /tenants/ GET endpoint that we're trying to test:

```python
TenantModel.query.json()
```

Because you've duplicated the logic from the resource in this test, the only thing this test is doing is essentially making sure that the Flask-Restful Resource works.

Can you think of a different way to check that the proper data is returned in the response, regardless of the implementation?

tests/integration/test_tenants.py
```diff
@@ -1,89 +1,67 @@
...
+            "phone": "1234567890",
+            "archived": False,
+        }
+
+    def test_create_tenant(self, valid_header):
+        response = self.client.post(
+            self.endpoint, json=self.new_tenant, headers=valid_header
+        )
+
+        assert response.json == TenantModel.find(response.json["id"]).json()
```
aedwardg: Just like with the `get_all_tenants` test, the logic here is identical to the logic in the endpoint.

For this one, we should probably be patching the `create` method of the `TenantModel` and asserting that it was called once with the correct arguments, like the `test_update_tenant` test was doing.

If you need an example, you can take a look at the `test_create_lease` test:
https://github.com/codeforpdx/dwellinglybackend/blob/6028ed50aa66bf038b24fafcaf46d6dfbbef1072/tests/integration/test_leases.py#L42-L50

Let me or @NickSchimek know if you have any questions about how or why to do this!

aedwardg (changes requested):
@Protopaco this is a good start. I love seeing it all changed over to `valid_header` finally!
It looks like you also did some refactoring here (I think based on a different issue originally?) and since you've already got that half done we may as well finish up the refactoring in this PR as well.

I've left some comments below on changes I'd like to see here to make sure everything is up to our newer standards. Thanks!

tests/integration/test_tenants.py
NickSchimek: Interesting, I would've never referred to this as monkey patching, but it is monkey patching. 👍

NickSchimek:
@Protopaco I think @aedwardg has some good points here. The tests are good, however they can be improved a bit. Please let us know if you want to pair or if you have any questions on how to improve these. We're basically trying to refactor and improve the code base so that we can propagate good techniques going forward.

Protopaco:
This is next on my list!  Thank you!

tests/integration/test_tenants.py
Protopaco: Is this more of what you're looking for?

```python
    def test_get_all_tenants(self, valid_header, create_tenant):
        tenants = [create_tenant().json(), create_tenant().json()]

        response = self.client.get(self.endpoint, headers=valid_header)

        assert response.json["tenants"] == tenants
```

Protopaco: I tried using the patch.object framework,

```python
        with patch.object(TenantModel, "query") as mock_query:
            response = self.client.get(self.endpoint, headers=valid_header)
```

but I kept getting errors:
```
TypeError: Object of type MagicMock is not JSON serializable
```

tests/integration/test_tenants.py
Protopaco: I'm getting the same 

```
TypeError: Object of type MagicMock is not JSON serializable
``` 
for this one, and it is modeled on the test_create_lease 

```python
        with patch.object(TenantModel, "create") as mock_create:
            response = self.client.post(
                self.endpoint, json=self.new_tenant, headers=valid_header
            )
```
I'm completely lost on this, but  I've been swamped with interviews/other projects the last couple of weeks, so my brain is basically tapioca at this point.

tests/integration/test_tenants.py
aedwardg: > 
> 
> Is this more of what you're looking for?
> 
> ```python
>     def test_get_all_tenants(self, valid_header, create_tenant):
>         tenants = [create_tenant().json(), create_tenant().json()]
> 
>         response = self.client.get(self.endpoint, headers=valid_header)
> 
>         assert response.json["tenants"] == tenants
> ```

Yes, this is a much better test since it checks that the endpoint logic (`TenantModel.query.json()`) is behaving the way we expect it to.

In regards to the patching, I'm not sure that we should do so here.  
If we wanted to, we could mock the `NobiruQuery.json()` method here and assert that it was called once, or alternatively, we could mock the `TenantModel.json()` method and assert that the mock's `call_count` is 2, but I think both of those are more confusing than this test.

tests/integration/test_tenants.py
aedwardg: Hm...now that I've taken a closer look at the Tenants endpoint, I think this create action is a lot harder to mock than I originally thought. Sorry for making such a difficult suggestion! 😅 

You can get around the error you are receiving by providing something that _is_ json serializable as the `return_value` of the mock, but the lease creation logic that's also in this `post` method messes it all up even when you do that.  Perhaps we can figure out a better way to test this on Wednesday or at a later date.  

Either way, however, so long as that lease creation logic is in there, we should also have a test checking that a lease gets created and associated with the tenant when `dateTimeEnd`, `dateTimeStart` and `propertyID` are present in the payload.  I believe the original `testTenantsPOST` test did this to some extent.

Protopaco: No problem.  I played with it for a while, but I don't feel like I have solid grasp on how patch works with beyond the straight forward examples we've used.  I will try and do a bit more research before Wednesday's meeting in case we want to take these to the next level.

How do we feel about these tests?:

```python

class TestCreate:
    def setup(self):
        self.endpoint = "/api/tenants"
        self.new_tenant = {
            "firstName": "Testy",
            "lastName": "McTesterson",
            "phone": "1234567890",
            "archived": False,
        }

    def test_create_tenant(self, valid_header):
        response = self.client.post(
            self.endpoint, json=self.new_tenant, headers=valid_header
        )

        assert response.status_code == 201
        assert response.json["firstName"] == self.new_tenant["firstName"]
        assert response.json["lastName"] == self.new_tenant["lastName"]
        assert response.json["phone"] == self.new_tenant["phone"]

    def test_create_tenant_with_lease(self, valid_header, create_property, create_lease):
        lease = {
            'occupants': 47,
            'dateTimeStart':"2021-06-19T07:00:00.000Z",
            'dateTimeEnd':"2022-07-16T07:00:00.000Z",
            'unitNum': '5351',
            "propertyID": create_property().id
        }

        response = self.client.post(
            self.endpoint, json={**self.new_tenant, **lease}, headers=valid_header
        )

        assert response.status_code == 201
        assert response.json["firstName"] == self.new_tenant["firstName"]
        assert response.json["lastName"] == self.new_tenant["lastName"]
        assert response.json["phone"] == self.new_tenant["phone"]
        assert response.json["unitNum"] == lease["unitNum"]
        assert response.json["occupants"] == lease["occupants"]
```

Protopaco: Also, I need to dig into it some more, but it appears to be a problem with the returned data from this endpoint.  The lease information is part of the main object, not a part of the "lease" object:

```python
{'id': 1, 
'firstName': 'Testy', 
'lastName': 'McTesterson', 
'fullName': 
'Testy McTesterson', 
'phone': '1234567890', 
'lease': '', 
'staff': [], 
'created_at': '06/15/2021 18:02:28', '
updated_at': '06/15/2021 18:02:28', 
'archived': False, 
'occupants': 47, 
'propertyID': 1, 
'unitNum': '5351'}
```

I believe it should look like: 
```python
{'id': 1, 
'firstName': 'Testy', 
'lastName': 'McTesterson', 
'fullName': 'Testy McTesterson', 
'phone': '1234567890', 
'staff': [], 
'created_at': '06/15/2021 18:02:28', 
'updated_at': '06/15/2021 18:02:28', 
'archived': False,
'lease': {
     'occupants': 47, 
     'propertyID': 1, 
     'unitNum': '5351'
     }
}

I will look more into it, but is this a known problem, or a problem at all?

**Comment by Protopaco** [comment]
@aedwardg @NickSchimek 
Before I get back into refactoring the tests, we wanted to refactor the create tenants endpoint, specifically the portion involved with creating a lease.  I believe this creates the same behavior without manipulating the request.json.   It still feels a little janky to me, but cleaner than what was in there before.

```python
class Tenants(Resource):
    @admin_required
    def get(self):
        return {"tenants": TenantModel.query.json()}

    @admin_required
    def post(self):
        tenantEntry = TenantModel.create(schema=TenantSchema, payload=request.json).json()

        if request.json.keys() >= {"dateTimeEnd", "dateTimeStart", "propertyID"}:
            leaseEntry = LeaseModel.create(
                schema=LeaseSchema,
                payload={**request.json, "tenantID": tenantEntry["id"]}
            ).json()
            tenantEntry = {**tenantEntry, **leaseEntry}

        return tenantEntry, 201
```

NickSchimek:
It looks like a big improvement 👍. It looks like the request has two different objects within the same json object. I'd like to improve this more, but we'd have to do it in a different ticket as it willl be breaking changes, because we should change the json request to be of a more correct structure.

Protopaco:
> I'd like to improve this more, but we'd have to do it in a different ticket as it will be breaking changes, because we should change the json request to be of a more correct structure.

I agree, this is the only instance where the tenant and lease data are handled this way.  I think it makes sense to pass the lease as a sub-object in the tenant object.  If we were to create a new ticket to handle this, I can also do the fix on the front end.  

Would these changes need to wait, until that work was finished, so we can refactor the tests for the final version?

aedwardg:
@Protopaco I agree, that's definitely a big improvement. Interesting use of set comparison logic too!  
I think technically, we'd want to check that `request.json.keys()` is a _proper_ superset though, since the request.json will never only have those three keys in them when creating a tenant.  To change that, you'd just use `>` instead of `>=`.

I agree with Nick that in the future we should change the structure of the json request, but for now I think your change here would be fine for this PR.  It would allow you to mock out the `create` method of `TenantModel` for the test, just make sure to pass in a `return_value` to `patch.object` that is JSON serializable.

NickSchimek:
> Would these changes need to wait, until that work was finished, so we can refactor the tests for the final version?

Yes, Let's wait on doing breaking changes. I will create a separate ticket for that once this PR is ready to merge.

I don't see any new changes/commits here. Once your ready for me to re-review this then please re-request review and I will get a notification that this is ready for me to re-review. thanks!

Screen Shot 2021-06-27 at 1 22 47 PM

Protopaco:
I pushed up those changes, and refactored the test_create_tenant test.  Everything seems to be working as desired.

I attempted to refactor the test_create_tenant_with_lease test using patch and the LeaseModel.  I couldn't get the tenantID to match up however.   I feel like it was a weird interaction between patch and how the resource is currently functioning.  When we do the next PR, and refactor the resource again, I will try and update that test as well.

NickSchimek:
@Protopaco This looked pretty good, however the tests worried me, and I couldn't tell from the tests what the return value was. Furthermore, I couldn't tell what the return value was for the resource either because the json method was not being used.

I went in there and changed the test to do a `assert request.json == {}` so that I could see the output. And this is what it was. (There is no lease data being serialized where it should be)

Screen Shot 2021-06-29 at 12 31 26 AM


I went ahead and thought I'd make some small changes, but then I started making more changes. Since the data from the frontend doesn't come in the correct format for us to do our validations. I decided to modify the format to the format we need, and used the create method to load and create all objects.

Here is what the response looks like now.

Screen Shot 2021-06-29 at 12 59 41 AM


Please review the changes and let me know if you have any questions. Thanks

Protopaco:
I think this makes perfect sense.  Better to just fix the problem then kick the can down the road.  We'll need to create a new ticket to address the changes on the frontend, but it should be an easy fix.

Looking at the tenants resource and _build_payload, are you constructing the data object to pass to TenantModel.create using the current data coming from the frontend?  Would it make more sense to skip this step and have the front send the correct data?  I don't think it would take more than a couple of lines to do.

NickSchimek:
I don’t think this is a breaking change for the frontend. Can you verify.

And yes, we can remove build payload if the frontend sends the correct json
structure.

NickSchimek:
I don’t like having multiples schemas just for the post load. I think I
know how to fix that. I think the include foreign keys is what caused
issues. And if that’s removed then we can do it all in one schema.

The reason I think include fk makes all fk required. But tenant fk is not
needed cause it will be added automatically.

Do u want to try cleaning up the lease schema file?

aedwardg (approved):
I think these new changes to the tests look good.  The refactored resource also looks good -- considering how the front end is sending the data.

I agree that it would be a lot nicer to consolidate the schemas into one and remove the `_build_payload` parsing from the controller by having the front end just send the lease info in an object rather than as separate fields.

Personally, I'd rather see those changes in one or two separate PRs since this one was only supposed to change `auth_headers` to `valid_header`, but I'll defer to @NickSchimek on whether you think those changes should be included in this PR or should be done separately.

NickSchimek:
Since the extra schemas were added in this PR then removing and fixing should also be in this PR before a merge.

Updating the front end and removing the build_payload can be either A. in this PR if the frontend is updated this week. B. separately if its not.

>  this one was only supposed to change auth_headers to valid_header

Someone did extra credit

PR #289, opened 2021-05-31
refactors tenant unit tests

### LINKED ISSUE #593: Refactor tenant tests

Protopaco:

Closes codeforpdx/dwellingly-app/issues/593


tests/unit/test_tenant.py
```diff
@@ -12,30 +11,22 @@ def setup(self):
...
-        )
-
-        # Create an active lease for the tenant
-        lease_active = create_lease(
-            tenant=tenant, property=PropertyModel.find(lease_expired.propertyID)
-        )
-
-        # Active lease should be the only one that shows up
-        assert tenant.json()["lease"] == lease_active.json()
-
```
aedwardg: This is testing that only the active lease is serialized in the `.json` method. This should not be removed.

tests/unit/test_tenant.py
```diff
@@ -12,30 +11,22 @@ def setup(self):
...
 
 @pytest.mark.usefixtures("empty_test_db")
-class TestTenantFactory:
-    def test_create_tenant(self, create_tenant):
-        assert create_tenant()
```
aedwardg: This is testing the `create_tenant` factory and should not be removed.

tests/unit/test_tenant.py
```diff
@@ -12,30 +11,22 @@ def setup(self):
...
+class TestTenantModel:
+    def test_json(self, create_tenant, create_lease):
+        tenant = create_tenant()
+        lease = create_lease(tenant=tenant)
+
+        assert tenant.json() == {
+            "id": tenant.id,
+            "firstName": tenant.firstName,
+            "lastName": tenant.lastName,
+            "fullName": "{} {}".format(tenant.firstName, tenant.lastName),
```
aedwardg, line 49: Not really related to this PR, but @NickSchimek when can we get rid of this ridiculous `fullName` field in the json? 😂

aedwardg (changes requested):
@Protopaco thanks for taking this on.  The unit test for the json method looks good 👍 we just need you to add back in a couple of the tests you deleted and this should be good to go.

Protopaco:
Whoops!  Sorry about that, seems I got a little delete happy.  Should we add these same factory tests to the notes unit tests?

aedwardg:
> Should we add these same factory tests to the notes unit tests?

Yes, each factory should have a test like that one.  I think most of them do, but there might be a couple of newer ones that are missing it.  But adding those tests should be saved for a separate issue/PR, not done in this one.

aedwardg:
Thanks for adding those two tests back in!

If you could just move `TestTenantFactory` to the bottom of the file and out of the `TestTenantModel` class, I think this will be good to go 🙂

Protopaco:
Yeah, I won't add those tests in this PR, but I will keep it in mind next time I'm bored.  ;)

tests/unit/test_tenant.py
NickSchimek: This test doesn't belong here. I think it does need to be removed, but it needs to be replaced. This is testing the `active` method in `models/lease.py`. It's testing this through the json method. I wouldn't mind if this is done in this PR or we make another one for it. maybe the latter. But the `unit/test_lease` file should have a test confirming that its active method works appropriately. And then this can be removed.

To test that the `tenant.json` method returns the active lease then we would need to stub this method and test that it is called, as the json method shouldn't retest the functionality of the active method here.

tests/unit/test_tenant.py
NickSchimek: This test is kind of crappy. It doesn't appear to be doing much. But I kinda like having factory tests. If something changes that causes the factory to fail and stop it from performing its job then we would get a lot of red tests, but seeing this test fail would point the developer to the root cause of the issue. While not entirely needed, it can be useful and is cheap to have. The user factory tests are a bit more involved as they test that the factory creates the right role too.

tests/unit/test_tenant.py
NickSchimek: I had no plans to get rid of it, but we certainly can and should. Someone needs to make an issue on the frontend to stop depending on it, and once that's completed then we can remove it.

NickSchimek (approved):
Thanks @Protopaco 💫 This looks good. See my replies below

PR #305, opened 2021-06-02
Use valid_header for Users integration tests

### LINKED ISSUE #598: Use `valid_header` for Users integration tests
### Purpose

### Proposed Solution
Replace the `auth_headers` fixture with `valid_header` in all integration tests in **`tests/integration/test_users.py`**.  

### Acceptance Criteria

qq1u:

Closes codeforpdx/dwellingly-app/issues/598


tests/integration/test_users.py
```diff
@@ -84,21 +84,21 @@ def test_get_pm_by_id(
     assert tenants_list == [lease_1.tenant.json(), lease_2.tenant.json()]
 
 
-def test_user_roles(client, auth_headers):
+def test_user_roles(client, test_database, valid_header):
```
NickSchimek, line 87: Declaring the `test_database` here isn't do anything anything for us. The functional based tests already have access to the `test_database`. FYI: We're moving away from functional based tests to class based tests which will utilize the `empty_test_db`, because the `test_database` takes `seedData(n)` time to spin up, and we plan to add a lot more seed data in the future, and when we do that our tests will become a lot slower. Where as the empty_test_db spins up very fast because there is no data, and it uses pytest fixtures which are supposedly cached.

Other than that this looks great! Thank you.

NickSchimek (approved):
Thanks @13528080556 💫 This looks great! I have a note below that explains the direction we are heading in regarding tests. Thanks for the contribution.

PR #306, opened 2021-06-06
Move staff_tenant import to tenant model

NickSchimek:
## Description

This is able to be done after having fixed the relationships for the
staff_tenant table in PR #231

Moved the staff_tenant import out of app.py and into the tenant model. This was previously unable to be done because of how the relationships were being defined.


aedwardg (approved):
Awesome! LGTM 🚀

PR #307, opened 2021-06-06
More refactoring - Moves routes into separate file

NickSchimek:
## Description

Created a config directory and moved the environment configs there.
Moved the route creations into config directory


aedwardg (approved):
This looks good.  I'm always a fan of cleaner app entry points and `app.py` looks a lot nicer now.

PR #309, opened 2021-06-06
Update installation instruction to include pyenv

NickSchimek:
## Description

Add pyenv install step. This will allow pipenv to automatically install
the correct python version. Hopefully reducing install issues.


qq1u (approved):
That is great!

PR #310, opened 2021-06-06
Refactor tickets integration tests

### LINKED ISSUE #597: Use `valid_header` for Tickets integration tests
### Purpose

### Proposed Solution
Replace the `auth_headers` fixture with `valid_header` in all integration tests in **`tests/integration/test_tickets.py`**.  

### Acceptance Criteria

aedwardg:
## Description

Replaces `auth_headers` with `valid_header` and refactors all tests in the file.

All tests are now separated into appropriate classes, which inherit fixtures and endpoint from a base configuration class.
Methods are stubbed out where appropriate, but I couldn't figure out a nice way to test the delete action here using the patch, so those tests still hit the endpoint.

Closes codeforpdx/dwellingly-app/issues/597

### DIFF
--- a/resources/tickets.py
+++ b/resources/tickets.py
@@ -1,7 +1,6 @@
 from db import db
 from flask_restful import Resource
 from models.tickets import TicketModel
-from models.tenant import TenantModel
 from utils.authorizations import pm_level_required
 from flask import request
 from schemas.ticket import TicketSchema
@@ -27,12 +26,10 @@ class Ticket(Resource):
 class Tickets(Resource):
     @pm_level_required
     def get(self):
-        if request.args and request.args["tenant_id"]:
-            return {
-                "tickets": TenantModel.find(request.args["tenant_id"]).tickets.json()
-            }
-        else:
-            return {"tickets": TicketModel.query.json()}
+        tickets = TicketModel.query
+        if "tenant_id" in request.args:
+            tickets = tickets.where(TicketModel.tenant_id == request.args["tenant_id"])
+        return {"tickets": tickets.json()}
 
     @pm_level_required
     def post(self):
--- a/tests/factory_fixtures/ticket.py
+++ b/tests/factory_fixtures/ticket.py
@@ -5,12 +5,12 @@ from schemas.ticket import TicketSchema
 
 
 @pytest.fixture
-def ticket_attributes(faker):
-    def _ticket_attributes(issue, tenant, author):
+def ticket_attributes(faker, create_tenant, create_join_staff):
+    def _ticket_attributes(issue=None, tenant=None, author=None):
         return {
-            "issue": issue,
-            "tenant_id": tenant.id,
-            "author_id": author.id,
+            "issue": issue or faker.sentence(),
+            "tenant_id": tenant.id if tenant else create_tenant().id,
+            "author_id": author.id if author else create_join_staff().id,
             "status": TicketStatus.New,
             "urgency": faker.random_element(("Low", "Medium", "High")),
         }
--- a/tests/integration/test_tickets.py
+++ b/tests/integration/test_tickets.py
@@ -10,6 +10,43 @@ class BaseConfig:
     endpoint = "/api/tickets"
 
 
+class TestTicketGET(BaseConfig):
+    def test_get(self, valid_header, create_ticket):
+        ticket = create_ticket()
+        response = self.client.get(f"{self.endpoint}/{ticket.id}", headers=valid_header)
+
+        assert response.status_code == 200
+        assert response.json == ticket.json()
+
+
+class TestTicketPUT(BaseConfig):
+    def test_update(self, valid_header, create_ticket):
+        ticket = create_ticket()
+        with patch.object(TicketModel, "update", return_value=ticket) as mock_update:
+            response = self.client.put(
+                f"{self.endpoint}/{ticket.id}",
+                json={"hello": "world"},
+                headers=valid_header,
+            )
+
+        mock_update.assert_called_once_with(
+            schema=TicketSchema, id=ticket.id, payload={"hello": "world"}
+        )
+
+        assert response.status_code == 200
+        assert response.json == ticket.json()
+
+
+class TestTicketDELETE(BaseConfig):
+    def test_delete(self, valid_header):
+        with patch.object(TicketModel, "delete") as mock_delete:
+            response = self.client.delete(f"{self.endpoint}/1", headers=valid_header)
+
+        mock_delete.assert_called_once_with(1)
+        assert response.status_code == 200
+        assert response.json == {"message": "Ticket removed from database"}
+
+
 class TestTicketsGET(BaseConfig):
     def test_tickets_get_all(self, valid_header, create_ticket):
         ticket_1 = create_ticket()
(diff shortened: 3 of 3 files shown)


tests/integration/test_tickets.py
```diff
@@ -1,134 +1,122 @@
...
-    assert response.json == {"message": "Ticket IDs missing in request"}
+from models.tickets import TicketModel
+from schemas.ticket import TicketSchema
+from unittest.mock import patch
+
+
+@pytest.mark.usefixtures("client_class", "empty_test_db")
+class BaseConfig:
+    endpoint = "/api/tickets"
+
```
NickSchimek, line 11: This is great! I like this design pattern! Good work!

tests/integration/test_tickets.py
```diff
@@ -1,134 +1,122 @@
...
+        create_ticket()
+
+        response = self.client.get(
+            f"{self.endpoint}?tenant_id={ticket_1.tenant.id}", headers=valid_header
+        )
+
+        assert response.status_code == 200
+        assert response.json == {"tickets": [ticket_1.json()]}
+
+    def test_tickets_get_one(self, valid_header, create_ticket):
```
NickSchimek: This method is testing the endpoint `tickets/:id` so I think for readability lets put it in a separate class than the methods testing the endpoint `tickets`. <- Since both endpoints hit a completely different class and method then they should be in a separate class in the test. So I think the class would be named using the singular version `TestTicketGET` and the method can simply be `test_get`.

tests/integration/test_tickets.py
```diff
@@ -1,134 +1,122 @@
...
+        ticket_1 = create_ticket()
+        create_ticket()
+        create_ticket()
+
+        response = self.client.get(
+            f"{self.endpoint}?tenant_id={ticket_1.tenant.id}", headers=valid_header
+        )
+
+        assert response.status_code == 200
+        assert response.json == {"tickets": [ticket_1.json()]}
```
NickSchimek, line 69: As a side note: I always find this a bit hard to parse. 
```python
    def get(self):
        if request.args and request.args["tenant_id"]:
            return {
                "tickets": TenantModel.find(request.args["tenant_id"]).tickets.json()
            }
        else:
            return {"tickets": TicketModel.query.json()}
```

If you want to see if this can be refactored to be simpler than I would accept that in this PR too.
What I would like to see is something like this.
```python
tickets = TicketModel.query
if tenant_id:
  tickets = tickets.where(tenant_id: tenant_id)
return tickets.json()
```

What I like about this is that the context always focuses on the return value (tickets). Where as the other one Focuses on Tenants first, and I find myself checking to see which file I'm reading, and asking why is it trying to find a Tenant when its serializing tickets. Obviously I need to read the next few commands and figure it out, but for someone who doesn't see this code as often as I do would likely get confused too the first time they come about it. So if possible I would like to make it simpler to read at some point in the future.

I think sqlalchemy has methods to do that. Or we can make our own wrappers to whatever they provide.

tests/integration/test_tickets.py
```diff
@@ -1,134 +1,122 @@
...
+        ticket = create_ticket()
+        response = self.client.get(f"{self.endpoint}/{ticket.id}", headers=valid_header)
+
+        assert response.status_code == 200
+        assert response.json == ticket.json()
+
+
+class TestTicketsPOST(BaseConfig):
+    def test_create_ticket(self, valid_header):
+        new_ticket = {
```
NickSchimek: We should create a factory for this if we don't already have on that returns valid ticket attributes.

tests/integration/test_tickets.py
```diff
@@ -1,134 +1,122 @@
...
+                self.endpoint, json=new_ticket, headers=valid_header
+            )
+
+        mock_create.assert_called_once_with(schema=TicketSchema, payload=new_ticket)
+
+        assert response.status_code == 201
+        assert response.json == {"message": "Ticket successfully created"}
+
+
+class TestTicketsPUT(BaseConfig):
```
NickSchimek: Let's name this TestTicketPUT. Since the class we're testing is the `Ticket` class.

tests/integration/test_tickets.py
```diff
@@ -1,134 +1,122 @@
...
+
+        mock_update.assert_called_once_with(
+            schema=TicketSchema, id=ticket.id, payload={"hello": "world"}
+        )
+
+        assert response.status_code == 200
+        assert response.json == ticket.json()
+
+
+class TestTicketsDELETE(BaseConfig):
```
NickSchimek, line 87: Same as what I said above for the get endpoint. Let's split these tests to separate the different methods we're testing. One class for each method.

tests/integration/test_tickets.py
```diff
@@ -1,134 +1,122 @@
...
+        ticket_2 = create_ticket()
+        delete_ids = {"ids": [ticket_1.id, ticket_2.id]}
+
+        response = self.client.delete(
+            self.endpoint, json=delete_ids, headers=valid_header
+        )
+
+        assert response.status_code == 200
+        assert response.json == {"message": "Tickets successfully deleted"}
+        assert TicketModel.query.all() == []
```
NickSchimek, line 100: Whenever we call the database. We need to refresh the session 😉 ha I know you know this already. :-)

NickSchimek (changes requested):
Thanks @aedwardg  - These look great. I just have a few organizational comments. 1 sassy comment, and a 1 off the rails comments.

tests/integration/test_tickets.py
aedwardg: Updated this, take a look and let me know what you think

tests/integration/test_tickets.py
aedwardg: I went ahead and updated the `ticket_attributes` factory in this PR so I could get rid of that manual one.  Let me know if that works for you

resources/tickets.py
```diff
@@ -27,12 +26,10 @@ def put(self, id):
...
-        if request.args and request.args["tenant_id"]:
-            return {
-                "tickets": TenantModel.find(request.args["tenant_id"]).tickets.json()
-            }
-        else:
-            return {"tickets": TicketModel.query.json()}
+        tickets = TicketModel.query
+        if "tenant_id" in request.args:
+            tickets = tickets.where(TicketModel.tenant_id == request.args["tenant_id"])
+        return {"tickets": tickets.json()}
```
NickSchimek, line 32: 😄

tests/factory_fixtures/ticket.py
```diff
@@ -5,12 +5,12 @@
...
-    def _ticket_attributes(issue, tenant, author):
+def ticket_attributes(faker, create_tenant, create_join_staff):
+    def _ticket_attributes(issue=None, tenant=None, author=None):
         return {
-            "issue": issue,
-            "tenant_id": tenant.id,
-            "author_id": author.id,
+            "issue": issue or faker.sentence(),
+            "tenant_id": tenant.id if tenant else create_tenant().id,
+            "author_id": author.id if author else create_join_staff().id,
```
NickSchimek, line 13: why aren't you display???

NickSchimek (approved):
Looks great! I really like that `where`. Makes it a lot simpler to understand what's going on.

Thanks for making those additional updates!

PR #311, opened 2021-06-07
Create user subclasses

NickSchimek:
## Description

We have 3 different types of users and depending on the type we need to
serialize differently, and assign relationships differently. This
construct will allow all our users to still behave like a user, but
we will be able to define different methods, relationships and serializations for
each type of user.

We'll end up removing the roles altogether as the current types can replace them. We may need roles in the future, but until then we might as well remove code that is no longer needed. We'll accomplish this in future PR's once the the frontend has made necessary changes. For example: We have multiple endpoints that handle roles (not needed) that the frontend depends (also not needed) these are not needed regardless if we keep roles or not.

So if your in the shell
```python
>>> admin = Admin.query.all()[0]
>>> admin.id
1
>>> user = UserModel.find(1)
>>> user.id
1

>>> admin
<'Admin' {'_sa_instance_state': <sqlalchemy.orm.state.InstanceState object at 0x10b357160>, 'created_at': datetime.datetime(2021, 6, 7, 17, 16, 34, 355652), 'updated_at': datetime.datetime(2021, 6, 7, 17, 16, 34, 355658), 'type': 'admin', 'role': <RoleEnum.ADMIN: 4>, 'lastName': 'tester', 'hash_digest': b'$2b$12$2Jv4P0B1.LlWp7.crOgI/OaTU8dZRr7942eVjxGW77Uy3vWNCM9uq', 'archived': False, 'id': 1, 'email': 'user1@dwellingly.org', 'firstName': 'user1', 'phone': '555-555-5555', 'password': None, 'lastActive': datetime.datetime(2021, 6, 7, 17, 16, 34, 355660)}>
>>> user
<'Admin' {'_sa_instance_state': <sqlalchemy.orm.state.InstanceState object at 0x10b357160>, 'created_at': datetime.datetime(2021, 6, 7, 17, 16, 34, 355652), 'updated_at': datetime.datetime(2021, 6, 7, 17, 16, 34, 355658), 'type': 'admin', 'role': <RoleEnum.ADMIN: 4>, 'lastName': 'tester', 'hash_digest': b'$2b$12$2Jv4P0B1.LlWp7.crOgI/OaTU8dZRr7942eVjxGW77Uy3vWNCM9uq', 'archived': False, 'id': 1, 'email': 'user1@dwellingly.org', 'firstName': 'user1', 'phone': '555-555-5555', 'password': None, 'lastActive': datetime.datetime(2021, 6, 7, 17, 16, 34, 355660)}>

>>> staff = Staff.query.all()[0]
>>> staff
<'Staff' {'_sa_instance_state': <sqlalchemy.orm.state.InstanceState object at 0x10b36a3a0>, 'created_at': datetime.datetime(2021, 6, 7, 17, 16, 38, 879350), 'updated_at': datetime.datetime(2021, 6, 7, 17, 16, 38, 879355), 'type': 'staff', 'role': <RoleEnum.STAFF: 3>, 'lastName': 'Joinstaff', 'hash_digest': b'$2b$12$FRLq8JULJ4CuT36wIwUoQ.4IULYcQqLXPvVEmzqOfZ8zEoFCKNHT6', 'archived': False, 'id': 10, 'email': 'janice@joinpdx.org', 'firstName': 'Janice', 'phone': '555-555-5555', 'password': None, 'lastActive': datetime.datetime(2021, 6, 7, 17, 16, 38, 879357)}>

>>> staff.tenants
[<'TenantModel' {'_sa_instance_state': <sqlalchemy.orm.state.InstanceState object at 0x10b3952e0>, 'updated_at': datetime.datetime(2021, 6, 7, 17, 16, 39, 910715), 'firstName': 'Renty', 'phone': '800-RENT-ALOT', 'created_at': datetime.datetime(2021, 6, 7, 17, 16, 39, 910711), 'id': 1, 'lastName': 'McRenter', 'archived': False}>]
>>> admin.tenants
Traceback (most recent call last):
  File "<console>", line 1, in <module>
AttributeError: 'Admin' object has no attribute 'tenants'
```

The user and admin variable contain the same objects.
the admin does not have tenants
the staff does have tenants

PropertyManager has properties. Admin and Staff do not

### DIFF
--- a/models/user.py
+++ b/models/user.py
@@ -32,7 +32,10 @@ class RoleEnum(Enum):
 class UserModel(BaseModel):
     __tablename__ = "users"
 
+    __mapper_args__ = {"polymorphic_identity": "user", "polymorphic_on": "type"}
+
     id = db.Column(db.Integer, primary_key=True)
+    type = db.Column(db.String())
     email = db.Column(db.String(100), unique=True, nullable=False)
     role = db.Column(db.Enum(RoleEnum), default=None)
     firstName = db.Column(db.String(100), nullable=False)
--- a/models/user.py
+++ b/models/user.py
@@ -95,7 +95,7 @@ class UserModel(BaseModel):
         return bcrypt.checkpw(bytes(plaintext_password, "utf-8"), self.hash_digest)
 
     def json(self):
-        return {
+        base_json = {
             "id": self.id,
             "firstName": self.firstName,
             "lastName": self.lastName,
@@ -107,6 +107,10 @@ class UserModel(BaseModel):
             "created_at": Time.format_date(self.created_at),
             "updated_at": Time.format_date(self.updated_at),
         }
+        return {**base_json, **self.serialize()}
+
+    def serialize(self):
+        return {}
 
     def widgetJson(self, propertyName, date):
         return {
--- a/data/seedData.py
+++ b/data/seedData.py
@@ -1,6 +1,9 @@
 from datetime import datetime, timedelta
 
 from models.user import UserModel, RoleEnum
+from models.users.admin import Admin
+from models.users.staff import Staff
+from models.users.property_manager import PropertyManager
 from models.property import PropertyModel
 from models.tenant import TenantModel
 from models.tickets import TicketModel, TicketStatus
@@ -16,7 +19,7 @@ def seedData():
     now = datetime.utcnow()
     future = now + timedelta(days=365)
 
-    user_1 = UserModel(
+    user_1 = Admin(
         email="user1@dwellingly.org",
         role=RoleEnum.ADMIN,
         firstName="user1",
@@ -26,7 +29,7 @@ def seedData():
         archived=False,
     )
     user_1.save_to_db()
-    user_2 = UserModel(
+    user_2 = Admin(
         email="user2@dwellingly.org",
         role=RoleEnum.ADMIN,
         firstName="user2",
@@ -36,7 +39,7 @@ def seedData():
         archived=False,
     )
     user_2.save_to_db()
-    user_3 = UserModel(
+    user_3 = Admin(
         email="user3@dwellingly.org",
         role=RoleEnum.ADMIN,
         firstName="user3",
@@ -46,7 +49,7 @@ def seedData():
         archived=False,
     )
     user_3.save_to_db()
-    user_mister_sir = UserModel(
+    user_mister_sir = PropertyManager(
         email="MisterSir@dwellingly.org",
         role=RoleEnum.PROPERTY_MANAGER,
         firstName="Mr.",
@@ -56,7 +59,7 @@ def seedData():
         archived=False,
     )
     user_mister_sir.save_to_db()
-    user_gray_pouponn = UserModel(
+    user_gray_pouponn = PropertyManager(
         email="GrayPouponn@dwellingly.org",
         role=RoleEnum.PROPERTY_MANAGER,
         firstName="Gray",
@@ -106,7 +109,7 @@ def seedData():
         },
         schema=UserRegisterSchema,
     )
-    user_janice_joinstaff = UserModel(
+    user_janice_joinstaff = Staff(
         email="janice@joinpdx.org",
         role=RoleEnum.STAFF,
         firstName="Janice",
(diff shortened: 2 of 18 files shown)


models/user.py
```diff
@@ -104,6 +107,10 @@ def json(self):
             "created_at": Time.format_date(self.created_at),
             "updated_at": Time.format_date(self.updated_at),
         }
+        return {**base_json, **self.serialize()}
+
+    def serialize(self):
+        return {}
```
yunnie, line 113: Not a critique, but a question. Why does the `serialize` method return an empty dictionary? Is this just an abstract method for subclasses to implement so that the `json` method itself doesn't have to be overwritten by each subclass?

yunnie:
I found it helpful to read the SQLAlchemy documentation on [Inheritance Configuration](https://docs.sqlalchemy.org/en/13/orm/extensions/declarative/inheritance.html) before reviewing the code. It definitely makes the relationship among the various types of users much clearer.

models/user.py
NickSchimek: Only critiques accepted. No questions allowed 😉 

Thanks for asking @yunnie You are spot on, it kinda acts as an abstract method. This method is optional for our subclasses to implement. I've heard this referred to as a hook method. This allows the subclasses to hook into the json method without needing to know how it works. The other option as you mentioned is to have the subclass override the json method. Perhaps calling super and then needing to know how to take those results and combine it with the extra functionality the subclass wants to provide. That tends to lead to a lot of duplicate code and knowledge in the subclasses on how to interact with the json method. It is a better coding style to use hook methods instead. There are other reasons too.

tl/dr
it returns an empty hash because the json method still needs to function without subclasses. If I wanted to make serialize required then I've done this in the past.

```python
def serialize(self):
    raise NotImplementedError('Some message for devs')
```

FYI: Pending users will be in our system, and when they're serialized then they will be of type `UserModel` (since they haven't been assigned a user type/role yet), and so this serialize method will be called and the empty hash will be returned.

NickSchimek:
Thanks for linking to the documentation. I forgot to do that.
We're on version 1.4 -> Here is the link for that version https://docs.sqlalchemy.org/en/14/orm/inheritance.html

Unable to upgrade to 2.0 because of flask-sqlalchemy ... ... ...

models/user.py
```diff
@@ -32,7 +32,10 @@ def has_role(cls, role):
 class UserModel(BaseModel):
     __tablename__ = "users"
 
+    __mapper_args__ = {"polymorphic_identity": "users", "polymorphic_on": "type"}
```
aedwardg: Should the polymorphic identity here be `"user"` (singular)? All the others are singular, but this one is plural.
I'm not sure if it really matters or not, but it seems like it would be good to stick with one or the other.

aedwardg (commented):
This looks like a good start! Seems to be how the different User types should have been designed the whole time.

I just have the one question in my other comment, but other than that this looks good.

models/user.py
NickSchimek: Good catch. I'll get that updated.

NickSchimek:
> This looks like a good start! Seems to be how the different User types should have been designed the whole time.

Yeah perhaps, I'd be a bit hesitant to do this from the beginning without getting a better understanding of the system. I think it works here because we truly do have user that are of different types rather than users that need different roles assigned.

Most systems, I think, need users that can have different roles assigned. But I think systems like this benefit from having different types of users. I believe casa, which is similar to this system in that they have different types of users, also has a setup with different models acting as different types of users. Not in the way sqlalchemy does it, but in a way that their framework provides.

And just to clarify this doesn't prevent us from implementing roles and assigning roles to the various users, but as it is now we do not have the need to do that, and these types do act as a role too.

models/user.py
```diff
@@ -32,7 +32,10 @@ def has_role(cls, role):
 class UserModel(BaseModel):
     __tablename__ = "users"
 
+    __mapper_args__ = {"polymorphic_identity": "user", "polymorphic_on": "type"}
```
NickSchimek, line 35: I'm thinking this should be: thoughts?

```suggestion
    __mapper_args__ = {"polymorphic_identity": "pending_user", "polymorphic_on": "type"}
```

aedwardg: To me it makes more sense to keep it as "user".  Inheritance-wise, I tend to think of it as, e.g., "all property managers are also users" rather than "all property managers are also pending users".

NickSchimek: sounds good! Thanks for the input.

PR #312, opened 2021-06-09
updates ticket schema to include note

### LINKED ISSUE #620: Update Ticket Schema to accept an optional notes field
Update ticket schema to accept an optional Notes field. 

A ticket should be able to be created with a note or without a note.
Create schema tests to validate new behavior

Protopaco:
## Description
This adds a Note field to the Ticket schema, and includes testing.  

I did it without changing the Notes schema, with requires a user_id field and rejects a user field.  I had to do some manipulation to the Note values because of this. 

Let me know if I should increase the scope of this PR and alter the Notes schema as well.

Closes codeforpdx/dwellingly-app/issues/620

### DIFF
--- a/schemas/notes.py
+++ b/schemas/notes.py
@@ -1,10 +1,16 @@
+from ma import ma
 from models.user import UserModel
 from models.tickets import TicketModel
+from models.notes import NotesModel
 from utils.time import time_format
-from marshmallow import fields, Schema, validates, ValidationError
+from marshmallow import fields, validates, ValidationError, post_load, EXCLUDE
 
 
-class NotesSchema(Schema):
+class NotesSchema(ma.SQLAlchemyAutoSchema):
+    class Meta:
+        model = NotesModel
+        unknown = EXCLUDE
+
     id = fields.Integer()
     text = fields.Str()
     user_id = fields.Integer(required=True)
@@ -28,3 +34,7 @@ class NotesSchema(Schema):
     def validates_existing_ticket(self, value):
         if not TicketModel.query.get(value):
             raise ValidationError(f"{value} is not a valid Ticket ID")
+
+    @post_load
+    def make_notes(self, data, **kwargs):
+        return NotesModel(**data)
--- a/tests/schemas/test_ticket_schema.py
+++ b/tests/schemas/test_ticket_schema.py
@@ -22,10 +22,10 @@ class TestTicketValidations:
         valid_payload = {
             "tenant_id": create_tenant().id,
             "author_id": author_id,
-            "note": {**create_note().json(), "user_id": author_id},
+            "notes": {**create_note().json(), "user_id": author_id},
         }
 
-        del valid_payload["note"]["user"]
+        del valid_payload["notes"]["user"]
 
         no_validation_errors = {}
         assert no_validation_errors == TicketSchema().validate(valid_payload)
--- a/tests/schemas/test_ticket_schema.py
+++ b/tests/schemas/test_ticket_schema.py
@@ -4,12 +4,28 @@ from schemas import TicketSchema
 
 @pytest.mark.usefixtures("empty_test_db")
 class TestTicketValidations:
-    def test_valid_payload(self, create_tenant, create_admin_user):
+    def test_valid_payload_without_note(self, create_tenant, create_admin_user):
 
         valid_payload = {
             "tenant_id": create_tenant().id,
             "author_id": create_admin_user().id,
         }
+
+        no_validation_errors = {}
+        assert no_validation_errors == TicketSchema().validate(valid_payload)
+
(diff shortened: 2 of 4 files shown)


tests/schemas/test_ticket_schema.py
```diff
@@ -4,7 +4,7 @@
 
 @pytest.mark.usefixtures("empty_test_db")
 class TestTicketValidations:
-    def test_valid_payload(self, create_tenant, create_admin_user):
+    def test_valid_payload_without_note(self, create_tenant, create_admin_user):
```
NickSchimek, line 7: I don't think this method name needs to be renamed. For me `test_valid_payload` tests the minimum set of required attributes. A Ticket accepts many more attributes then simply a tenant_id and an author_id, but these are apparently the only attributes that are required for the payload to be valid. So let's not rename this, otherwise we could end up with a very long name if we're going to state every attribute that is not part of the valid payload in the method name.

tests/schemas/test_ticket_schema.py
```diff
@@ -13,3 +13,19 @@ def test_valid_payload(self, create_tenant, create_admin_user):
 
         no_validation_errors = {}
         assert no_validation_errors == TicketSchema().validate(valid_payload)
+
+    def test_valid_payload_with_note(
```
NickSchimek, line 17: I don't think this test is necessary. We already have a note schema that has validation tests for a note.

tests/schemas/test_ticket_schema.py
```diff
@@ -13,3 +13,19 @@ def test_valid_payload(self, create_tenant, create_admin_user):
...
+    ):
+
+        author_id = create_admin_user().id
+        valid_payload = {
+            "tenant_id": create_tenant().id,
+            "author_id": author_id,
+            "note": {**create_note().json(), "user_id": author_id},
+        }
+
+        del valid_payload["note"]["user"]
```
NickSchimek: This is a bit confusing. I'm not sure what the intent is here. Also your passing in a created note json fields to the valid_payload. 

The valid payload should represent what we would receive from the frontend, and looking at this I'm not quite sure what that is. I should be able to read the test and have a good understanding of the behavior of the application. Here, I'm not so sure what's going on with note.

schemas/notes.py
```diff
@@ -1,10 +1,16 @@
...
 from models.user import UserModel
 from models.tickets import TicketModel
+from models.notes import NotesModel
 from utils.time import time_format
-from marshmallow import fields, Schema, validates, ValidationError
+from marshmallow import fields, validates, ValidationError, post_load, EXCLUDE
 
 
-class NotesSchema(Schema):
+class NotesSchema(ma.SQLAlchemyAutoSchema):
```
NickSchimek, line 9: Originally we were not using an autoschema here and explicitly defining the field attributes. If we're going to switch this to an auto schema then I don't think we need all of these explicitly defined field attributes below.

schemas/notes.py
```diff
@@ -1,10 +1,16 @@
...
+    class Meta:
+        model = NotesModel
+        unknown = EXCLUDE
```
NickSchimek, line 12: I don't think this is necessary. Is there a reason for adding `unknown=EXCLUDE` ?

NickSchimek (changes requested):
I see some failing tests so I'm not sure if this was ready for me to review. 

I haven't really given much thought about updating the notes schema yet. If that's something you want to do in this PR then go for it, and it looks like you already went for it. I think for the postload function we only want that to run when a ticket is being created; otherwise it shouldn't run.

Protopaco:
> I see some failing tests so I'm not sure if this was ready for me to review.
> 
> I haven't really given much thought about updating the notes schema yet. If that's something you want to do in this PR then go for it, and it looks like you already went for it. I think for the postload function we only want that to run when a ticket is being created; otherwise it shouldn't run.

Thanks for bringing this back up.  I think I hit a wall with this one, and meant to come back to it.  I think I got a little confused when it came to the schemas and post_load function.  By the end I was just trying things to see if they would work.  

I've been trying to find the conversation we had about the post_load function, but it might have been during a meeting.  I remembered that you suggested I look to emergency contacts as an example.

I started over from scratch, but it still isn't working. I think my approach might be flawed.  What I was trying to do is, inside of the TicketSchema, create a method that, if a note is passed, will create a new one.  I am not seeing a way to pass the newly created ticket.id to the NoteModel so it can create a new note.  It also seems like in the flow of creating a ticket, the schema is the inappropriate place have that functionality.

Would it make more sense to include that functionality in the resource, like we just did for leases in the create tenant endpoint?  That way if notes are included in the response, they can be created after the ticket is created.  Does that make sense?

NickSchimek:
Closing this as there seems to be a new PR for this issue #339

PR #313, opened 2021-06-10
changed note serialization format to iso

### LINKED ISSUE #615: Change to iso format when serializing a note object for created_at/updated_at fields
## Description

The frontend would like to order the notes by the created_at date. That is a bit difficult to do with the current format that it is being sent in. Update the format to use the iso date format from the Time class.

yunnie:

Closes codeforpdx/dwellingly-app/issues/615

### DIFF
--- a/models/notes.py
+++ b/models/notes.py
@@ -18,6 +18,6 @@ class NotesModel(BaseModel):
             "ticket_id": self.ticket_id,
             "text": self.text,
             "user": self.user.full_name(),
-            "created_at": Time.format_date(self.created_at),
-            "updated_at": Time.format_date(self.updated_at),
+            "created_at": Time.to_iso(self.created_at),
+            "updated_at": Time.to_iso(self.updated_at),
         }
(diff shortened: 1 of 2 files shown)


yunnie:
I didn't see a centralized place to change data serialization. It looks like all of the serialization is handled in the various `.json()` methods.

aedwardg (approved):
Looks good to me! Thanks!

NickSchimek:
The centralized place is here.

https://github.com/codeforpdx/dwellinglybackend/blob/60971ea6981bc6a850cddc847422745c9113ea44/utils/time.py#L4-L5

Basically we would remove iso_format and just change time_format to be the iso_format. That would take care of 98% of it. <- I took the time to fact check this calculation so no need to check yourself. Because I think there are only one or two places where we don't use format_date. Now probably 3.

But we should probably do this in a different PR. And then this PR would end up being reverted back to the way it was. Using the format_date.

PR #316, opened 2021-06-14
627 edit note

### LINKED ISSUE #627: Adds update note endpoint & tests
## Description
Create an update action that will update a note resource.
This will be a patch method

Use the `update` method from the base model

Create test that asserts the update method is called with correct arguments when the endpoint is called.

**Additional context**
Tied to:
https://github.com/codeforpdx/dwellingly-app/issues/626

Protopaco:
## Description
This is my first attempt at the update notes endpoint.  

I added an .update function to the Nobiru list, to add in that redundancy.  I couldn't find any built in sqlalchemy update functions, so I removed the old entry and added the new one.  

I also feel like my note endpoint functionality will likely need to be improved.  It works properly, and I couldn't figure out a better way to update the the notes data, but it is still pretty clunky.

Let me know what you think!

Closes codeforpdx/dwellingly-app/issues/627


Protopaco:
I attempted to rebase this branch onto the latest version of development.  I'm not sure if I did it properly, please let me know if there is a better way for me to delete those older commits.  Thanks!

nobiru/nobiru_list.py
```diff
@@ -16,3 +16,12 @@ def delete(self, entity):
...
             abort(404, f"{entity._name()} not found")
+
+    @collection.replaces(1)
+    def update(self, entity, new_entity):
+        try:
+            self.remove(entity)
+            self.append(new_entity)
+            db.session.commit()
+        except ValueError:
+            abort(404, f"{entity.name()} note found")
```
NickSchimek: This is good. I kinda don't like the replace behavior because it will change the Notes id, but looking at the test it doesn't seem to do that in this case. I'm not 100% sure though, because of how sqlalchemy works.

Let's assume it does modify the id then I would probably say this: Of course there is not really anything wrong with that. Sometimes it is hard to avoid when updating collections. However, I would like to avoid it if possible. I'm pretty sure we'll have use for this. I'm not sure if we do right now. Either way this is really cool! I didn't know the `@collection.replace(1)` exists.

tests/integration/test_notes.py
```diff
@@ -50,3 +48,27 @@ def test_it_deletes_one_note(
         mock_delete.assert_called_once_with(note)
         assert response.status_code == 200
         assert response.json == {"message": "Note deleted"}
+
+
+@pytest.mark.usefixtures("client_class", "empty_test_db")
+class TestUpdate:
+    def setup(self):
+        self.endpoint = "/api/tickets"
+        self.new_text = "We don't need no water"
```
NickSchimek, line 57: Yes we do

tests/integration/test_notes.py
```diff
@@ -50,3 +48,27 @@ def test_it_deletes_one_note(
...
+
+    def test_it_updates_one_note(self, valid_header, create_note):
+        note = create_note()
+        # ticket = note.ticket
+
+        # with patch.object(ticket.notes, "update") as mock_update:
```
NickSchimek: Let's clean up this commented out code.

resources/notes.py
```diff
@@ -26,4 +26,14 @@ class Note(Resource):
...
         return {"message": "Note deleted"}
+
+    @pm_level_required
+    def patch(self, ticket_id, id):
+        ticket = TicketModel.find(ticket_id)
+        updated_note = NotesModel.update(
+            schema=NotesSchema, id=id, payload={"text": request.json["text"]}
+        )
+        ticket.notes.update(NotesModel.find(id), updated_note)
+        return updated_note.json()
```
NickSchimek: Let me know your thoughts on this. I think the flow I would like to see is something like this:

```python
ticket = TicketModel.find(ticket_id)
note = ticket.notes.find(id)
return note.update(schema=NotesSchema, payload=request.json).json()
```
What this should do is automatically return a 404 if ticket or note is not found.
If note exists then we can update it and serialize the new note. And then of course Marshmallow will handle any validation errors for us automatically.

What do you think?

NickSchimek (changes requested):
I think this works and is really great! Thanks for the description. I outlined what I would like to see in the controller. Let me know if thats not helpful though.

nobiru/nobiru_list.py
Protopaco: Yeah, I did a dive into the sqlalchemy docs to try and find a straight update collections function like you could do with postgres, but I couldn't find one.  I was hoping that someone else might have encountered one.  

I think that because we are prescribing the note id in the resource, and then appending it straight into the db, that is keeping the id consistent.  If we used the NotesModel.create() method it would create a new note id during that process.  

I will cut it out for now.

resources/notes.py
Protopaco: So add a find method to the nobiru list, and then use the note.update to actually change the note entry in the db?  I think that makes a lot of sense, and less clunky than my solution.  😆

Protopaco:
So I pushed up a different solution to this problem.  I'm not sure if it is cleaner, but it does what is asked.  😄   

I couldn't find any built in collection methods for finding an item.  I am sure they must exist, but I couldn't find one.  So instead , because ticket.notes is an InstrumentedList, and we are searching for the element within that list by it's id, I just did a quick search through the list, returning the first element that the passed id.  I wanted the solution to be universal, in case we needed it for other scenarios. 

Please let me know if there is a built in method, or query rather than just looping through the list.

Thanks!

nobiru/nobiru_list.py
```diff
@@ -16,3 +16,10 @@ def delete(self, entity):
             db.session.commit()
         except ValueError:
             abort(404, f"{entity._name()} not found")
+
+    def find(self, id):
+        try:
+            return next(x for x in self if x.json()["id"] == id)
```
NickSchimek: `x.json()["id"] == id`
We shouldn't be serializing an object to extract an attribute for comparison in our code. Serialization methods are the data representation of an object meant for frontend consumption, not for being used as logic in our code. We may not be serializing the id of all objects in the code base, hence this will fail when an object doesn't serialize the id attribute.

nobiru/nobiru_list.py
```diff
@@ -16,3 +16,10 @@ def delete(self, entity):
...
+
+        except StopIteration:
+            abort(404, "Not found")
```
NickSchimek: `Not found` is a big generic. Is there any way we can specify what was not found?

resources/notes.py
```diff
@@ -26,4 +26,12 @@ class Note(Resource):
...
         ticket.notes.delete(NotesModel.find(id))
+
         return {"message": "Note deleted"}
+
+    @pm_level_required
+    def patch(self, ticket_id, id):
+        ticket = TicketModel.find(ticket_id)
+        note = ticket.notes.find(id)
+
+        return note.update(schema=NotesSchema, payload=request.json, id=id).json()
```
NickSchimek: This looks good. The `update` method looks like its an instance method here, but it is actually calling the class method which I find a bit confusing. Also the id parameter seems a bit weird. If we are calling update on note then update should know that the object being updated is note, and shouldn't need the id. 

**Option A.**
I would like to see this slightly improved. Either call update on the class and make it clear which id is being used.
```python
NoteModel.update(schema=NotesSchema, payload=request.json, id=note.id
```

However, I don't quite like that, because for one: We already have the note object. And this class method will hit the database again to find the note object. When I created these class method they had made sense at the time, but they're not static. They can be evolved and improved.

**Option B.**
I think I would rather see an instance method called `update` that updates an instance.

```python
note.update(schema=NotesSchema, payload=request.json)
```

**Extra Credit**
It would also be nice to not have to pass in the schema argument and be able to either infer that or ask the object for the schema.

```python
note.update(request.json)
```

tests/integration/test_notes.py
```diff
@@ -50,3 +48,34 @@ def test_it_deletes_one_note(
...
+            f"{self.endpoint}/{note.ticket_id}/notes/{note.id}",
+            json={"text": self.new_text},
+            headers=valid_header,
+        )
+
+        assert response.status_code == 200
+        assert response.json["id"] == note.id
+        assert response.json["text"] == self.new_text
+
+    def test_it_updates_one_note_fails(self, valid_header, create_ticket):
```
NickSchimek: I don't think this test is necessary.

NickSchimek: oh I see. Maybe you are trying to test the `find` implementation on NobiruList. Then I think this test is necessary, but it is in the wrong test file.

NickSchimek (changes requested):
I just made some more comments. They don't all need to be addressed. Good work!

nobiru/nobiru_list.py
Protopaco: I hear you.  I'm a little stumped on how to find a note, and verify that it belongs to the ticket, and make it universally applicable.  it doesn't appear as if sqlalchemly has an built in methods to find data within an InstrumentedList.  The only other solution I can think of is to kill this avenue, and do a verification at the endpoint level.  Something akin to:
```python
    def patch(self, ticket_id, id):
        ticket = TicketModel.find(ticket_id)
        note = NotesModel.find(id)

        if note.ticket_id == ticket.id:
            return note.update(schema=NotesSchema, payload=request.json, id=id).json()

        return {"message": "Note not found"}, 404
```

resources/notes.py
Protopaco: That is all possible, but wouldn't mean messing with the .update method in the BaseModel? I'm happy to do it, but I wanted to make sure that is what you are envisioning.

nobiru/nobiru_list.py
NickSchimek: well, I think your original implementation of the find method is good. Let's just not serialize it for comparison.
```suggestion
            return next(x for x in self if x.id == id)
```

Does this work?

resources/notes.py
NickSchimek: Let me think about this some more. I'm very unnerved by the fact that you can call a class level method from an instance of the object. Which means there is no separation between class and instance methods in python. It looks like they all share the same namespace. That does not make me a happy programmer.

`note.update(schema=NotesSchema, payload=request.json, id=id)` <- This looks like an instance method, but it's a class method.

nobiru/nobiru_list.py
Protopaco: Oh!  That does work.  Sorry I misunderstood. Thanks!

NickSchimek: Here's the documentation on how to do this with a sql query.
https://docs.sqlalchemy.org/en/14/tutorial/orm_related_objects.html#common-relationship-operators

resources/notes.py
NickSchimek: I put in PR #329 to address this issue. Review it. Let me know.

nobiru/nobiru_list.py
Protopaco: I didn't realize you could do a straight sql query with sqlalchemy.  Do we want to do a query in this instance, or does just using x.id work well enough?

tests/nobiru/test_nobiru_list.py
```diff
@@ -65,3 +65,23 @@ def test_delete_with_non_existant_entity(self):
             err.exconly()
             == "werkzeug.exceptions.NotFound: 404 Not Found: DummyRelation not found"
         )
+
+    def test_find(self):
+        fake = self.create_fake_model()
+        value_id = fake.values[0].id
```
NickSchimek: I'm looking at this, and I forgot I had wrote these helpers. Just thinking out loud here. I think values is kind of a confusing name. I think children would be more clear. Because it would be the associated children of the fake model. I think values is not clear what its doing unless you have all the context in your head. What do you think?

tests/nobiru/test_nobiru_list.py
```diff
@@ -65,3 +65,23 @@ def test_delete_with_non_existant_entity(self):
...
+
+        with patch.object(fake.values, "find") as mock_find:
+            fake.values.find(value_id)
+
+        mock_find.assert_called_once_with(value_id)
```
NickSchimek: I don't think it makes sense to do patching here. Essentially this is patching the find method that is executing in the test. So none of the application logic actually runs.

I think what we want to test is that the object is returned when executed. The behavior of the find method can be defined as it returns the object for the provided id; otherwise it returns an error message when not found.
```suggestion
        value = fake.values[0]
        assert value == fake.values.find(value.id)
```

NickSchimek (changes requested):
This looks awesome! Thanks for figuring all this out with me. It will sure help scope things in the future. Just one small change requested for the test. This look great! Thanks again 💫

tests/nobiru/test_nobiru_list.py
Protopaco: Sorry  I didn't get back on this, I've been distracted by the new job for the last couple of weeks.  Did you still want to change this to children?

NickSchimek: No worries, I'm more interested to know if you would think that wording would make more sense or not. I'm ok to leave it as is for now unless you have strong opinion about changing the wording

PR #317, opened 2021-06-17
629 refactor test_emergency_contact_authorization to no auth headers

### LINKED ISSUE #629: Use role-level fixtures in `test_emergency_contact_authorization.py`
### Purpose
**Step 3 of 3** in moving away from using `auth_headers` in the integration tests (sub-issue of #547)

Authorization tests are for testing whether a user has the role/permissions to perform a certain action.  To do this, we don't need to hit the login endpoint every time (which is what `auth_headers` is doing).   For more information, see #545 

### Proposed Solution
Replace the `auth_headers` fixture with with the appropriate role-level fixture (`admin_header`, `staff_header`, `pm_header`) in **`test_emergency_contact_authorization.py`** tests. This is located in a separate `authorizations` test directory.  

NOTE:
- This may require altering the tests to use `empty_test_db` as well if the test doesn't already.
- There is no header fixture for "pending" users, because these users have no role and will not be authenticated.  To test that a pending user cannot perform an action, write the API call with no header information.

### Acceptance Criteria
- [ ] All integration tests in `test_emergency_contact_authorization.py` (i.e., checks if the user has permission to perform the action) should use the appropriate role-level fixture instead of `auth_headers`.

ovsyankimo:
## Description
Refactor test_emergency_contact_authorization.py to use role-level fixtures, not auth_headers fixture.
Closes codeforpdx/dwellingly-app/issues/629

### DIFF
--- a/tests/authorizations/test_emergency_contact_authorization.py
+++ b/tests/authorizations/test_emergency_contact_authorization.py
@@ -7,10 +7,7 @@ class TestEmergenyContactsAuthorizations:
     def setup(self):
         self.endpoint = "/api/emergencycontacts"
 
-    def test_emergency_contacts_POST(self, auth_headers):
-        endpoint = "/api/emergencycontacts"
-
-        newContact = {
+        self.newContact = {
             "name": "Narcotics Anonymous",
             "description": "Cool description",
             "contact_numbers": [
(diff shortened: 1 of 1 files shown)


tests/authorizations/test_emergency_contact_authorization.py
```diff
@@ -19,23 +16,52 @@ def test_emergency_contacts_POST(self, auth_headers):
             ],
         }
 
+        self.user_id = 1
+
+    def test_auth_header_is_required(self):
+        response = self.client.post(self.endpoint, json=self.newContact)
+
+        assert is_valid(response, 401)
```
aedwardg, line 24: Let's change all of these over to
```python
assert response.status_code == 401
```

aedwardg (commented):
Looks great @ovsyankimo!

I do think we should probably change all of these `is_valid()` assertions over to just checking the status code though since `is_valid` isn't really testing anything useful (see https://github.com/codeforpdx/dwellinglybackend/pull/209#discussion_r575921252).  We've been slowly doing this over the last few months, though that's not immediately apparent by looking at the tests.

It might also be good to use the `emergency_contact_attributes` fixture which uses Faker instead of using the hardcoded `self.newContact`.  That wasn't in the issue description though so that could always be done in a different ticket instead.

tests/authorizations/test_emergency_contact_authorization.py
NickSchimek: I'm fine for leaving is_valid in for now. Flask-restful guarantees that it will be valid json response, If we switch off of flask-restful then... I'm fine either way here.

NickSchimek (approved):
Thanks for the PR @ovsyankimo and thanks for reviewing @aedwardg

PR #318, opened 2021-06-20
Improvemen user model

qq1u:
## Description
The "_hash_digest " field is now deleted and the "password " field is used

Closes codeforpdx/dwellingly-app/issues/###

### DIFF
--- a/models/user.py
+++ b/models/user.py
@@ -38,8 +38,8 @@ class UserModel(BaseModel):
     firstName = db.Column(db.String(100), nullable=False)
     lastName = db.Column(db.String(100), nullable=False)
     phone = db.Column(db.String(20), nullable=False)
-    hash_digest = db.Column(db.LargeBinary(60))
-    password = db.Column(db.String(), default=None, onupdate=None)
+    # hash_digest = db.Column(db.LargeBinary(60))
+    _password = db.Column('password', db.LargeBinary(60))
     archived = db.Column(db.Boolean, default=False, nullable=False)
     lastActive = db.Column(db.DateTime, default=datetime.utcnow)
 
@@ -56,10 +56,16 @@ class UserModel(BaseModel):
         collection_class=NobiruList,
     )
 
-    def __init__(self, **kwargs):
-        super(UserModel, self).__init__(**kwargs)
-        self.hash_digest = UserModel._hash_pw(kwargs["password"])
-        self.archived = False
+    @property
+    def password(self):
+        return self._password
+
+    @password.setter
+    def password(self, plaintext_password):
+        self._password = bcrypt.hashpw(
+            plaintext_password.encode('utf-8'),
+            bcrypt.gensalt(current_app.config["WORK_FACTOR"]),
+        )
 
     def update_last_active(self):
         self.lastActive = datetime.utcnow()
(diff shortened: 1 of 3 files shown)


aedwardg (dismissed):
This is nice, very pythonic!
Great job @13528080556!

tests/unit/test_user.py
```diff
@@ -116,8 +116,7 @@ def test_update_class_method(self, create_join_staff, faker):
         UserModel.update(UserSchema, user.id, {"email": email})
         lookedup = UserModel.find(user.id)
         assert lookedup.email == email
-        assert lookedup.password is None
-        assert lookedup.hash_digest == user.hash_digest
+        assert lookedup.password == user.password
```
NickSchimek: I think we can delete this assertion. We're now no longer need to test behavior that we implemented. This is testing behavior provided by sqlalchemy.

models/user.py
```diff
@@ -81,15 +86,8 @@ def validate_reset_password(token):
...
-    @staticmethod
-    def _hash_pw(plaintext_password):
-        return bcrypt.hashpw(
-            bytes(plaintext_password, "utf-8"),
-            bcrypt.gensalt(current_app.config["WORK_FACTOR"]),
-        )
-
     def check_pw(self, plaintext_password):
-        return bcrypt.checkpw(bytes(plaintext_password, "utf-8"), self.hash_digest)
+        return bcrypt.checkpw(plaintext_password.encode("utf-8"), self.password)
```
NickSchimek: I think I would rather see this: And then delete the password getter. As password should not be needed outside of this class.
```suggestion
        return bcrypt.checkpw(plaintext_password.encode("utf-8"), self._password)
```

models/user.py
```diff
@@ -56,10 +55,16 @@ class UserModel(BaseModel):
         collection_class=NobiruList,
     )
 
-    def __init__(self, **kwargs):
-        super(UserModel, self).__init__(**kwargs)
-        self.hash_digest = UserModel._hash_pw(kwargs["password"])
-        self.archived = False
+    @property
+    def password(self):
+        return self._password
```
NickSchimek: I think this can be removed. It's not necessary to ever access the password outside of this object.

NickSchimek (changes requested):
This looks great! Thanks for doing this.
I have a couple small suggestions. Let me know if they wont work for some reason.

models/user.py
```diff
@@ -38,8 +38,7 @@ class UserModel(BaseModel):
     firstName = db.Column(db.String(100), nullable=False)
     lastName = db.Column(db.String(100), nullable=False)
     phone = db.Column(db.String(20), nullable=False)
-    hash_digest = db.Column(db.LargeBinary(60))
-    password = db.Column(db.String(), default=None, onupdate=None)
+    _password = db.Column("password", db.LargeBinary(60))
```
NickSchimek, line 44: What does this do `db.Column("password", db.LargeBinary(60)`?
I'm not familiar with passing a string as the first argument.

qq1u: String as the first argument, which will be the name of the field in the table.

models/user.py
```diff
@@ -59,10 +58,16 @@ class UserModel(BaseModel):
         collection_class=NobiruList,
     )
 
-    def __init__(self, **kwargs):
-        super(UserModel, self).__init__(**kwargs)
-        self.hash_digest = UserModel._hash_pw(kwargs["password"])
-        self.archived = False
+    @property
+    def password(self):
+        raise AttributeError("password field is not allowed to access")
```
NickSchimek, line 63: I like this better! Good work!

NickSchimek (approved):
Excellent! Thanks 🙏

PR #319, opened 2021-06-24
Use `valid_header` for emergency contacts integration tests

### LINKED ISSUE #594: Use `valid_header` for Emergency Contacts integration tests
### Purpose

### Proposed Solution
Replace the `auth_headers` fixture with `valid_header` in all integration tests in **`tests/integration/test_emergency_contacts.py`**.  

### Acceptance Criteria

gussmith23:
## Description

Replaces `auth_headers` with `valid_header` in emergency contacts tests, so that we can move away from testing authorization in these tests.

Additionally, uses `test_database` where necessary. The original issue had mentioned using `empty_test_db`, but I found that the tests depended on the populated test db.

This PR seems to introduce two warnings on the emergency contacts tests, but I can't figure out how to view the warnings with pytest -_-

Closes codeforpdx/dwellingly-app/issues/594

### DIFF
--- a/tests/integration/test_emergency_contacts.py
+++ b/tests/integration/test_emergency_contacts.py
@@ -13,11 +13,6 @@ import pytest
 # 1. Action on non-existent entries
 # 2. Duplicate creation
 
-# Also test for appropriate auth restriction, such as:
-# 1. Valid access token
-# 2. Valid role (admin, property-manager, pending, etc.)
-
-
 endpoint = "/api/emergencycontacts"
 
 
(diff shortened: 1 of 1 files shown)


tests/integration/test_emergency_contacts.py
```diff
@@ -113,19 +104,17 @@ def test_emergency_contacts_PUT(client, auth_headers):
...
-    response = client.put(
-        f"{endpoint}/{id}", json=updatedInfo, headers=auth_headers["admin"]
-    )
+    response = client.put(f"{endpoint}/{id}", json=updatedInfo, headers=valid_header)
     assert is_valid(response, 404)  # NOT FOUND
     assert response.json == {"message": "Emergency contact not found"}
 
 
-def test_emergency_contacts_DELETE(client, auth_headers):
+def test_emergency_contacts_DELETE(test_database, client, valid_header):
```
NickSchimek, line 112: Is there a reason for adding test_database? I wouldn't think it's required because it is already running given the nature of how these tests are setup.

NickSchimek: haha nevermind, I forget it was because of auth_headers is why test_database was not needed before.

NickSchimek (approved):
Excellent! Thanks for the PR!

PR #321, opened 2021-06-29
Add gatekeeper utility to check request params

### LINKED ISSUE #642: Create architecture/methods for restricting the parameters being sent to the model from the controller
## Description

Currently we send all request parameters to the model for updating/creating a resource. This is not an issue, because we explicitly tell Marshmallow to ignore and drop any parameters that are not defined in the schema, however it does allow the requestor to send parameters that we do not want them to be able to update. The way marshmallow wants you to do things is to create multiple schemas whenever you want to allow different parameters for any given request. This seems great, but marshmallow is already doing many things and I think we can simplify things by removing this responsibility from marshmallow and allow it to do what it does best. validate the parameters that we do send it.

I believe it would be better to declare which parameters are allowed for any particular controller action rather than sending everything to marshmallow to figure out. 

Some questions we may want to ask:
Do we want to raise errors if the endpoint receives any parameters that are not permitted or simply ignore them and try to process the request. I think for development purposes it is good to raise errors, because it then tells the developer that if they want that parameter then it needs to be added to the accepted list rather than having to try to figure out why something isn't working.

### Task
 - Update the Put action in the Tickets resource to only send the allowed parameters to marshmallow.
 The following should be the allowed parameters for updating a ticket:
    - issue
    - urgency
    - status
    - tenant_id
 - Optionally raise errors if any parameter sent is not allowed
 - test as needed

aedwardg:
## Description

Use the `@allowed_params` decorator to restrict an endpoint to only
accept certain fields in the JSON request.

Decorator takes a set of strings as its only argument.

Closes codeforpdx/dwellingly-app/issues/642

### DIFF
--- a/tests/integration/test_gatekeeper.py
+++ b/tests/integration/test_gatekeeper.py
@@ -1,4 +1,5 @@
 import pytest
+from flask import current_app, request
 from tests.factory_fixtures.dummy_resource import DummyResource
 from utils.gatekeeper import GatekeeperError
 
@@ -7,17 +8,20 @@ from utils.gatekeeper import GatekeeperError
 class TestGatekeeper:
     def test_empty_set_allows_no_params(self):
         with pytest.raises(GatekeeperError):
-            self.client.put("/api/dummy", json={"hi": "there"})
+            with current_app.test_request_context(json={"hi": "there"}):
+                DummyResource.put(request)
 
     def test_raises_error_for_unspecified_params(self):
         DummyResource.dummy_params.update({"hello", "world"})
 
         with pytest.raises(GatekeeperError) as error_info:
-            self.client.put("/api/dummy", json={"hi": "there"})
+            with current_app.test_request_context(json={"hi": "there"}):
+                DummyResource.put(request)
         assert "Invalid request field" in str(error_info.value)
 
         DummyResource.dummy_params.clear()
 
     def test_subset_of_allowed_params_is_okay(self):
         DummyResource.dummy_params.update({"hello", "world", "again"})
-        self.client.put("/api/dummy", json={"hello": "there"})
+        with current_app.test_request_context(json={"hello": "there"}):
+            DummyResource.put(request)
--- a/tests/integration/test_gatekeeper.py
+++ b/tests/integration/test_gatekeeper.py
@@ -1,5 +1,5 @@
 import pytest
-from flask import current_app, request
+from flask import current_app
 from tests.factory_fixtures.dummy_resource import DummyResource
 from utils.gatekeeper import GatekeeperError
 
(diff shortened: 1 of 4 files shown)


utils/gatekeeper.py
```diff
@@ -0,0 +1,33 @@
...
+        return verify_fields
+
+    return decorator
+
+
+def invalid_field_error(field_set):
+    if current_app.env == "development":
+        raise GatekeeperError(field_set)
+    else:
+        return {"message": "Invalid request field"}, 400
```
NickSchimek: Let's switch this around. That way we raise the useful error in dev and test.
```suggestion
    if current_app.env == "production":
        return {"message": "Invalid request field"}, 400
    else:
        raise GatekeeperError(field_set)
```

Also, I'm not fond of using the string `"production"` for equality matcher.

That could change or easily be misspelled. I would like to define some constants in the env file for use across the app instead.

NickSchimek (commented):
This looks good! I have one suggestion below. I'd say it would be best writing some integration tests to check this functionality, but I don't want it being tested on any of the app resources. It would be good to create a Dummy resource in a test that we could check.

This code will, however, always be executing during all the integration tests we have. So with that in mind we can probably pass on the integration Dummy test if you have difficulties setting it up. If no integration tests then some unit tests would be good if possible. Otherwise, as long as you've manually tested it then I'm not too concerned.

tests/integration/test_gatekeeper.py
```diff
@@ -0,0 +1,23 @@
+import pytest
+from tests.factory_fixtures.dummy_resource import DummyResource
+from utils.gatekeeper import GatekeeperError
+
+
+@pytest.mark.usefixtures("client_class")
+class TestGatekeeper:
+    def test_empty_set_allows_no_params(self):
+        with pytest.raises(GatekeeperError):
+            self.client.put("/api/dummy", json={"hi": "there"})
```
NickSchimek: I would like to not have to add logic to the routes file simply for testing if possible. We don't need to test routing so could we bypass it by calling the method directly?

DummyResource().put ?

aedwardg: That makes sense.  The only thing is, the decorator looks at `request.json.keys()`, so I'll have to figure out how to properly mock the request object and pass it to that method.

aedwardg: Alright, I updated it.  I think it's a lot better now.  Let me know what you think 😄

utils/gatekeeper.py
```diff
@@ -0,0 +1,33 @@
...
+            return endpoint(*args, **kwargs)
+
+        return verify_fields
+
+    return decorator
+
+
+def invalid_field_error(field_set):
+    if current_app.env == "production":
+        return {"message": "Invalid request field"}, 400
```
NickSchimek, line 21: Another option here would be to return the endpoint with an argument that consisted of only the subset of allowed params, essentially filtering out out any unwanted params. And then we can always use that argument in the action instead of request.json - I'm not sure if that's possible though.

NickSchimek (approved):
Excellent work! Thanks for doing this. It looks good :-)

PR #339, opened 2021-07-12
First step towards updating ticket schema

### LINKED ISSUE #620: Update Ticket Schema to accept an optional notes field

Protopaco:
## Description
I started over on my run at this ticket.  I feel like I've made some progress, but there is something crucial I seem to be missing.  

I added the post_load to the NotesSchema in CreateNotesSchema.   I also added the list of notes to the TicketSchema, that references CreateNotesSchema.   I also added a Ticket Schema Test to see if my additions worked.

The test isn't causing any errors, but it doesn't seem to be returning the intended information.  The ticket_json includes a list of notes objects, but they appear to be dictionaries rather than NotesModel instances as is intended.  It also appears that the post_load function is not being called, which is probably the problem.  

I've been working on this for a couple of hours and I can't figure out how to call that post_load function, so it can return the full NotesModel.  Does anyone else have an idea of what I am missing?

Closes codeforpdx/dwellingly-app/issues/620

### DIFF
--- a/schemas/notes.py
+++ b/schemas/notes.py
@@ -1,6 +1,7 @@
 from models.user import UserModel
 from models.tickets import TicketModel
-from marshmallow import fields, Schema, validates, ValidationError
+from models.notes import NotesModel
+from marshmallow import fields, Schema, validates, ValidationError, post_load
 
 
 class NotesSchema(Schema):
@@ -24,3 +25,9 @@ class NotesSchema(Schema):
     def validates_existing_ticket(self, value):
         if not TicketModel.query.get(value):
             raise ValidationError(f"{value} is not a valid Ticket ID")
+
+
+class CreateNotesSchema(NotesSchema):
+    @post_load
+    def create_note(self, data, **kwargs):
+        return NotesModel.create(**data)
--- a/tests/schemas/test_ticket_schema.py
+++ b/tests/schemas/test_ticket_schema.py
@@ -1,6 +1,7 @@
 import pytest
 from schemas import TicketSchema
-from models.user import UserModel
+from models.tickets import TicketModel
+from tests.attributes import note_attrs
 
 
 @pytest.mark.usefixtures("empty_test_db")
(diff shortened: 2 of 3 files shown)


tests/schemas/test_ticket_schema.py
```diff
@@ -13,3 +14,18 @@ def test_valid_payload(self, create_tenant, create_admin_user):
...
+class TestTicketSerialization:
+    def test_ticket_serialization(
+        self, create_ticket, create_note, ticket_attributes, note_attributes
+    ):
+        ticket = create_ticket()
+        user = UserModel.query.filter_by(id=ticket.author_id).first()
+        create_note(user, ticket, "text")
+
+        ticket_schema = TicketSchema()
+        ticket_json = ticket_schema.dump(ticket)
```
NickSchimek: We don't use marshmallow for Serializing objects. At least not yet. 
What we use it for is input validation and Deserialization.

For this test we want to test that the schema Deserializes the input correctly.

What I've done in the past is to use the same method we would use when creating an object in our endpoint. Which should be the `create` method. And then check if the object is created successfully. That's technically not a unit test, but our `create` method is dependable and doesn't change that often so I think it's OK to use in this test, and it's a better test than a pure unit test. 

The property and tenant schema test files have examples of this.

NickSchimek (changes requested):
Sorry it took me so long to review this. See my comment in the test section.

I can also finish this up for you too since I know you've been busy lately.

NickSchimek:
@Protopaco I updated the test to fail appropriately. The next step is to update the schema to get the test to pass.

Let me know if you'd like to finish it off. Thanks!

schemas/notes.py
```diff
@@ -24,3 +25,9 @@ def validate_existing_user(self, value):
     def validates_existing_ticket(self, value):
         if not TicketModel.query.get(value):
             raise ValidationError(f"{value} is not a valid Ticket ID")
+
+
+class CreateNotesSchema(NotesSchema):
+    @post_load
+    def create_note(self, data, **kwargs):
+        return NotesModel.create(**data)
```
NickSchimek, line 33: I believe the post_load should belong on the Tickets schema instead of the notes schema.

PR #342, opened 2021-07-18
update revoked_token id field type

### LINKED ISSUE #662: Type error. storing/retrieving string as an id
**Describe the bug**
We have been using sqllite which doesn't enforce types. So this bug goes unnoticed.

Using postgres as the database the following error occurs
```python
self = <sqlalchemy.dialects.postgresql.psycopg2.PGDialect_psycopg2 object at 0x113ef6400>, cursor = <cursor object at 0x113b11220; closed: -1>
statement = 'SELECT revoked_tokens.created_at AS revoked_tokens_created_at, revoked_tokens.updated_at AS revoked_tokens_updated_at...revoked_tokens_id, revoked_tokens.jti AS revoked_tokens_jti \nFROM revoked_tokens \nWHERE revoked_tokens.id = %(pk_1)s'
parameters = {'pk_1': '7a2ab50d-8d3e-4278-9a18-ec65c1cfe879'}, context = <sqlalchemy.dialects.postgresql.psycopg2.PGExecutionContext_psycopg2 object at 0x113f47ac0>

    def do_execute(self, cursor, statement, parameters, context=None):
>       cursor.execute(statement, parameters)
E       sqlalchemy.exc.DataError: (psycopg2.errors.InvalidTextRepresentation) invalid input syntax for type integer: "7a2ab50d-8d3e-4278-9a18-ec65c1cfe879"
E       LINE 3: WHERE revoked_tokens.id = '7a2ab50d-8d3e-4278-9a18-ec65c1cfe...
E                                         ^
E
E       [SQL: SELECT revoked_tokens.created_at AS revoked_tokens_created_at, revoked_tokens.updated_at AS revoked_tokens_updated_at, revoked_tokens.id AS revoked_tokens_id, revoked_tokens.jti AS revoked_tokens_jti
E       FROM revoked_tokens
E       WHERE revoked_tokens.id = %(pk_1)s]
E       [parameters: {'pk_1': '7a2ab50d-8d3e-4278-9a18-ec65c1cfe879'}]
E       (Background on this error at: https://sqlalche.me/e/14/9h9h)

../../../.local/share/virtualenvs/dwellinglybackend-S6sCqKnp/lib/python3.8/site-packages/sqlalchemy/engine/default.py:717: DataError
```

**To Reproduce**
Run this test with a postgresql database:
- `tests/integration/test_users.py::TestUserLogin::test_refresh_user`

**Expected behavior**
Test should pass.

qq1u:
fix 662

Closes codeforpdx/dwellingly-app/issues/662

### DIFF
--- a/models/revoked_tokens.py
+++ b/models/revoked_tokens.py
@@ -5,9 +5,10 @@ from models.base_model import BaseModel
 class RevokedTokensModel(BaseModel):
     __tablename__ = "revoked_tokens"
 
-    id = db.Column(db.String(36), primary_key=True)
-    jti = db.Column(db.String(120), nullable=False)
+    id = db.Column(db.Integer, primary_key=True)
+    jti = db.Column(db.String(36), nullable=False)
 
     @classmethod
-    def is_jti_blacklisted(cls, jti):
-        return bool(cls.query.get(jti))
+    def is_jti_revoked(cls, jti):
+        token = cls.query.filter_by(jti=jti).scalar()
+        return token is not None
--- a/models/revoked_tokens.py
+++ b/models/revoked_tokens.py
@@ -5,7 +5,7 @@ from models.base_model import BaseModel
 class RevokedTokensModel(BaseModel):
     __tablename__ = "revoked_tokens"
 
-    id = db.Column(db.Integer, primary_key=True)
+    id = db.Column(db.String(36), primary_key=True)
     jti = db.Column(db.String(120), nullable=False)
 
     @classmethod
--- a/app.py
+++ b/app.py
@@ -48,7 +48,7 @@ def create_app(env):
     @app.jwt.token_in_blocklist_loader
     def check_if_token_in_blocklist(jwt_header, decrypted_token):
         jti = decrypted_token["jti"]
-        return RevokedTokensModel.is_jti_blacklisted(jti)
+        return RevokedTokensModel.is_jti_revoked(jti)
 
     # Format jwt "Missing authorization header" messages
     @app.jwt.unauthorized_loader
--- a/data/seedData.py
+++ b/data/seedData.py
@@ -267,8 +267,6 @@ def seedData():
     )
     note_contacted_tenant.save_to_db()
 
-    RevokedTokensModel(jti="855c5cb8-c871-4a61-b3d8-90249f979601").save_to_db()
-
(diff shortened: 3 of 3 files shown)


NickSchimek:
Yeah, This change will work. I don't think we need the jti column anymore if we're going to store the jti as the id of the table.

I'm a bit hesitant to approve this without checking how this model is used in the application. It seems the original implementation meant for the jti column to be used, but obviously that is not what's being done. Perhaps this change is Okay though. Did you find anything to suggest otherwise?

qq1u:
> Yeah, This change will work. I don't think we need the jti column anymore if we're going to store the jti as the id of the table.
> 
> I'm a bit hesitant to approve this without checking how this model is used in the application. It seems the original implementation meant for the jti column to be used, but obviously that is not what's being done. Perhaps this change is Okay though. Did you find anything to suggest otherwise?



I think the original meaning of the program is to blocktlist n the future
However, there is currently no move to blacklist the token in the program.

models/revoked_tokens.py
```diff
@@ -5,7 +5,7 @@
 class RevokedTokensModel(BaseModel):
     __tablename__ = "revoked_tokens"
 
-    id = db.Column(db.Integer, primary_key=True)
+    id = db.Column(db.String(36), primary_key=True)
```
NickSchimek: I don't think the solution is to change the id field to a String. That does make the test pass, but will cause other issues that we are apparently missing tests for.

NickSchimek (changes requested):
When a user logs out we should store their jwt in this table, unfortunately there is no logic that I can see that handles logging out a user. I'm not sure how all this was initially setup with the frontend, but that does seem incorrect.

I don't think changing the column types to make the test pass is the correct solution.

If you look at app.py you will find the following code which calls `is_jwt_blacklisted`

https://github.com/codeforpdx/dwellinglybackend/blob/6307272acdf11ef542f0ddd9b96fc363855ac45c/app.py#L48-L51

And if you look at that method you might be able to see why there is an error:

https://github.com/codeforpdx/dwellinglybackend/blob/6307272acdf11ef542f0ddd9b96fc363855ac45c/models/revoked_tokens.py#L11-L13

It is callling the method `get` with the jwt string. get is not what we want to use here as it looks up by the id column. But we want to find a row in the table by the jwt column.

PR #344, opened 2021-07-19
Monkeypatch env setting for tests

NickSchimek:
The previous setup bled through all the environments.

Monkeypatch the environment in the conftest file so that it doesn't
bleed through all environments.

An autouse fixture defines the Flask environment for tests that do
not use the app context. For tests that do need an app context
then it also needs to be defined before the app loads which is
done in the app method.

https://docs.pytest.org/en/6.2.x/reference.html#monkeypatch

### DIFF
--- a/config/environment.py
+++ b/config/environment.py
@@ -39,7 +39,6 @@ class Development(Default):
 
 
 class Testing(Default):
-    os.environ["FLASK_ENV"] = "testing"
     TESTING = True
     SQLALCHEMY_DATABASE_URI = os.environ.get("TEST_DATABASE_URL") or "sqlite://"
     WORK_FACTOR = 4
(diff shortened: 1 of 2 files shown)


qq1u:
This line of code here, 'ENV = "testing"', needs to be used later?

aedwardg (dismissed):
This does the job, and switches back once testing is over.  Seems to be okay for our purposes right now.

NickSchimek:
> This line of code here, 'ENV = "testing"', needs to be used later?

We need to set the environment to testing when the tests run. Flask docs says to do it via env vars, but we want an automated way to do it so that users don't have to manually edit the flask env.

I think I found a better way to do it. It's been updated.

PR #350, opened 2021-08-03
removed ArchivedProperty class and assocaited code

### LINKED ISSUE #638: Remove `ArchiveProperty` class and associated code
## Description
This can be done through the update action. No additional code should be needed.

yunnie:
## Description
* Removed ```ArchiveProperty``` class from ```resources/property.py```. 
* Removed route from ```config/routes.py```
* Modified authorization and integration tests so that archiving of a single property is done through the ```put``` method in the ```property``` resource. 

Closes codeforpdx/dwellingly-app/issues/638


tests/integration/test_properties.py
```diff
@@ -152,8 +152,10 @@ def test_archive_property_by_id(self, valid_header, create_property):
         test_property = create_property()
 
         """The archive property endpoint should return a 200 code when successful"""
-        responseSuccess = self.client.post(
-            f"/api/properties/archive/{test_property.id}", headers=valid_header
+        responseSuccess = self.client.put(
+            f"/api/properties/{test_property.id}",
+            headers=valid_header,
+            json={"archived": not test_property.archived},
```
NickSchimek: When we create a property the archive value is set to False unless otherwise specified. So the value being passed into archived would be True. I do find this line a bit confusing to read, and I can see how it would be more confusing if someone doesn't have this knowledge.
Perhaps we can change `test_property = create_property()` to `test_property = create_property(archived=False)`
And line 158 to:
```suggestion
            json={"archived": True},
```
What do you think?

tests/authorizations/test_property_authorization.py
```diff
@@ -24,8 +24,10 @@ def test_archive_property_by_id(self, create_property):
         test_property = create_property()
 
         """The server responds with a 401 error if a non-admin tries to archive"""
-        responseNoAdmin = self.client.post(
-            f"/api/properties/archive/{test_property.id}"
+
+        responseNoAdmin = self.client.put(
+            f"/api/properties/{test_property.id}",
+            json={"archived": not test_property.archived},
```
NickSchimek, line 30: This is excellent! Thanks for adding this.

tests/integration/test_properties.py
yunnie: Your suggestion makes sense. The intention is to test that a property is properly archived. Using ```json={"archived": True}``` makes the intention clear.

NickSchimek (approved):
Thanks, I'll get this merged after the FE is updated.

tests/integration/test_properties.py
```diff
@@ -152,8 +152,10 @@ def test_archive_property_by_id(self, valid_header, create_property):
         test_property = create_property()
 
         """The archive property endpoint should return a 200 code when successful"""
-        responseSuccess = self.client.post(
-            f"/api/properties/archive/{test_property.id}", headers=valid_header
+        responseSuccess = self.client.put(
+            f"/api/properties/{test_property.id}",
```
scybert, line 156: what's the f for?

scybert (approved):

tests/integration/test_properties.py
NickSchimek: That's a good question. It's for string interpolation. Python has a few different ways to do string interpolation. This particular way is called "f-strings".

**Resources:**

- https://www.programiz.com/python-programming/string-interpolation
- https://www.python.org/dev/peps/pep-0498/

PR #351, opened 2021-08-05
Add updated_at and created_at to property model json

### LINKED ISSUE #671: Property model should send created_at and updated_at to frontend
**Purpose**
The Properties grid page needs to show an "Added On" data column that uses the created_at field so we should send that to the backend. The updated_at might as well be sent along with it both for consistency and so we can use it for any sorting options or showing it to the user in the future

**Proposed Solution**
Add it to the PropertyModel's json method

**Acceptance Criteria**
- [x] created_at and updated_at are returned to the frontend when querying for properties

**Additional context**

Chris-Boe:
## Description
Return property updated_at and created_at in property model json serialization so the frontend can consume it

Closes codeforpdx/dwellingly-app/issues/671

### DIFF
--- a/models/property.py
+++ b/models/property.py
@@ -3,6 +3,7 @@ from nobiru.nobiru_list import NobiruList
 from models.base_model import BaseModel
 from models.user import UserModel
 from models.property_assignment import PropertyAssignment
+from utils.time import Time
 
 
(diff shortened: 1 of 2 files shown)


aedwardg (approved):
Looks good to me 👍

PR #358, opened 2021-08-23
Initial work to remove role column from UserModel

### LINKED ISSUE #660: Update auth to check user credentials before allowing access
**Describe the bug**
Currently in the utils.authorization file we check the jwt to see if the user has the approrpiate role. However, we do not do any checks to see if the user is active.

We should grab the user identity in the jwt. Pull the user from the database. And check if they're active and authorized to access the endpoint.

Test has been written to show desired behavior. Uncomment and make it pass.
https://github.com/codeforpdx/dwellinglybackend/blob/b3c6dcee918c52ab597affb12af4394b066feb22/tests/integration/test_users.py#L108

NickSchimek:
## Description
We don't really have different roles so they are being removed. We have
different types of users which have been implemented using STI.

This is the first phase to remove roles from the backend. We will still
have a role per se as far as the frontend code is concerned.

The next phase will be to update the frontend to move off of integers as
their Role, and use the typed strings. Along with changing any hard
coded dependencies on Roles. The final phase will be back in the backend
to remove the final remnants of Roles.

Doing it in a 3 phase step as to not break anything for other
developers :-)

This is something I've been planning, but mostly doing it now because of the issue with inviting users as described in issue codeforpdx/dwellingly-app/issues/678 - This should be resolved by itself after phase 2 is completed.

This also fixes codeforpdx/dwellingly-app/issues/660 as users are now instantiated and verified as part of the jwt_required authentication process. It also ensures that the user is Active.

### DIFF
--- a/app.py
+++ b/app.py
@@ -32,16 +32,22 @@ def create_app(env):
     # initialize Mail
     app.mail = Mail(app)
 
-    # check the user role in the JSON Web Token (JWT)
+    @app.jwt.user_identity_loader
+    def user_identity_lookup(user):
+        return user.id
+
+    @app.jwt.user_lookup_loader
+    def user_lookup_callback(_jwt_header, jwt_data):
+        identity = jwt_data["sub"]
+        return UserModel.query.filter_by(id=identity, archived=False).one_or_none()
+
     @app.jwt.additional_claims_loader
-    def role_loader(identity):
-        user = UserModel.find(identity)
+    def role_loader(user):
         return {
             "email": user.email,
             "phone": user.phone,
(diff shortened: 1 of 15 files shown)


Chris-Boe (changes requested):
Just needs merge conflicts resolved. I tested with frontend PR#717 and things seem to look good

NickSchimek:
oh, I should probably add the do not merge label, because this appears to be ready, but I don't think it is.

NickSchimek:
Thanks for checking it out!!! Frontend PR 717 is not related to this PR though. This should work with the dev branch in the frontend.

PR #361, opened 2021-08-30
Update seed file to run validations when creating data

### LINKED ISSUE #515: Seed data should be validated by schemas
## Describe the bug
Much of the current seed data is not being validated by the schemas when it's being seeded.  This means there is no guarantee the data in our development and test databases is data that could actually be added to the db through normal use of the app.  

Entries that are not validated with the schemas can contain anything that is allowed by the database constraints (which are far more permissive than our schema constraints and do not enforce formatting of any kind).

## Proposed solution 
For an example of a properly validated entry, check out the [property entries](https://github.com/codeforpdx/dwellinglybackend/blob/3eff729e212ac2039d1c3e1ac84978db8bedc7d2/data/seedData.py#L143-L154).  Everything in the `seedData` file should follow this format.

Specifically, the following entry types need to be updated to use schemas:

- Users
- Tickets
- Notes
- Emergency Contacts
- Contact Numbers
- Leases

## To Reproduce
**Steps to reproduce the behavior:**
1. Go to `seedData.py`
2. Change the `ticket_id` in one of the notes to 42 and save the file (there is no ticket with an id of 42)
3. Reseed the database with `pipenv run flask db recreate`
4. DB is successfully reseeded even though the linked ticket does not exist

## Expected behavior
Nothing should be seeded into the database that would otherwise raise a `ValidationError`.

## Screenshots
image

NickSchimek:
Updated seed file to use validations
Also started generating some mass data using Faker.

Closes codeforpdx/dwellingly-app/issues/515


aedwardg (dismissed):
Looks good. It'll be nice knowing all the sample data satisfies the schemas.

PR #371, opened 2021-09-05
Randomly generate loads of seed data with faker

### LINKED ISSUE #569: Add lots of test data
**Purpose**
We should start adding more test data to the populate db script in the backend so that we can start testing the limits of the app's performance, UI layouts, grid paging/sorting/filtering, etc.

**Proposed Solution**
Add about 30-100+ additional items of each model (users with differing roles, tickets, properties, tenants both housed and unhoused, emergency contacts)

**Acceptance Criteria**
- [x] Db script adds 30+ **tickets** with various settings
- [x] Db script adds 30+ **users** with various settings
- [x] Db script adds 30+ **tenants** with various settings
- [x] Db script adds 30+ **properties** with various settings
- [x] Db script adds 30+ **emergency contacts** with various settings

**Additional context**

NickSchimek:

Closes codeforpdx/dwellingly-app/issues/569

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -5,4 +5,4 @@ ignore = E203,E501,W503
 exclude = __init__.py
 per-file-ignores =
     shelltools.py:F401
-    seed.py:F405
+    seed.py:F405,E712
--- a/data/seed.py
+++ b/data/seed.py
@@ -6,28 +6,40 @@
 # implicit imports in models will fail to load.
 #
 
+import random
+from faker import Faker
+
 from models.user import UserModel, RoleEnum
 from models.users.admin import Admin
 from models.users.staff import Staff
 from models.users.property_manager import PropertyManager
 from models.property import PropertyModel
 from models.tenant import TenantModel
-from models.tickets import TicketModel, TicketStatus
+from models.tickets import TicketModel
 from models.notes import NotesModel
 from models.emergency_contact import EmergencyContactModel
 from models.contact_number import ContactNumberModel  # noqa: F401
 from models.lease import LeaseModel
 from schemas import *  # noqa: F403
-from utils.time import Time
-from faker import Faker
+from tests.factory_fixtures.contact_number import contact_number_attrs
+from tests.factory_fixtures.emergency_contact import emergency_contact_attrs
+from tests.factory_fixtures.property import property_attrs
+from tests.factory_fixtures.tenant import tenant_attrs
+from tests.factory_fixtures.lease import lease_attrs
+from tests.factory_fixtures.ticket import ticket_attrs
+from tests.factory_fixtures.notes import note_attrs
+
+MIN_HOUSED_TENANTS = 150
+MIN_TENANTS = 300
+MAX_TENANTS = 500
 
 
 class Seed:
-    @classmethod
-    def data(cls):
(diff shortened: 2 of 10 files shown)


aedwardg (commented):
This is exciting!

Just a couple things I noticed:

First, I noticed that when I reseeded the database, it said that there were no archived properties created.  Do we want to always have at least 1, or is this okay?

image

Also, when I tried to go to the Properties page, I got an error (postgres dev db).  This doesn't happen on the `development` branch, so I think it might be related to something here.

image

NickSchimek:
Thanks for reviewing this @aedwardg !!!

I will look into what's causing the error.

Regarding the 0 archived properties.
I think that's OK, but what concerns me more is that every time this is ran the same random values will be generated. I get 0 archived properties too.
There is not really such a thing as random when generating a random number with a computer.
I thought that maybe Python might fake it well, but it turns out it doesn't and is deterministic. So `random.randint(0, 15)` will always return 0 the first time.

```python
>>> import random
>>> random.randint(0, 15)
0
```

I'll improve the random functionality and check out what's wrong with displaying properties. Thank you.

NickSchimek:
Well, it looks like it seeds from the system time, or specific random value provided by os.
https://github.com/python/cpython/blob/main/Lib/random.py#L134

But, I find it strange that I haven't seen it be something other than 0 yet.

I wasn't instantiating a random object. I think if I do that we'll see better results. Alternatively I can ask the user to smash the keys on their keyboard for a seed value.

NickSchimek:
That error does occur on development too. Looks like something might have changed in the frontend recently. 

I would like to see system tests being created whenever something changes that is currently not covered. 

ah found the issue. It looks like it was just introduced. Going to revert it on the frontend https://github.com/codeforpdx/dwellingly-app/pull/690

aedwardg:
For me, it doesn't give that error when I re-seed the db on the backend development branch. Here's a screenshot with db reseeded on backend `development` and frontend `development` pre-revert (HEAD is commit 8298855)

image

NickSchimek:
@aedwardg right when you reseed the data with the current data it does not give the error.
Try creating a new property and then viewing the page.
I just updated the frontend dev branch so its fixed on development now.

aedwardg:
> Try creating a new property and then viewing the page.

Aha! That's sneaky, good catch!
I'm glad we're finally getting all this new data created so we can expose this type of bug 🙂

aedwardg:
@NickSchimek 
I can't reproduce `random.randint` always giving 0 first.

image

NickSchimek:
😵‍💫 There was a slight flaw in my logic. It's fixed. 
```
8 Active admins created
2 Archived admins created
47 Active staff created
47 Archived staff created
90 Active property managers created
39 Archived property managers created
5 Pending users created
3 Denied users created
74 Active properties created
9 Archived properties created
341 Active tenants created
28 Archived tenants created
241 Leases created
1584 Tickets created
41 Emergency Contacts created
```

NickSchimek:
I'm going to merge and reseed the remote server. Unless you have anything else?

aedwardg (approved):
Awesome, should be good to merge now!

PR #372, opened 2021-09-06
Move faker library to production

NickSchimek:
This is needed to seed the remote development server.


NickSchimek:
PR https://github.com/codeforpdx/dwellinglybackend/pull/373 is a better solution.

PR #376, opened 2021-09-30
update tickets URL

### LINKED ISSUE #701: opentickets widget returns incorrect link
**Describe the bug**
The opentickets widget resource gives a link of "/tickets", which leads to a 404.

**To Reproduce**
1. Go to '/dashboard'
2. Click on 'Open Tickets'
3. This goes to "/tickets", which gives a 404.

**Expected behavior**
I would expect this to be a page displaying tickets, I'm assuming "/manage/tickets" is the correct URL

 - OS: Windows 10 (using WSL2)
 - Browser: Firefox 
 - Version: 92.0.1

TNorbury:
## Description
Returns the correct link of "/manage/tickets" from the opentickets widget.

Closes codeforpdx/dwellingly-app/issues/701

### DIFF
--- a/resources/widgets.py
+++ b/resources/widgets.py
@@ -55,7 +55,7 @@ class Widgets(Resource):
         return {
             "opentickets": {
                 "title": "Open Tickets",
-                "link": "/tickets",
+                "link": "/manage/tickets",
                 "stats": [
                     [
                         {


NickSchimek:
Thanks for the PR @TNorbury - This change looks good. However, may I suggest an alternate solution? I don't think the backend should be responsible for knowing the frontend links. I think the widgets section in the backend was something that was hashed together pretty quickly, I would prefer to see the url removed from the backend, and make the frontend responsible for this information. What do you think?

NickSchimek:
THis may be why this bug exists. A frontend link/route was changed, and it wasn't obvious the backend had hard-coded this.

I don't think there should be any hard-coded links of this kind in the backend, and that the frontend should be responsible for maintaining their links.

NickSchimek:
I will go ahead and merge this in. Please feel free to open frontend issue to remove the dependency on hard-coded backend information. I got some open issues that are the start of refactoring this dashboard (widget) endpoint, so hopefully it will get cleaned up eventually.

NickSchimek (approved):

TNorbury:
@NickSchimek  Yeah, that makes sense! Really all we need from the backend is just the number of tickets and such, all the text should live on the front end. I'll go ahead and up an issue for that 👍🏻

NickSchimek:
Awesome, thank you!

PR #377, opened 2021-09-30
Fix notes time format

htharker42:
##The notes model was using incorrect time formatting on the JSON object it returned.  I updated the date/time formatting in the JSON object for the notes model to match the date/time format used throughout the system. 

Doesn't close a specific issue.


Chris-Boe (dismissed):

Chris-Boe (changes requested):
Change looks good but might want to start with a fresh branch from most recent development, not sure what happened to the commit history here 😅

PR #383, opened 2021-10-07
Add clarification for postgresql version

Dylan-Floyd:
## Description
The PostgreSQL download page shows multiple versions, but the guide didn't specify which version should be used. I added "version 14" next to the link to reduce ambiguity.


NickSchimek (approved):
Thanks!

PR #386, opened 2021-10-12
704/remove backend dependency

### LINKED ISSUE #704: Remove dependency on backend for dashboard widgets
**Purpose**
Follow up to discussion in [this PR](https://github.com/codeforpdx/dwellinglybackend/pull/376)
Currently the two widgets on the dashboard get all their information (text, links, and data) from the backend server, which can lead to some issues if one changes and not the other. 

**Proposed Solution**
This would change things so that front-end information is defined in the front-end, and just the raw data will be retrieved from the backend. I'm thinking that we'd still make the same server call inside of the Dashboard component, but then construct the objects with the various text and links, and add the data from the server to the objects. These objects would then be passed to the modules as usual.

**Acceptance Criteria**
- [ ] Text and links for the dashboard widgets are defined in the front end app

TNorbury:
Please connect this Pull Request to the issue it is resolving by using the controls on the right side of this PR.
closes codeforpdx/dwellingly-app#704 

This doesn't provided any change in behavior, it was rather just a re-assigning of who was responsible for what data. Now the frontend handles widget information such as the title, link, and labels. Whereas the backend is just concerned with providing the raw data.

There is a complementary PR in the frontend repo that'll need to be closed at the same time:  codeforpdx/dwellingly-app#710

### DIFF
--- a/tests/integration/test_widgets.py
+++ b/tests/integration/test_widgets.py
@@ -1,0 +1,31 @@
+import pytest
+
+
+@pytest.mark.usefixtures("client_class", "empty_test_db")
+class TestWidgets:
+    def setup(self):
+        self.password = "strongestpasswordever"
+
+    def test_get_widgets(self, create_admin_user):
+        user = create_admin_user(pw=self.password)
+        response = self.client.post(
+            "/api/login",
+            json={"email": user.email, "password": self.password},
+        )
+        header = {"Authorization": f"Bearer {response.json['access_token']}"}
+        widgetResponse = self.client.get("/api/widgets", headers=header)
+
+        assert widgetResponse.status_code == 200
+        widgetJson = widgetResponse.json
+        assert len(widgetJson["opentickets"]) == 2
+        assert len(widgetJson["managers"]) == 0
+
+        # check for correct keys for new tickets
+        assert "new" in widgetJson["opentickets"]
+        assert "allNew" in widgetJson["opentickets"]["new"]
+        assert "unseen24Hrs" in widgetJson["opentickets"]["new"]
+
+        # check for correct keys for in-progress tickets
+        assert "inProgress" in widgetJson["opentickets"]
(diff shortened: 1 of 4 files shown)


TNorbury:
The failed test will be fixed by #385

tests/integration/test_widgets.py
```diff
@@ -0,0 +1,31 @@
...
+        self.password = "strongestpasswordever"
+
+    def test_get_widgets(self, create_admin_user):
+        user = create_admin_user(pw=self.password)
+        response = self.client.post(
+            "/api/login",
+            json={"email": user.email, "password": self.password},
+        )
+        header = {"Authorization": f"Bearer {response.json['access_token']}"}
+        widgetResponse = self.client.get("/api/widgets", headers=header)
```
NickSchimek: Thanks for writing tests! Let's simplify this as login is already well tested, and it will make the test a bit easier to read.
```suggestion
    def test_get_widgets(self, valid_header):
        widgetResponse = self.client.get("/api/widgets", headers=valid_header)
```

tests/integration/test_widgets.py
```diff
@@ -0,0 +1,31 @@
...
+
+        assert widgetResponse.status_code == 200
```
NickSchimek: style: Let's use snake_case instead of camelCase in the backend where possible

NickSchimek (changes requested):
Thank You! - I think this looks good.

@Chris-Boe Since this change impacts the front-end, can you give this a review?

I've requested a couple small changes to the test file.

Chris-Boe:
> Thank You! - I think this looks good.
> 
> @Chris-Boe Since this change impacts the front-end, can you give this a review?
> 
> I've requested a couple small changes to the test file.

Yeah I plan on testing this with the frontend PR this weekend! Should get to it tomorrow after I finish up my own PR's

NickSchimek (approved):

Chris-Boe (approved):

PR #387, opened 2021-10-12
document steps I took to get psql working

TNorbury:
## Description
While setting up psql on my machine, ran into some hickups. Added documentation on what I did to get it working.


NickSchimek (approved):
Thanks! This is great!

PR #388, opened 2021-10-17
Enable user to turn off/on logging from sqlalchemy

NickSchimek:
Sometimes the logging is useful, but other times it just gets in the
way. Enable the user to turn off/on logging as they desire from their
`.env` file.

For example when tests fail you'll often see large output from this logger, which can be helpful sometimes, but most of the time you just want to see the failure message.

### DIFF
--- a/app.py
+++ b/app.py
@@ -56,7 +56,7 @@ def create_app(env):
     def format_unauthorized_message(message):
         return {app.config["JWT_ERROR_MESSAGE_KEY"]: message.capitalize()}, 401
 
-    if os.getenv("DB_LOGGING"):
+    if os.getenv("DB_LOGGING") == "ON":
         logging.basicConfig()
         logging.getLogger("sqlalchemy.engine").setLevel(logging.INFO)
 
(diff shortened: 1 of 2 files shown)


app.py
```diff
@@ -55,8 +56,10 @@ def check_if_token_in_blocklist(jwt_header, decrypted_token):
     def format_unauthorized_message(message):
         return {app.config["JWT_ERROR_MESSAGE_KEY"]: message.capitalize()}, 401
 
-    logging.basicConfig()
-    logging.getLogger("sqlalchemy.engine").setLevel(logging.INFO)
+    if os.getenv("DB_LOGGING"):
```
TNorbury: Would it be better to do something like:
```suggestion
    if os.getenv("DB_LOGGING") == "ON":
```

Just thinking that some people (namely me), might try to turn logging off by setting the env variable to "OFF"

TNorbury (dismissed):
Looks good!

app.py
aedwardg: This is a fair point. I can definitely see people changing this to "OFF" later instead of commenting it out.

aedwardg (dismissed):

app.py
NickSchimek: Yes, thank you!

PR #390, opened 2021-10-17
Fix UserInvite to create the correct type of user

### LINKED ISSUE #678: An invited user is not granted authorization to the app
**Description**

The UserInvite resource in the backend will always create a User of type User. Meaning that they are not an authorized user until approved. The frontend uses this to create a property manager, and possibly other uses as well.

When used to create a property manager the FE provides the role. And since this is an Admin only function it makes sense that the admin would not need to approve the user after inviting the user.

Currently in the Invite we always create the User with `UserModel`, however if we are given a role in the payload then we should create that type of user as provided by the role. 

One option would be to create the user using the role that is passed. For instance if given a property-manager role then we should create the user with the PropertyManager class. 

Another option would be to create the user using the UserModel class, but assign the type of the user when creating the user.

NickSchimek:
When we switched to STI for users this resource was not updated. I
believe it was assumed that inviting users created PendingUsers, but
this endpoint is actually used to create one of the three types of
users.

Fixes codeforpdx/dwellingly-app/issues/678

### DIFF
--- a/resources/user_invite.py
+++ b/resources/user_invite.py
@@ -1,9 +1,12 @@
 from flask_restful import Resource
 from flask import request
 from utils.authorizations import admin_required
-from models.user import UserModel
+from models.users.property_manager import PropertyManager
+from models.users.staff import Staff
+from models.users.admin import Admin
 from schemas import UserSchema
 from resources.email import Email
+from models.user import RoleEnum
 import string
 import random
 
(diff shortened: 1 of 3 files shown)


NickSchimek:
Thinking about this some more. This is actually a frontend issue. It should be updated to send type instead of role

PR #391, opened 2021-10-18
Implement strategy for unique validations

### LINKED ISSUE #709: Attempting to grant a user access returns a 400
**Describe the bug**
When granting access to a user via the Request Access page, a 400 error is returned when access is granted

**To Reproduce**
1. From the admin dashboard, click on a user under the "Request for Access section"
2. Choose a role and hit "Grant Access"
3. A message will pop up saying "Request failed with status code 400"

**Expected behavior**
I expect the user to be granted access (I assume, I'm actually ignorant of what the expected behavior is

**Screenshots**
image

 - OS: Windows 10 (WSL)
 - Browser Firefox
 - Version 92.0.1

NickSchimek:
## Description

Avoid the false positives when validating a unique field. 
Uses Anthony's context implementation to pass the original value of fields, except the context is assigned automatically. It just needs to be explicitly set in the model, and then it will be available in the schema's for validations.

Refactored the app, and implemented this on all unique fields.

Fixes codeforpdx/dwellingly-app/issues/709

### DIFF
--- a/schemas/property.py
+++ b/schemas/property.py
@@ -27,13 +27,11 @@ class PropertySchema(ma.SQLAlchemyAutoSchema):
 
     @validates("name")
     def validates_uniqueness_of_name(self, value):
-        def _assigning_name():
-            if "name" in self.context and value == self.context["name"]:
-                return False
-            return True
-
-        if _assigning_name() and PropertyModel.find_by_name(value):
+        if self.context.get("name") != value and PropertyModel.find_by_name(value):
             raise ValidationError("A property with this name already exists")
+
+    @validates("name")
+    def validates_prescense_of_name(self, value):
         if blank(value):
             raise ValidationError("Property name cannot be blank")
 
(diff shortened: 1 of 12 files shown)


TNorbury (dismissed):

schemas/property.py
```diff
@@ -27,13 +27,11 @@ def validates_property_manager_ids(self, value):
...
-            if "name" in self.context and value == self.context["name"]:
-                return False
-            return True
-
-        if _assigning_name() and PropertyModel.find_by_name(value):
+        if self.context.get("name") != value and PropertyModel.find_by_name(value):
             raise ValidationError("A property with this name already exists")
+
+    @validates("name")
+    def validates_prescense_of_name(self, value):
```
Chris-Boe:
```suggestion
    def validates_presence_of_name(self, value):
```

Chris-Boe (dismissed):
Thanks for fixing this!

Chris-Boe (approved):

PR #392, opened 2021-10-22
Remove url helper4 emails

### LINKED ISSUE #461: Remove the url_for helper methods in the emails with the FE specific route
**Purpose**
url_for generates the link to the BE, but we need to direct the user to the FE page. There should be a config var that can be used to grab the base url that will be used in production.

yunnie:
## Description
Remove ```url_for``` from email templates. Replaced with an app configuration variable ```FRONTEND_BASE_URL```. 

Closes codeforpdx/dwellingly-app/issues/461

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -4,9 +4,6 @@ PYTHONPATH=${PWD}:${PYTHONPATH}
 ## Uncomment to see logging from sqlalchemy
 # DB_LOGGING=ON
 
-## Uncomment to set front end base url
-#FRONTEND_BASE_URL="https://localhost:3000"
-
 ## To set the database to postgresql. Uncomment the env vars below
 # To set the database to postgresql. Uncomment the env vars below
 # DEV_DATABASE_URL=postgresql://localhost/dwellingly_development
--- a/.env.example
+++ b/.env.example
@@ -4,6 +4,9 @@ PYTHONPATH=${PWD}:${PYTHONPATH}
 ## Uncomment to see logging from sqlalchemy
 # DB_LOGGING=ON
 
+## Uncomment to set front end base url
+#FRONTEND_BASE_URL=https://localhost:3000
+
 ## To set the database to postgresql. Uncomment the env vars below
 # DEV_DATABASE_URL=postgresql://localhost/dwellingly_development
 # TEST_DATABASE_URL=postgresql://localhost/dwellingly_test
--- a/.env.example
+++ b/.env.example
@@ -5,8 +5,9 @@ PYTHONPATH=${PWD}:${PYTHONPATH}
 # DB_LOGGING=ON
 
 ## Uncomment to set front end base url
-#FRONTEND_BASE_URL=https://localhost:3000
+#FRONTEND_BASE_URL="https://localhost:3000"
 
 ## To set the database to postgresql. Uncomment the env vars below
+# To set the database to postgresql. Uncomment the env vars below
 # DEV_DATABASE_URL=postgresql://localhost/dwellingly_development
 # TEST_DATABASE_URL=postgresql://localhost/dwellingly_test
--- a/config/environment.py
+++ b/config/environment.py
@@ -23,7 +23,7 @@ class Default(object):
     MAIL_DEFAULT_SENDER = os.environ.get("MAIL_DEFAULT_SENDER")
 
     # Default front end url
-    FRONTEND_BASE_URL = os.environ.get("FRONTEND_BASE_URL")
+    FRONTEND_BASE_URL = os.environ.get("FRONTEND_BASE_URL", "http://localhost:3000")
 
     # Configure JWT error message key
     JWT_ERROR_MESSAGE_KEY = "message"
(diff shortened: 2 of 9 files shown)


yunnie:
I overrode ```FRONTEND_BASE_URL``` for the testing environment so tests wouldn't fail if people didn't update the ```.env``` file. But I'm not sure this is the best approach.

config/environment.py
```diff
@@ -22,6 +22,9 @@ class Default(object):
     MAIL_PASSWORD = os.environ.get("MAIL_PASSWORD")
     MAIL_DEFAULT_SENDER = os.environ.get("MAIL_DEFAULT_SENDER")
 
+    # Default front end url
+    FRONTEND_BASE_URL = os.environ.get("FRONTEND_BASE_URL")
```
NickSchimek: Maybe we can do this instead? And then we can remove the override below, and it wont be necessary to set any env vars in development. If the env var set it will be used; otherwise localhost is used.

```suggestion
    FRONTEND_BASE_URL = os.environ.get("FRONTEND_BASE_URL", "http://localhost:3000")
```

.env.example
```diff
@@ -4,6 +4,10 @@ PYTHONPATH=${PWD}:${PYTHONPATH}
 ## Uncomment to see logging from sqlalchemy
 # DB_LOGGING=ON
 
+## Uncomment to set front end base url
+#FRONTEND_BASE_URL="https://localhost:3000"
```
NickSchimek: If the suggested update in the config file is used then we don't this.

resources/email.py
```diff
@@ -9,16 +10,27 @@ class Email:
     @staticmethod
     def send_reset_password_msg(user):
         token = user.reset_password_token()
+        FRONTEND_BASE_URL = app.config["FRONTEND_BASE_URL"]
```
NickSchimek, line 13: This seems strange to me to set constants in methods. But maybe that's how it's done in Python?
Can we move this up to the class level so that we don't have to do this in the other methods too?
Or can we access the env var directly from the email template without passing it in?

tests/unit/test_email.py
```diff
@@ -0,0 +1,40 @@
...
+            for template in templates:
+
+                email = render_template(
+                    template,
+                    FRONTEND_BASE_URL=FRONTEND_BASE_URL,
+                    user=user,
+                    token=token,
+                )
+
+                link = "{}/changePassword?token={}".format(FRONTEND_BASE_URL, token)
```
NickSchimek: Generally for tests I try to be specific. Looking at this I'm not quite sure what the URL is.

Perhaps we can do:

```python
assert "https://localhost:3000/changePassword?token=some_token_name_thats_set_000" in email
```

NickSchimek: Or instead of setting the testing config, we could do:

```python
                email = render_template(
                    template,
                    FRONTEND_BASE_URL="https://my_url.com,
                    user=user,
                    token="my_token",
                )
```

And then test that those passed in arguments are interpolated and rendered.

```python
assert "https://my_url.com/changePassword?token=my_token"in email
```

Since this is a unit test we can pass in any params we need.

tests/unit/test_email.py
```diff
@@ -0,0 +1,40 @@
+from app import create_app
+import os
+from flask import render_template
+
+
+class TestEmailTemplateRender:
+    def test_reset_password_html_msg(self, user_attributes):
+        user = user_attributes()
+        # need to find a user fixture and a token fixture
+        token = "fake-token"
```
NickSchimek: a token fixture is not necessary for this test. Just pass in the string, and test it outputs correctly.

NickSchimek (commented):
Thanks for writing that test!!! Looks good overall. I made a couple small comments. Take them or leave them.

I'm surprised our style linter accepts all caps for parameter names. It seems kind of strange to me to use parameter name with all caps. Does that actually set a constant in python? or can it still be modified?

config/environment.py
NickSchimek: Using http here since we don't have https setup for dev env

yunnie: Didn't realize that ```os.environ.get``` accepted multiple parameters.

resources/email.py
yunnie: I tried setting ```FRONTEND_BASE_URL``` as a class constant. The problem is that ```app``` requires an application context. I get ```RuntimeError:  Working outside of application context.``` error. I tried setting ```FRONTEND_BASE_URL``` in a context: 

```with app.app_context(): 
    FRONTEND_BASE_URL = app.config["FRONTEND_BASE_URL"]```

The only way I could get this to work was to put assign the constants in the methods.

NickSchimek: Interesting, do we need to use app.config here? Could we instead use os.environ.get ? Since it's technically not application config, and rather an environment variable. But I don't know if we'd run into a different issue doing that.

And then we'd have to assign it in the .env file I think. Instead of the config file. 

I think I like defining it in the config file rather than the .env file.

yunnie:
Setting constants in the method calls does seem odd. I did try to set ```FRONTEND_BASE_URL``` as a class variable, but ran into a number of problems with Flask's context manager.

NickSchimek:
Yeah, the downsides of using Flask. I've ran into issues with that before too.

NickSchimek (approved):
Thanks!

PR #394, opened 2021-10-24
Issue #719 changed the user model type column nullable to false

### LINKED ISSUE #719: User type should probably be not nullable
Change the db structure on the user model to make the type column `nullable = False`.

LogMoss:
## Description
I changed the user model type column nullable to false.

Closes codeforpdx/dwellingly-app/issues/719

### DIFF
--- a/models/user.py
+++ b/models/user.py
@@ -39,7 +39,7 @@ class UserModel(BaseModel):
     __mapper_args__ = {"polymorphic_identity": "user", "polymorphic_on": "type"}
 
     id = db.Column(db.Integer, primary_key=True)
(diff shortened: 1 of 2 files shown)


NickSchimek:
Thanks, it looks like the tests are failing due to the auto schema for the validations. 

Are you able to update the schema?

NickSchimek:
I think what needs to be added is this. 

`type = fields.Str(required=False)` to the schemas/user.py file

We don't need to validate this field since the user has no control over it.

NickSchimek (approved):
Thanks!

PR #409, opened 2021-11-04
Update creat tenant factory

### LINKED ISSUE #640: Create/update the create_tenant factory to optionally create assigned join staff
## Description
Create/update the create_tenant factory to optionally create assigned join staff

Tenants can have assigned join staff. When we create a tenant we should have the option to have join staff assigned to that tenant. I think this can be done within the existing create_tenant factory as an option. If the option passes specific join staff objects then assigned them to the tenant. If the option passes an `join_staff =3` a number of assigned staff members then simply create that many join staff and assign them to the tenant. Otherwise do not assign any create and staff members to the tenant. 

If that is to complex then create another factory that creates assigned join staff to the tenant. so either modify the existing factory or create a new factory.

These updates should be unit tested through the tenant model.

yunnie:

Closes codeforpdx/dwellingly-app/issues/640

I know the request was to add a ```join_staff=3``` variable to the ```create_tenant()``` fixture, but  I think ```create_tenant(3)```  is confusing. I would assume that it would create 3 tenants, but it creates a single tenant with 3 assigned staff.

I've set it up so that you need to use the ```create_join_staff``` fixture directly. 

```
  staff = [create_join_staff().id for _ in range(n)]
  create_tenant(staff)
```

It's an additional step, but I think the intension is clearer.

### DIFF
--- a/tests/integration/test_staff_tenant.py
+++ b/tests/integration/test_staff_tenant.py
@@ -37,9 +37,9 @@ class TestStaffTenant:
     def test_when_provided_no_staff_params(
         self, valid_header, create_tenant, create_join_staff
     ):
-        tenant_1 = create_tenant([])
-        tenant_2 = create_tenant([])
-        keep_tenant = create_tenant([])
+        tenant_1 = create_tenant()
+        tenant_2 = create_tenant()
+        keep_tenant = create_tenant()
         StaffTenantLink(
             tenant_id=tenant_1.id, staff_id=create_join_staff().id
         ).save_to_db()
@@ -72,7 +72,7 @@ class TestStaffTenant:
     def test_when_provided_staff_params(
         self, valid_header, create_tenant, create_join_staff
     ):
-        tenant = create_tenant([])
+        tenant = create_tenant()
         old_staff = create_join_staff()
         StaffTenantLink(tenant_id=tenant.id, staff_id=old_staff.id).save_to_db()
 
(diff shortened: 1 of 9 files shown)


yunnie:
But if everyone feels that ```create_tenant(n)``` is preferable, I can make the change.

tests/integration/test_staff_tenant.py
```diff
@@ -37,9 +37,9 @@ class TestStaffTenant:
     def test_when_provided_no_staff_params(
         self, valid_header, create_tenant, create_join_staff
     ):
-        tenant_1 = create_tenant()
-        tenant_2 = create_tenant()
-        keep_tenant = create_tenant()
+        tenant_1 = create_tenant([])
```
NickSchimek: I'm not fond of having to know that an empty array needs to be passed in to the fixture in order for the method call to work. Can we make the parameter optional?

tests/unit/test_tenant.py
```diff
@@ -60,3 +59,8 @@ def test_json(self, create_tenant, create_lease):
 class TestTenantFactory:
     def test_create_tenant(self, create_tenant):
         assert create_tenant()
+
+    def test_create_tenant_with_staff(self, create_tenant, create_join_staff):
+        staff = [create_join_staff().id for _ in range(3)]
+        tenant = create_tenant(staff)
```
NickSchimek, line 41: This is good.
Do you think it would be beneficial to create the staff inside of the create_tenant fixture instead?
Maybe do something like? `create_tenant(staff=3)`. But then there is the downside of not being able to control how the staff is created. Perhaps the create_tenant fixture could be smart. If it's given a list of staff objects then it simply assigns those, if its provided an integer then it creates that many staff members and assigns them. If there are no arguments then it creates a tenant without any assigned staff.

What do you think?

NickSchimek (commented):
> create_tenant(3) is confusing. I would assume that it would create 3 tenants

Yes, I would assume that as well too. Good catch.

Everything looks good. I made a couple comments.

There are 6 commits in this PR, can you rebase/remove the irrelevant ones and force push?

Thanks!

NickSchimek (approved):
Thanks @yunnie !

PR #420, opened 2021-11-27
Serialize tenants for staff user requests

### LINKED ISSUE #725: Create Edit/View JOIN Staff/Admin page
**Purpose**
A way to view and edit the user properties of a JOIN Staff member or Admin, and promote or demote from Admin

**Proposed Solution**
Take the Add Join Staff page and just replicate it over into a View/Edit page, just like our other ones

**Acceptance Criteria**
- [x] Should have an Edit pencil button to activate edit mode
- [x] Fields should be pre-populated with the user's info
- [x] First, Last, Email, and Phone should be editable
- [ ] Checkbox to promote or demote a user from Admin
- [ ] List of Tenants assigned to JOIN Staff member should be shown
- [ ] X button shown next to each tenant row to remove the tenant (while in edit mode) - See Property View page
- [x] Save/Cancel buttons to Save all new settings or reset form to original values
- [ ] "Reassign" link on the JOIN Staff card and the name of the staff member should link to the View page for that user
- [ ] Get rid of "Set out of office" text on the JOIN Staff card view
- [ ] Increase the font a bit and the size of the JOIN Staff cards to reduce the amount of whitespace and increase readability of the card text

**Additional context**

Chris-Boe:
## Description
Returns tenants with user get requests for staff users

Closes codeforpdx/dwellingly-app/issues/725


models/users/admin.py
```diff
@@ -1,9 +1,19 @@
...
 class Admin(UserModel):
     __mapper_args__ = {"polymorphic_identity": "admin"}
 
+    tenants = db.relationship(
+        "TenantModel",
+        secondary=StaffTenantLink.tablename(),
+        collection_class=NobiruList,
+        viewonly=True,
+    )
+
```
NickSchimek, line 16: The way things are now: This will always return an empty array. An admin cannot be assigned tenants. I had asked Hugh about this, and he said that only join staff (Staff) can be assigned tenants, and that Admin's should not be able to be assigned tenants. So that's what's been implemented.

If this needs to be changed so that both Admin and Staff can be assigned tenants. Then let me know. I did see in the frontend that logic there had implemented that only admins can be assigned staff, which I submitted changes for in one of my PR's. So it seems there may be a disconnect or misunderstanding here somewhere.

Staff has an association with tenants.
https://github.com/codeforpdx/dwellinglybackend/blob/c992a1ed8f881b69ca67a0e5de8b46daa17fb445/models/users/staff.py#L10-L15

